PR #10, opened 2019-03-10
Vasilis/docs

vasilismsr:


doc/spec/estimation/forest.rst
```diff
@@ -1,10 +1,156 @@
 Forest Based Estimators
 =======================
 
-\
```
kbattocchi: As I recall this line is necessary because of how sphinx identifies page headings.  At the top of a document, if you have a heading followed by regular whitespace and then a subheading then the first heading is assumed to be the title and the subheading is assumed to be a subtitle, rather than a document subheading.  This can be avoided by incorporating an escaped space (or any actual text) in between the document title and the subheading.

doc/spec/estimation/forest.rst
```diff
@@ -1,10 +1,156 @@
...
+    Y_{i,t}^{DR} = \hat{\E}[Y \mid T=t, x, W_i] - 1\{T_i=t\} \frac{Y_i - \hat{\E}[Y \mid T=t, x, W_i]}{\hat{\E}[1\{T=t\} \mid x, W_i]} 
+
+where :math:`\hat{\E}[Y \mid T=t, x, W_i]` and :math:`\hat{\E}[1\{T=t\} \mid x, W_i]` are first stage estimates of the
+corresponding conditional expectations. These two regression functions are fitted in a similar manner
+as in the continuous treatment case. However, in the case of discrete treatment, the model for the treatment is 
+a multi-class classification model and should support :code:`predict_proba`.    
+
+For more details on the input parameters of the orthogonal forest classes and how to customize
+the estimator checkout the two modules:
+    - :py:class:`~econml.ortho_forest.DiscreteTreatmentOrthoForest`
```
kbattocchi: The indentation on these lines is causing a build error.  I think the bullets do not need to be indented, but you may need a blank line before the first one.

doc/spec/estimation/forest.rst
```diff
@@ -1,10 +1,156 @@
...
+    - :py:class:`~econml.ortho_forest.ContinuousTreatmentOrthoForest`
+
+
+For more examples check out our 
+`OrthoForest Jupyter notebook <https://github.com/Microsoft/EconML/blob/master/notebooks/Orthogonal%20Random%20Forest%20Examples.ipynb>`_ 
+
+Examples
+--------
```
kbattocchi: Consider making this a sub-subheading by using a different underline character, such as `^^^^^^^`

doc/spec/estimation/forest.rst
vasilismsr: So what do I need to add here?

kbattocchi: Just don't delete line 4.

vasilismsr: Done

doc/spec/estimation/forest.rst
vasilismsr: Done

doc/spec/spec.rst
```diff
@@ -8,8 +8,8 @@ This is the spec.
     api
     estimation
     inference
-    benchmark
```
kbattocchi: There's another build break because now the benchmark.rst file isn't referenced from anywhere else; please delete the .rst file to correspond to this change.

vasilismsr: Done

PR #15, opened 2019-03-21
Correct DML shapes for vectors

kbattocchi:
This set of changes makes it so that DML can work with vector T or Y, and also makes X optional to comply with the signature.


kbattocchi:
This addresses issues #16 and #17 for DML, but I want to make another pass to make sure that all other estimators also handle these cases.

PR #23, opened 2019-03-27
Add test to run each notebook, reporting all errors

kbattocchi:


kbattocchi:
This set of changes does two things:
1. Tests that we can run each jupyter notebook, collecting all errors encountered.
2. Better modularizes our build pipeline so that the notebook and documentation tests run in parallel to our other tests (with notebooks and documentation being tested only on macOS with Python 3.6).

kbattocchi:
@moprescu How long would you expect the ortho notebook to run?  It looks like that's what's causing the build failure...

kbattocchi:
@moprescu Even with Vasilis's perf improvements the notebooks still take 45 minutes to run.  Any ideas about opportunities for further performance improvement?

vsyrgkanis:
I think we should separate the OJ dataset. That’s the part that takes 45. In particular the final bootstrap. Without the final bootstrap the whole notebook just takes at most 7min. We could also just remove that final bootstrap.

vsyrgkanis:
Maybe we could do a bootstrap example in the synthetic data cells above but not in the OJ

kbattocchi:
@vasilismsr It looks like if we just use fewer trees (say n_trees=100 instead of 2000) we can cut the time way down, at the cost of having wider confidence intervals.  Then we could add some explanatory text pointing out that for greater fidelity you could use more trees but we used fewer for the sake of running more quickly.  Do you see a problem with this?

vasilismsr:
that seems a good solution too

PR #24, opened 2019-03-28
add double ml notebook and fix a minor typo in dml doc

heimengqi:
please take a look at my notebook, especially the last part (multi treatment and multi output). Let me know if I am doing anything wrong or anything I need to add.


vsyrgkanis:
I would change how we plot the cross price elasticities in the final cells and write some more comments in the way that you process the data to fit cross price. For how to depict: can we do something like: we have 3x3 subplots and each subplot contains the cross price elasticity as a function income. This way we can visualize it as a matrix and each matrix contains a plot. Also I would like to have an example that uses the FistaRegressor from here: http://contrib.scikit-learn.org/lightning/generated/lightning.regression.FistaRegressor.html as our final model and we use 'nuclear' as the penalty. Similar to what we had in the demo presentation. This would show how to perform nuclear norm penalization for multiple treatments and multiple outcomes. This is important for latent factor models (i.e. the products interact in some low rank latent space).

kbattocchi:
@vasilismsr From an economist's point of view, is the nuclear norm the right regularization?  Low rank is nice, but it seems like they'd be more likely to try to apply a mixed effects model of some sort (e.g. shrink all individual own-price elasticities towards one value and all cross-price elasticities towards another).

vsyrgkanis:
Yeap. Low rank is the right way for high dimensional product spaces. Such latent factor models have been studied in pricing (see athey and blei, or the paper we wrote with taddy).

kbattocchi:
@vasilismsr But here we aren't in a high-dimensional regime; we're regressing the three log quantites on the three prices interacted with income (plus an intercept), so there are only 18 coefficients total.

Maybe we should see if we can find (or generate) a high-dimensional example as well, so that we can demonstrate using the trace norm in a more appropriate setting.

vsyrgkanis:
I know. But the notebook is more for expository purposes, i.e. you could even do this. And we can say in the cell above that this would be more appropriate when you have many products and you believe they interact in a latent space. I think its an important special case and we could show we can handle it. We could in principle have a simulated data example too where we have many products.

notebooks/Double Machine Learning Examples.ipynb
```diff
@@ -0,0 +1,965 @@
...
...Q0FDc3NzKrXXo0KEYDAZ8fX0xGo2EhIQA4OvrS2JiIgDbt29n1qxZ5ObmkpaWhre3N/fee+8N7cTGxhIdHc2gQYMAMBqNNGvWrNz+hRBCiOp2s9/b1alWha2yRqDMxdvbm3Xr1t2wLTMzk7Nnz+Lh4UFqaip9+vRh48aNxMXF0bt3b0aOHElAQECF2i8pKWH//v3Y2dndsH3atGkMGzaM77//nu7du7Nly5Zy27KxsQHAwsICg8FQ+s1mYWFBcXEx+fn5PPvss0RGRtKyZUumT5/+l9M0aK3x9vZm//79FXoPQgghRE2QU5TD37f+nSd9nqSvW825t1ju2SrHgAEDyM3NLb253Wg08uKLLzJu3Djq1at3w7FeXl68+uqrzJw5s8LtDx48+IaRr6NHjwLXLgn6+vryyiuvEBQUxK+//oqDgwNZWVmVfi+/BytXV1eys7NZu3Zt6b7r227fvj0pKSmlYauoqIjjx49Xul8hhBCiKqyMXsnRy0eru4w/kbBVDqUU69ev56uvvsLT0xMvLy9sbW2ZMWPGXx4/adIkdu3axenTpyvU/sKFC4mMjMTPz49OnTqxZMkSAObPn4+Pjw/+/v7Y2dkxdOhQ/Pz8sLKywt/fn3nz5t3ye3FycuLpp5/G19eXESNGEBwcXLpv3LhxTJo0iYCAAIxGI2vXruWVV17B39+fgIAA9u3bB8D69etxc3Nj//79DBs2jCFDhtxyHUIIIYQ5TPSbyLLBy2rUqBaA0lpXdw2lgoKC9B/nboqJiaFjx47VVJGoDPmaCSGEqCpaa1ZEr2Ck50ga2jas0r6VUoe11kHlHScjW0IIIYSotd4/8j7zo+azMWFjdZdyUxK2hBBCCFErfXL8Ez469hGjvUbzWKfHqrucmzJb2FJKtVdKHb3uT6ZS6nlz9SeEEEKIO8f6+PW8F/keg1oN4o1ub9S46R6uZ7apH7TWsUAAgFLKEjgHrDdXf0IIIYS4MxQZi1h5fCW9mvfi3T7vYmlhWd0llamq5tkaACRorc9UUX9CCCGEqKMMlgZWDFlBPat6WFtaV3c55aqqe7YeAr6oor6EEEIIUQdFXYpi+r7pFJUU4WrnSj1DvfJPqgHMHraUUtbAfcBXN9k/QSkVqZSKTElJMXc5lZKcnMz999+Pp6cnHh4eTJ06lcLCQuDa+oQNGjSgc+fOdOjQgZdeeqn
```
kbattocchi: Could we standardize the Y-axis on these plots?  Otherwise I think it's easy to miss that the slopes and intercepts are quite different.

kbattocchi: I love the idea of adding some test to describe the findings.  I think another big takeaway is that modeling the cross-price effects is important: when we take into account substitution into other brands (as we do in this estimate but not the previous one), then we see that the own-price elasticity is in the -3 to -4 range for all of the brands.  On the other hand, when we did not take that into account the previous estimate was closer to -2.6, so we would have underestimated the effect of price changes on demand.

kbattocchi: Hmm.  What's going on with the Minute Maid to Tropicana plot, where the estimate is way outside the confidence intervals for a large part of the range?  Do we have a bug?  Maybe worth investigating (possibly as a separate issue if you don't have time to look into it now).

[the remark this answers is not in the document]
kbattocchi: Annoyingly these comments don't seem to show the relevant section of the code in the GitHub UI unless you go to the diff for the file...

kbattocchi:
As we discussed over lunch, see what happens if you add `shuffle=True` to the `KFold` constructor in the DML `fit` method and then use many fewer bootstrap samples - does that solve the confidence interval issue?

notebooks/Double Machine Learning Examples.ipynb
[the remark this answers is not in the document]
heimengqi: Adding polynomial featurizer will show more variation with income, but not like Miruna's 'S' shaped curve. In this case since we don't know the true TE, it's hard to say which degree of polynomial features fits best. Another issue here, bootstrap CI works even worse for higher degrees, will dig into it.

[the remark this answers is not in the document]
kbattocchi: I wonder if this would still be true even if you add `shuffle=True` to the cross validation...

notebooks/Double Machine Learning Examples.ipynb
```diff
@@ -0,0 +1,965 @@
...
...s8zMTBITEyVsCSGEqLa01tXuIa0aFbYqOgJlLcnJydjZ2eHt7U1cXBwAjRo1wmg0EhUVxfz58ysUtrTWjB8/nv/+97/XHTt48CDff/89ERERfPrpp3zwwQeltmU0Gku+yQwGQ8kIm8FgoLi4GIC5c+fSsGFDDh48iNlsxsnJ6YZ1LViwgNtvv73cH4sQQghhKzlFOfxt8994uMPD3OZzm63LKSFztirowoULTJ48malTp16XnF999VVmzpyJnZ1dhdocMGAA69atK7ktmZGRwYkTJ7h48SJms5mRI0fyn//8h5iYGABcXV3Jzs6u9MeQmZlJ48aNMRgMrFy5EpPJ9Kft3n777bz77rsUFRUBkJCQQE5OTqX7FUIIIaxpeexyDqQdsHUZ16lRI1u2kpeXR0hICEVFRdjb2zNu3DimT59+3Xk9evS4YRuTJk3iqaeeAqBZs2bs2rWr5FhAQACvvfYagwcPxmw2YzQaiYiIwNnZmYkTJ5YsM/HbyNeECROYPHkyzs7Ov2unvKZMmcLIkSP57LPP6NevH3Xr1gWgY8eO2NvbExwczIQJE5g2bRopKSmEhoaitaZBgwZ8+eWXAPTu3Ztjx45x5coVfHx8WLZsmYyACSGEsKlJHSfRrXE3whpdP6fZltQf5x7ZUlhYmP7tqbrfxMXF0b59extVJG6GfO2EEEJYm9aaD2I/YLjfcOo71a/SvpVS+7XWZSY7uY0ohBBCiBprwc8LmBczj2+SvrF1KTckYUsIIYQQNdJHRz7i/cPvM8p/FOMCxtm6nBuyWthSSrVVSh245r8spdRT1upPCCGEELeO9YnreTv6bQY1H8QLXV+odss9XMtqE+S11vFACIBSyg44Day3Vn9CCCGEuDUUmYpYfmQ5PZv05M3eb2JnqNgqAFWtqp5GHAAkaa1PVFF/QgghhKiljHZGPrj9A+rY18HBzsHW5ZSpquZs3Q98XEV9CSGEEKIWijkfw8s7X6bIXISXsxd1jHVsXVK5WD1sKaUcgLuBz25w/DGlVLRSKvrChQvWLqdS7OzsCAkJITAwkODgYObMmVOy9tWWLVtQSrFs2bKS83/++WeUUsyaNQu4ui7Wb1veWMKKFSs4c+ZMyetHHnmEo0eP3nS7KSkprFmzpszzDhw4QPfu3QkMDKRjx4588sknN923EEIIUZq49Dj+tvlv7D+/n5zC
```
moprescu: Can you un-scale the income? It makes sense to keep the notebooks consistent.

vsyrgkanis: I actually liked the idea of scaled income. I’m thinking we should scale it in the other notebooks for consistency.

moprescu: I just think that it makes the plots the harder to interpret, especially if someone wants to compare this to their knowledge of how this curve is supposed to look like.

heimengqi: Do we even need to scale W? Per my understanding, only distance based model need to be scaled before training? 

Another finding for the ORF notebook, the x-axis label should be "Income" instead of "log(Income)"

PR #28, opened 2019-04-06
Vasilis/local linear

vasilismsr:
Some small updates to the ORF notebook


kbattocchi:
@vasilismsr Can you rebase onto master (after pulling the latest version), and then force push to this branch?  It's showing a bunch of commits that have already been squashed into master.

PR #30, opened 2019-04-08
Refactor DML classes to use a more general RLearner base class

kbattocchi:
This pull request refactors the DML classes, adding an internal base class that should allow us to generalize to "3ML" a la the Orthogonal Statistical Learning paper by Dylan and Vasilis.


kbattocchi:
I've also made two other changes:
* I've added the option to thread random state through this class
* I've added support for discrete treatments

These are visible as separate commits.

PR #43, opened 2019-04-26
fix bug in code

heimengqi:
Keith, Please confirm.


kbattocchi:
What is the problem that you're trying to fix?  Your code is causing our discrete treatment unit test to fail, so either the existing code is okay or else your fix is also not correct (or the test is broken and should also be fixed...).

heimengqi:
Sorry I thought I was wrong...I change "T_train, T_test = T[train_idxs], T_out[test_idxs]" to "T_train, T_test = T_out[train_idxs], T_out[test_idxs]", but I just tried to run it and the original code is right. I will close this PR. Sorry about it.

PR #44, opened 2019-04-29
Mehei/scoring

heimengqi:
add scoring function


econml/dml.py
```diff
@@ -216,21 +236,26 @@ def fit(self, X, T_res, Y_res):
 
                 self._model.fit(cross_product(self._featurizer.fit_transform(X), T_res), Y_res)
 
-            def predict(self, X):
+            def predict(self, X, T_res=None):
```
kbattocchi: Is this extra argument necessary?  How does this differ from multiplying the result of the existing `predict` method (which returns a `d_y × d_t` matrix for each row) by the residuals `T_res`?  I think the calling code should probably just perform that multiplication instead of changing this method to take an extra argument.

econml/dml.py
```diff
@@ -134,6 +134,26 @@ def effect(self, T0, T1, X):
...
+        if Y.ndim == 1:
+            Y = reshape(Y, (-1, 1))
+        if W is None:
+            W = np.empty((shape(Y)[0], 0))
+        Y_test_pred = np.zeros(shape(Y) + (self._n_splits,))
+        T_test_pred = np.zeros(shape(T) + (self._n_splits,))
+        for ind in range(self._n_splits):
+            Y_test_pred[:, :, ind] = reshape(self._models_y[ind].predict(X, W), shape(Y))
+            T_test_pred[:, :, ind] = reshape(self._models_t[ind].predict(X, W), shape(T))
+        Y_test_pred = Y_test_pred.mean(axis=2)
```
kbattocchi: Is the correct measurement to average the predictions first (as you do here) and then calculate the average MSE per row, or calculate the MSE for all first stage models and then average over them after?

If this calculation is the correct one, it may be worth putting in a comment explaining why this is better.

econml/dml.py
```diff
@@ -134,6 +134,26 @@ def effect(self, T0, T1, X):
             T1 = self._one_hot_encoder.transform(reshape(self._label_encoder.transform(T1), (-1, 1)))[:, 1:]
         return super().effect(T0, T1, X)
 
+    def score(self, Y, T, X, W=None):
```
kbattocchi: There need to be tests covering this method.

kbattocchi: Also consider whether to include an example of its use in the notebook.

econml/dml.py
heimengqi: Yes you are right. I tried your suggestion and it gives the same scores.

econml/dml.py
heimengqi: @vasilismsr what do you think? Let me clarify the two options here:

1. first get Y test prediction and T test prediction by averaging the result from models_y and models_t, then get residuals and run final model to get MSE. (The way I did it here.)

2. for each Y test prediction and T test prediction from models_y[i] and models_t[i], get residuals and run final model to get MSE[i], averaging MSE to get final MSE.

vsyrgkanis: Hm. I see what Keith is proposing. It might actually be a better way. But hard to say. I suspect since when we fit the true model we never used the average of the nuisance predictions, but we either used the one or the other nuisance on each data point, then maybe evaluating each nuisance individually and then averaging the qualities is better than averaging the predictions and then deriving a single quality metric for the average prediction. Both seem ok to me at a high level. I don't see any theoretical justification towards one or the other. Both seem correct theoretically from a worst-case performance perspective.

econml/dml.py
heimengqi: In order to make the test pass, I have to accept x=None and W=None, same with the fit function. Can we accept X is None in this scenario?

heimengqi: I will create another issue on it.

kbattocchi: Yes, I think treating `X=None` the same way as the existing methods makes sense (that is, `X=None` will internally use a column of ones for X).

heimengqi: Sure. I agree.

econml/dml.py
heimengqi: I tried both methods on simulation data, the method Kieth proposed always gave a larger MSE (sometimes double the original MSE), but the model sequence is consistent with the original method sorted by MSE. It makes sense since the MSE from single nuisance prediction should be larger than MSE after averaging. I didn't find any case they show different model selection result. It's hard to say which one is better. BTW, on the OJ data, these two give pretty similar MSE.

econml/dml.py
```diff
@@ -134,6 +134,34 @@ def effect(self, T0, T1, X):
...
+                T_test_pred[:, :, ind] = reshape(self._models_t[ind].predict(X, W)[:, 1:], shape(T))
+            else:
+                T_test_pred[:, :, ind] = reshape(self._models_t[ind].predict(X, W), shape(T))
+            Y_test_pred[:, :, ind] = reshape(self._models_y[ind].predict(X, W), shape(Y))
+        Y_test_pred = Y_test_pred.mean(axis=2)
+        T_test_pred = T_test_pred.mean(axis=2)
+        Y_test_res = Y - Y_test_pred
+        T_test_res = T - T_test_pred
+        effects = reshape(self._model_final.predict(X), (-1, shape(Y)[1], shape(T)[1]))
+        Y_test_res_pred = reshape(np.einsum('ijk,ik->ij', effects, T_test_res), shape(Y))
```
vasilismsr: Shouldn't we just call: effect(X) to get the effects, so that we avoid all repeating all the einsum calculations?

vasilismsr:
Maggie, can you also add some cells to the dml notebook, where we use the scoring function to do model selection (e.g. what degree polynomial to use in the featurizer)?

econml/dml.py
heimengqi: Yes you are right, I can get the Y_test_res_pred = self.effect(np.zeros(shape(T)), T_test_res, X), it gave me the same MSE. But it only works for multiple T and multiple Y, because I set slightly different rule  to deal with Y vector and T vector, effect(T0, T1, X) doesn't work for me when T is a vector or Y is a vector. Is that OK i still use my current code?

vasilismsr: Yeap. It's good to stick to the current version.

PR #45, opened 2019-04-29
Pin matplotlib version to fix Python 3.5 builds

kbattocchi:
Matplotlib 3.1 requires Python version >= 3.6; since we don't take a hard dependency on matplotlib (we only use it in tests) we should continue to use older versions so that we can still test against Python 3.5

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ test_suite = econml.tests
 tests_require =
     pytest    
     pytest-xdist
-    matplotlib
+    matplotlib < 3.1
     pandas
     jupyter
 ; TODO: remove matplotlib, pandas dependencies when cleaning up tests


jpweytjens:
If matplotlib is only used for tests, is it possible to move matplotlib to the `tests_require` dependencies? This would enable installing econml in environments where matplotlib > 3.1.

kbattocchi:
@jpweytjens Now matplotlib is also used by the CATE interpreter in non-test code, but I'll see what we can do to improve the experience here.

kbattocchi:
@jpweytjens This will be addressed by #226 and will be part of our v0.7.0 release, which will be completed within the next two weeks.  Thanks for bringing this back to our attention.

PR #48, opened 2019-05-01
Enable setting all Keras fit options (to allow callbacks, etc.)

kbattocchi:


kbattocchi:
This set of changes allows users to pass additional options to Keras's `fit` method when using Deep IV.  Although this would allow passing `{"verbose": 0}` to disable the progress bars in the Deep IV notebook, I haven't done this: although it makes the notebook look cleaner after running it makes it much harder to assess progress during the run.

There's another mostly unrelated commit that should address issues we were hitting where parallel notebook tests were hitting contention on the IPython history - we now pass an additional argument to disable IPython history to avoid this.

vasilismsr:
Looks good

Issue #18, opened 2019-03-22
Allow single-row values for T0 and T1 when computing the effect (with defaults 0 and 1)


kbattocchi:
In many cases (single treatment, single response models with binary treatments or linear responses) setting T0 to 0 and T1 to 1 for all rows of the data would be a sensible default that would save users from having to explicitly provide arrays.

kbattocchi:
@vasilismsr @moprescu 
I think this change will be a nice ergonomic improvement.  However, should we also then reorder the arguments to `effect` so that T0 and T1 come last (i.e. `effect(X=None, T0=0, T1=1)`)?  Otherwise with the current signature the user would have to specify `X` as a keyword argument in the common case where it's not None (e.g. `myModel.effect(X=X)`) because otherwise it would be interpreted as the now optional T0 argument.

vasilismsr:
Yes. I agree! we should be able to say 
model.effect(X) 
for binary treatment models

PR #49, opened 2019-05-01
Reorder effect arguments; allow scalar treatments

kbattocchi:


kbattocchi:
Addresses #18

vasilismsr:
I suspect we might want to simplify our jupyter notebooks so that instead of the uglier: const_marginal_effect(X), we just call effect(X), whenever applicable.

kbattocchi:
@vasilismsr Updated the notebooks as requested.

PR #52, opened 2019-05-02
added the example notebook with bootstrap confidence interval example

ginward:
Added the example notebook with bootstrap confidence interval to the Deep IV notebook.


ginward:
I guess some of these tests failed because of the bootstrapping taking too much time to run ...

PR #56, opened 2019-05-03
Added bootstrap confidence intervals example in Deep IV Notebook

ginward:


ginward:
I added an example of using the bootstrap wrapper to estimate the confidence interval to the Deep IV example notebook. @kbattocchi

kbattocchi:
Thanks for the contribution, but I'm going to hold off on incorporating these changes until after we fix bootstrap with more than one thread - I don't want people to get the wrong idea that bootstrap in general works with Deep IV since such a big limitation currently exists.

ginward:
@kbattocchi I think the problem with the multi-threading is that keras is using the same graph for multiple threads. I guess the workaround is to use multi processing - because multi-threading in python is single threading anyway due to thread locks ...

kbattocchi:
@ginward I agree regarding the problem.  However, I think a better solution would be to have the class clone the m and h models upon initialization, so that whether threading or multiprocessing is used the user doesn't get an error.

I've got a separate set of changes I'm making that will require some minor changes to the `h` argument anyway (to allow a much higher number of samples without a significant performance penalty), so I'll see if I can incorporate the cloning simultaneously.

ginward:
I am not sure it is recommended implementation to use separate graphs in multi-threading. My understanding is that Keras recommends using one graph per process? I might be wrong though, as I am still learning about it.

On 9 May 2019, at 15:50, Keith Battocchi <notifications@github.com<mailto:notifications@github.com>> wrote:


@ginward<https://github.com/ginward> I agree regarding the problem. However, I think a better solution would be to have the class clone the m and h models upon initialization, so that whether threading or multiprocessing is used the user doesn't get an error.

I've got a separate set of changes I'm making that will require some minor changes to the h argument anyway (to allow a much higher number of samples without a significant performance penalty), so I'll see if I can incorporate the cloning simultaneously.

ginward:
> @ginward I agree regarding the problem. However, I think a better solution would be to have the class clone the m and h models upon initialization, so that whether threading or multiprocessing is used the user doesn't get an error.
> 
> I've got a separate set of changes I'm making that will require some minor changes to the `h` argument anyway (to allow a much higher number of samples without a significant performance penalty), so I'll see if I can incorporate the cloning simultaneously.

@kbattocchi From my understanding, python's [GIL](https://stackoverflow.com/a/1294402/5705174) will limit the use of multi-threading, making it effectively single threaded application. So if I am not mistaken, multi-processing is better. But that might require you to change the way you send the tasks to workers, because currently you are sending instances, which is not supported in multi-processing. If you insist on multi-threading, I guess you will have to find a way to create new graphs, which I find quite counter-intuitive ...

PR #57, opened 2019-05-03
Fix effect

kbattocchi:
My last changes to this file broke `effect`, but we weren't calling it anywhere so we didn't pick it up.

### DIFF
--- a/econml/tests/test_deepiv.py
+++ b/econml/tests/test_deepiv.py
@@ -93,6 +93,7 @@ class TestDeepIV(unittest.TestCase):
             deepIv.fit(Y=y, T=t, X=x, Z=z)
             # do something with predictions...
             deepIv.predict(T=t, X=x)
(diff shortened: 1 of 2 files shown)


econml/tests/test_deepiv.py
```diff
@@ -93,6 +93,7 @@ def test_deepiv_shape(self):
             deepIv.fit(Y=y, T=t, X=x, Z=z)
             # do something with predictions...
             deepIv.predict(T=t, X=x)
+            deepIv.effect(x, np.zeros_like(t), t)
```
moprescu: Can we also add a test with no treatments passed in?

kbattocchi: @moprescu Not yet, because apparently I forgot to enable deep IV to take vectors for treatments and responses...

kbattocchi: @moprescu Now addressed by #59

Issue #16, opened 2019-03-22
Ensure that we can use 1D inputs for Y and T


kbattocchi:
Right now, some estimators require that Y and T are 2D arrays, even when they just have a single column.  The library would be more ergonomic if we allowed them to be 1D vectors as well.

PR #59, opened 2019-05-03
Enable deep IV to be used with vectors

kbattocchi:


kbattocchi:
Addresses #16 for Deep IV

PR #76, opened 2019-06-19
Enable internal bootstrap on all estimators

kbattocchi:


kbattocchi:
These changes make it so that every estimator's initializer has a trailing `inference` parameter, which can be set to `None` (the default), in which case everything behaves as before, or to `bootstrap` (or to an instance of `BootstrapOptions` to use non-default bootstrap settings), in which case we will fit a bootstrap estimator to provide confidence intervals.

Going forward, estimators can opt in to other inference methods that may make sense on a case-by-case basis.  This will involve overriding `effect_interval` (and any other `_interval` methods as appropriate), and then implementing the new techinques.

econml/dml.py
```diff
@@ -50,18 +50,23 @@ class _RLearner(LinearCateEstimator):
         If RandomState instance, random_state is the random number generator;
         If None, the random number generator is the RandomState instance used
         by `np.random`.
+
+    inference: string, inference method, or None
+        Method for performing inference.  This estimator supports 'bootstrap'
+        (or an instance of `BootstrapOptions`)
     """
 
     def __init__(self, model_y, model_t, model_final,
```
moprescu: Why do you need to declare at init time whether you are planning to do inference or not?

econml/cate_estimator.py
```diff
@@ -37,7 +85,9 @@ def fit(self, Y, T, X=None, W=None, Z=None):
         self
 
         """
-        pass
+        if hasattr(self, '_bootstrap'):
```
moprescu: Does this mean that it automatically fits the bootstrap CI at fit time? That would be terrible for some of the slow algos, such as the ORF.

econml/tests/test_bootstrap.py
```diff
@@ -150,3 +150,26 @@ def test_backends(self):
                 lower, upper = bs.predict_interval(x, lower=10, upper=90)
                 for bound in [lower, upper]:
                     self.assertEqual(np.shape(est.predict(x)), np.shape(bound))
+
```
moprescu: Can you add some more tests for other estimators?

econml/dml.py
kbattocchi: You need to declare it at init time because then we must fit the replicates when the class itself is fit, which (as you note below) may have a large performance penalty.

For some subclasses we'll be able to support different, more performant options.

econml/tests/test_bootstrap.py
kbattocchi: Yes, that's a good idea.  Will do.

PR #82, opened 2019-07-15
Enable code coverage

kbattocchi:


kbattocchi:
This change turns on automatic code coverage analysis, so that we can more easily determine which lines are covered by tests.

PR #126, opened 2019-10-30
fix score function and add tests

heimengqi:


econml/dml.py
```diff
@@ -244,17 +244,22 @@ def score(self, Y, T, X=None, W=None):
...
+        T_test_pred = np.zeros(shape(T) + (n,))
+        for ind in range(n):
             if self._discrete_treatment:
                 T_test_pred[:, :, ind] = reshape(self._models_t[ind].predict(X, W)[:, 1:], shape(T))
             else:
                 T_test_pred[:, :, ind] = reshape(self._models_t[ind].predict(X, W), shape(T))
             Y_test_pred[:, :, ind] = reshape(self._models_y[ind].predict(X, W), shape(Y))
         Y_test_pred = Y_test_pred.mean(axis=2)
         T_test_pred = T_test_pred.mean(axis=2)
+        if shape(Y) != shape(Y_test_pred):
```
kbattocchi: From the code coverage, it looks like this condition (and the one for T below) is never met during any of our tests, so the following line never executes.  (This is also true of the analagous existing lines in `fit_nuisances`, unfortunately).

Would you mind adding a test (or modifying an existing one) so that these lines are actually necessary?  I think that using `Lasso` as a model and having a vector (rather than 1-column array) target would do it because `Lasso`'s `predict` always returns an array.

kbattocchi: Oops, I had it backwards: if the target is an (n,1) array during fit, predict will return a vector instead.

heimengqi: No worries I got what you meant. Lasso will always return (n,) not matter you input (n,1) or (n,). BTW, I just deleted the two lines of checking code under my function since I have already reshaped the predict output in the previous lines. Seems like there is no need to double check whether the shape matches with each other. But I still changed the test a little bit to cover your checking code.

kbattocchi: Great.  But could you please also use Lasso for model_t when it's not a discrete treatment, to cover the other line I missed?

heimengqi: Yes on it. Forgot model T on first run.

PR #128, opened 2019-10-31
Migrate notebooks to new-style inference

kbattocchi:
Switch the DML notebook to use statsmodels inference (much faster than bootstrap), and move the OrthoForest notebook to use inference in `fit` instead of wrapping everything in a `BootstrapEstimator`.


kbattocchi:
Note that the overall build time isn't sped up too much, because although we reduce the DML notebook time from ~20 minutes to ~4 minutes, the OrthoForest notebook still takes ~19 minutes.  After that, the next longest set of tests only takes ~10 minutes total, so it would be worth seeing whether we could further improve this.

Issue #127, opened 2019-10-31
Bootstrap tests are flaky, unprincipled


kbattocchi:
The `test_bootstrap.TestBootstrap.test_internal_options` test fails sporadically due to its random data generation.

We could fix a single random seed so that the test is deterministic.  However, the test itself is also bad for a few reasons:

1. It tests that the estimated effect is "usually" within the lower and upper bounds of the confidence interval; we should also be testing that the ground truth is within the bounds
2. It is not averaging over several instances of the problem, but over several X, T1, T2 tuples for a single instance.
3. The numerical values were set arbitrarily so that the test usually passes; they should instead be set to a number grounded in theory.

PR #130, opened 2019-11-01
Remove flaky part of flaky test that's affecting builds

kbattocchi:
#127 details some problems with one of our bootstrap tests.  This PR is a temporary measure to enable more consistent builds by simply removing the two flaky parts of the test.  The desired improvements listed in the issue will need to be handled by a separate PR at a later date.

### DIFF
--- a/econml/tests/test_two_stage_least_squares.py
+++ b/econml/tests/test_two_stage_least_squares.py
@@ -40,7 +40,7 @@ class Test2SLS(unittest.TestCase):
 
     @pytest.mark.slow
     def test_hermite_approx(self):
-        n = 500000
+        n = 50000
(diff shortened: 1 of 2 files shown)


kbattocchi:
There was also another test that was occasionally causing build machines to run out of memory, so I've reduced the number of rows we use for that one, too.

econml/tests/test_two_stage_least_squares.py
```diff
@@ -40,7 +40,7 @@ def test_hermite_results(self):
 
     @pytest.mark.slow
     def test_hermite_approx(self):
-        n = 500000
+        n = 50000
```
kbattocchi: This test was occasionally causing failures because it would exhaust the memory of the build machines, so I've reduced the number of rows.

PR #131, opened 2019-11-02
Vasilis/ortho learner

vasilismsr:


econml/_ortho_learner.py
```diff
@@ -0,0 +1,250 @@
...
+
+def _crossfit(model, folds, *args, **kwargs):
+    model_list = []
+    for idx, (train_idxs, test_idxs) in enumerate(folds):
+        model_list.append(clone(model, safe=False))
+
+        args_train = ()
+        args_test = ()
+        for var in args:
+            args_train += (var[train_idxs],)
```
kbattocchi: Shouldn't this also have an `if var is not None:` guard, as with the keyword arguments?

econml/dml.py
```diff
@@ -25,242 +25,7 @@
...
-        Y_test_pred = Y_test_pred.mean(axis=2)
-        T_test_pred = T_test_pred.mean(axis=2)
-        Y_test_res = Y - Y_test_pred
-        T_test_res = T - T_test_pred
-        effects = reshape(self._model_final.predict(X), (-1, shape(Y)[1], shape(T)[1]))
-        Y_test_res_pred = reshape(np.einsum('ijk,ik->ij', effects, T_test_res), shape(Y))
-        mse = ((Y_test_res - Y_test_res_pred)**2).mean()
-        return mse
-
+from ._rlearner import _RLearner
```
kbattocchi: Am I missing something, or is this _rlearner file not present in this set of commits?

PR #137, opened 2019-11-06
Vasilis/drlearner

vasilismsr:


moprescu:
@vasilismsr @heimengqi 
I think we should extend the drlearner to accept multiple discrete treatment. The extension should be straight forward as it's already done in the ORF.

vasilismsr:
The new implementation already does that.

econml/tests/test_drlearner.py
```diff
@@ -0,0 +1,558 @@
...
+
+        ϕ = featurizer.fit_transform(X)
+
+        v_X = cross_product(ϕ, e_t)
+        v_W = cross_product(W, e_t)
+
+        M = np.linalg.pinv(cross_product(WX, hstack([np.ones((shape(WX)[0], 1)), ϕ, W])))
+        e_y = rand_sol(vstack([M, v_X.T]), vstack([-M @ (v_X @ a_X + v_W @ a_W), np.zeros((shape(v_X)[1],))]))
+
+        return e_t, e_y
```
kbattocchi: Delete this since it's not used by any test

econml/tests/test_drlearner.py
```diff
@@ -0,0 +1,558 @@
...
+
+
+# all solutions to underdetermined (or exactly determined) Ax=b are given by A⁺b+(I-A⁺A)w for some arbitrary w
+# note that if Ax=b is overdetermined, this will raise an assertion error
+def rand_sol(A, b):
+    """Generate a random solution to the equation Ax=b."""
+    assert np.linalg.matrix_rank(A) <= len(b)
+    A_plus = np.linalg.pinv(A)
+    x = A_plus @ b
+    return x + (np.eye(x.shape[0]) - A_plus @ A) @ np.random.normal(size=x.shape)
```
kbattocchi: Delete this since it's not used by any test.

econml/utilities.py
```diff
@@ -1468,31 +1468,34 @@ def __init__(self, fit_intercept=True, fit_args={}):
...
             if np.any(np.equal(sample_weight, 1) & np.not_equal(sample_var, 0)):
                 warnings.warn(
-                    """Variance was set to non-zero for an observation with sample_weight=1!
-                       sample_var represents the variance of the original observations that are
-                       summarized in this sample. Hence, cannot have a non-zero variance if only
-                       one observations was summarized. Inference will be invalid!""")
+                    "Variance was set to non-zero for an observation with sample_weight=1! " +
+                    "sample_var represents the variance of the original observations that are " +
+                    "summarized in this sample. Hence, cannot have a non-zero variance if only " +
+                    "one observations was summarized. Inference will be invalid!")
```
kbattocchi: Likewise remove + operators here

econml/utilities.py
```diff
@@ -1468,31 +1468,34 @@ def __init__(self, fit_intercept=True, fit_args={}):
...
             if np.any(np.equal(sample_weight, 1) & np.not_equal(np.sum(sample_var, axis=1), 0)):
                 warnings.warn(
-                    """Variance was set to non-zero for an observation with sample_weight=1!
-                       sample_var represents the variance of the original observations that are
-                       summarized in this sample. Hence, cannot have a non-zero variance if only
-                       one observations was summarized. Inference will be invalid!""")
+                    "Variance was set to non-zero for an observation with sample_weight=1! " +
+                    "sample_var represents the variance of the original observations that are " +
+                    "summarized in this sample. Hence, cannot have a non-zero variance if only " +
+                    "one observations was summarized. Inference will be invalid!")
```
kbattocchi: Likewise remove + operators here

econml/tests/test_drlearner.py
```diff
@@ -0,0 +1,558 @@
...
+                while True:
+                    arr = np.random.choice(['a', 'b', 'c'], size=sz)
+                    # ensure that we've got at least two of every element
+                    _, counts = np.unique(arr, return_counts=True)
+                    if len(counts) == 3 and counts.min() > 1:
+                        return arr
+            else:
+                return np.random.normal(size=sz)
+
+        d_y = 0
```
kbattocchi: Do we not support y arrays with shape (n,1)?  I think we should...

econml/tests/test_drlearner.py
```diff
@@ -0,0 +1,558 @@
...
+        is_discrete = True
+        d_t = 1
```
kbattocchi: Do we not support vectors for T?  I think we should...

econml/tests/test_drlearner.py
```diff
@@ -0,0 +1,558 @@
...
+                        if d_t > 1:
+                            cm = self.assertWarns(Warning)
+                        else:
+                            cm = ExitStack()  # ExitStack can be used as a "do nothing" ContextManager
+                        with cm:
+                            effect_shape2 = (n if d_x else 1,) + ((d_y,) if d_y > 0 else())
+                            eff = est.effect(X, T0='a', T1='b')
+                            self.assertEqual(shape(eff), effect_shape2)
+
+    def test_can_use_vectors(self):
```
kbattocchi: Can you add a comment to each test which is virtually identical to a DML test so that we can consolidate them into generic tests in a later pass?

econml/tests/test_drlearner.py
```diff
@@ -0,0 +1,558 @@
...
+                            for models in [(GradientBoostingClassifier(), GradientBoostingRegressor(),
+                                            RandomForestRegressor(n_estimators=100, max_depth=5, min_samples_leaf=50)),
+                                           (GradientBoostingClassifier(), GradientBoostingRegressor(),
+                                            RandomForestRegressor(n_estimators=100, max_depth=5, min_samples_leaf=50)),
+                                           (LogisticRegression(solver='lbfgs', multi_class='auto'),
+                                            LinearRegression(), StatsModelsLinearRegression())]:
+                                for multitask_model_final in [False, True]:
+                                    if (not isinstance(models, StatsModelsLinearRegression)) and (sample_var is not None):
+                                        continue
+                                    print(X, W, sample_weight, sample_var, featurizer, models, multitask_model_final)
```
kbattocchi: Any benefit to printing?

econml/tests/test_drlearner.py
```diff
@@ -0,0 +1,558 @@
...
+        interval = dml.coef__interval(T=2, alpha=0.05)
+        point = dml.coef_(T=2)
+        assert len(interval) == 2
+        lo, hi = interval
+        assert lo.shape == hi.shape == point.shape
+        assert (lo <= point).all()
+        assert (point <= hi).all()
+        assert (lo < hi).any()  # for at least some of the examples, the CI should have nonzero width
+
+    def test_drlearner_all_attributes(self):
```
kbattocchi: Is there some overlap with the CATE API test?  Could the two tests be consolidated?

econml/tests/test_drlearner.py
```diff
@@ -0,0 +1,558 @@
...
+                                                    est.model_cate(T=t).feature_importances_)[-1], 0)
+                                            else:
+                                                true_coef = np.zeros(len(out_feat_names))
+                                                true_coef[0] = .5 * t
+                                                np.testing.assert_allclose(
+                                                    est.model_cate(T=t).coef_, true_coef, rtol=0, atol=.15)
+                                                np.testing.assert_allclose(
+                                                    est.model_cate(T=t).intercept_, t, rtol=0, atol=.15)
+
+    def test_linear_drlearner_all_attributes(self):
```
kbattocchi: Is there a reason to have separate tests for DRLearner vs. LinearDRLearner?  Looks like a lot of duplicated code.

econml/inference.py
```diff
@@ -131,3 +131,80 @@ def coef__interval(self, *, alpha=0.1):
 
     def intercept__interval(self, *, alpha=0.1):
         return self.statsmodels.intercept__interval(alpha)
+
+
+class StatsModelsInferenceDiscrete(Inference):
```
kbattocchi: Can we create a common base class for StatsModelsInference and StatsModelsInferenceDiscrete?  Looks like almost everything is extremely similar and I'd rather not have so much duplicated code, because if we change it in one place we need to remember to change the parallel copy, too.

econml/cate_estimator.py
```diff
@@ -348,3 +408,94 @@ def coef__interval(self, *, alpha=0.1):
     @BaseCateEstimator._defer_to_inference
     def intercept__interval(self, *, alpha=0.1):
         pass
+
+
+class StatsModelsCateEstimatorDiscreteMixin(BaseCateEstimator):
+
+    def _get_inference_options(self):
```
kbattocchi: Could we ever have an estimator that supports both discrete and continuous statsmodels inference (depending on the treatments it's applied to)?  If so, is the fact that each mixin sets the `statsmodels` dictionary in a mutually exclusive way a problem?

econml/cate_estimator.py
```diff
@@ -348,3 +408,94 @@ def coef__interval(self, *, alpha=0.1):
...
+
+        Returns
+        -------
+        coef: (n_x,) or (n_y, n_x) array like
+            Where n_x is the number of features that enter the final model (either the
+            dimension of X or the dimension of featurizer.fit_transform(X) if the CATE
+            estimator has a featurizer.)
+        """
+        _, T = self._expand_treatments(None, T)
+        ind = (T @ np.arange(1, T.shape[1] + 1)).astype(int)[0] - 1
```
kbattocchi: Add a comment explaining that this is just un-one-hot-encoding T.

kbattocchi: Why the `astype(int)`?

kbattocchi: Aren't you unnecessarily adding and subtracting 1?
```suggestion
        ind = (T @ np.arange(T.shape[1])).astype(int)[0]
```

econml/cate_estimator.py
```diff
@@ -348,3 +408,94 @@ def coef__interval(self, *, alpha=0.1):
...
+        ----------
+        T: alphanumeric
+            The input treatment for which we want the coefficients.
+        alpha: optional float in [0, 1] (Default=0.1)
+            The overall level of confidence of the reported interval.
+            The alpha/2, 1-alpha/2 confidence interval is reported.
+
+        Returns
+        -------
+        lower, upper: tuple(type of :attr:`coef_(T)`, type of :attr:`coef_(T)`)
```
kbattocchi: I don't think this style of cross ref works, I believe it should be
```suggestion
        lower, upper: tuple(type of :meth:`coef_(T)<coef_>`, type of :meth:`coef_(T)<coef>`)
```

econml/cate_estimator.py
```diff
@@ -348,3 +408,94 @@ def coef__interval(self, *, alpha=0.1):
...
+        ----------
+        T: alphanumeric
+            The input treatment for which we want the coefficients.
+        alpha: optional float in [0, 1] (Default=0.1)
+            The overall level of confidence of the reported interval.
+            The alpha/2, 1-alpha/2 confidence interval is reported.
+
+        Returns
+        -------
+        lower, upper: tuple(type of :attr:`intercept_(T)`, type of :attr:`intercept_(T)`)
```
kbattocchi: Likewise
```suggestion
        lower, upper: tuple(type of :meth:`intercept_(T)<intercept_>`, type of :meth:`intercept_(T)<intercept_>`)
```

econml/cate_estimator.py
```diff
@@ -348,3 +408,94 @@ def coef__interval(self, *, alpha=0.1):
     @BaseCateEstimator._defer_to_inference
     def intercept__interval(self, *, alpha=0.1):
         pass
+
+
+class StatsModelsCateEstimatorDiscreteMixin(BaseCateEstimator):
```
kbattocchi: Consider refactoring to share some logic with `StatsModelsCateEstimatorMixin`

econml/tests/test_drlearner.py
vasilismsr: So that when it crushes I know for which combination of variables it crushed.

econml/tests/test_drlearner.py
vasilismsr: At some point. I don't know what the CATE API test is exactly testing. I suspect we should have this here too. Double testing is better than no testing and at some point we can clean up. But there might be intricacies with each module and some combinations not caught by general CATE API tests. We should have though general CATE API tests either way to have a global minimal guarantee across modules

econml/tests/test_drlearner.py
vasilismsr: Yeap. The two have different input params and diffeerent attributes after fitting.

vasilismsr: Also the drlearner has no interval capabilities

econml/inference.py
vasilismsr: I would move it in a future PR

econml/tests/test_drlearner.py
kbattocchi: You can use `self.subTest` for that.  See the DML test for the CATE API.

econml/cate_estimator.py
vasilismsr: I would move it to a future PR. It might be problematic. Because for DML with discrete treatments you should use StatsModelsInference, but with DRLearner you should use StatsModelsInferenceDiscrete. So the user would need to know when to set the correspnding parameter of the merged class, which might be weird. Now the docstring points to the right class. We could in the back end have a parent class of both classes, but not change the front end of two classes.

econml/tests/test_drlearner.py
kbattocchi: Okay.  At some point I'd like to consolidate the repeated logic but that can be done later.

econml/tests/test_drlearner.py
vasilismsr: Not for now I am throwing an error, and I test that the error is correctly raised. We can amend in a future PR. I'll add a TODO comment.

econml/inference.py
kbattocchi: Ok. #143

econml/tests/test_drlearner.py
vasilismsr: We do. amended test to test for that

econml/cate_estimator.py
vasilismsr: I would again similarly push it to a future PR.

kbattocchi: Ok

econml/cate_estimator.py
kbattocchi: Ok.  No need to worry too much about such complications at this stage I guess.

econml/drlearner.py
```diff
@@ -0,0 +1,633 @@
...
+            def fit(self, Y, T, X=None, W=None, *, sample_weight=None):
+                assert np.ndim(Y) == 1, "Can only accept single dimensional outcomes Y! Use Y.ravel()."
+                if (X is None) and (W is None):
+                    raise AttributeError("At least one of X or W has to not be None!")
+                if np.any(np.all(T == 0, axis=0)) or (not np.any(np.all(T == 0, axis=1))):
+                    raise AttributeError("Provided crossfit folds contain training splits that " +
+                                         "don't contain all treatments")
+                XW = self._combine(X, W)
+                filtered_kwargs = _filter_none_kwargs(sample_weight=sample_weight)
+                self._model_propensity.fit(XW, np.matmul(T, np.arange(1, T.shape[1] + 1)).ravel().astype(int),
```
kbattocchi: Consider creating a utility method for untransforming one-hot treatments - similar logic seems to have been put into many different places.

econml/drlearner.py
```diff
@@ -0,0 +1,633 @@
...
+                return Y_pred
+
+        class ModelFinal:
+            def __init__(self, model_final, featurizer, multitask_model_final):
+                self._model_final = clone(model_final, safe=False)
+                self._featurizer = clone(featurizer, safe=False)
+                self._multitask_model_final = multitask_model_final
+                return
+
+            def _filter_none_kwargs(self, **kwargs):
```
kbattocchi: Isn't this the same as the module-level function at the top of the file?

econml/drlearner.py
```diff
@@ -0,0 +1,633 @@
...
+                out_kwargs = {}
+                for key, value in kwargs.items():
+                    if value is not None:
+                        out_kwargs[key] = value
+                return out_kwargs
+
+            def fit(self, Y, T, X=None, W=None, nuisances=None, *, sample_weight=None, sample_var=None):
```
kbattocchi: Seems like `None` is not a valid value for nuisances (the first line of this method will throw).  Can you make it keyword-only to make this more obvious?
```suggestion
            def fit(self, Y, T, X=None, W=None, *, nuisances, sample_weight=None, sample_var=None):
```

econml/drlearner.py
```diff
@@ -0,0 +1,633 @@
...
+
+            def predict(self, X=None):
+                if (X is not None) and (self._featurizer is not None):
+                    X = self._featurizer.transform(X)
+                if self._multitask_model_final:
+                    return self.model_cate.predict(X)
+                else:
+                    return np.array([mdl.predict(X) for mdl in self.models_cate]).T
+
+            def score(self, Y, T, X=None, W=None, nuisances=None, *, sample_weight=None, sample_var=None):
```
kbattocchi: Again, consider making nuisances keyword-only

```suggestion
            def score(self, Y, T, X=None, W=None, *, nuisances, sample_weight=None, sample_var=None):
```

econml/drlearner.py
```diff
@@ -0,0 +1,633 @@
...
+                if (X is not None) and (self._featurizer is not None):
+                    X = self._featurizer.transform(X)
+                Y_pred, = nuisances
+                if sample_weight is None:
```
kbattocchi: `np.average` already handles `weights=None`, so no need to do this

econml/drlearner.py
```diff
@@ -0,0 +1,633 @@
...
+                    if value is not None:
+                        out_kwargs[key] = value
+                return out_kwargs
+
+            def fit(self, Y, T, X=None, W=None, nuisances=None, *, sample_weight=None, sample_var=None):
+                Y_pred, = nuisances
+                if (X is not None) and (self._featurizer is not None):
+                    X = self._featurizer.fit_transform(X)
+                filtered_kwargs = _filter_none_kwargs(sample_weight=sample_weight, sample_var=sample_var)
+                if self._multitask_model_final:
```
kbattocchi: Could the same result be achieved by wrapping the model with the MultitaskModelWrapper?

econml/drlearner.py
```diff
@@ -0,0 +1,633 @@
...
+        score: float
+            The MSE of the final CATE model on the new data.
+        """
+        # Replacing score from _OrthoLearner, to enforce Z=None and improve the docstring
+        return super().score(Y, T, X=X, W=W)
+
+    @property
+    def multitask_model_cate(self):
+        if not self._multitask_model_final:
+            raise AttributeError("Separte CATE models were fitted for each treatment! Use model_cate.")
```
kbattocchi: Spelling:
```suggestion
            raise AttributeError("Separate CATE models were fitted for each treatment! Use model_cate.")
```

econml/drlearner.py
```diff
@@ -0,0 +1,633 @@
...
+
+    Attributes
+    ----------
+    models_propensity: list of objects of type(model_propensity)
+        A list of instances of the model_propensity object. Each element corresponds to a crossfitting
+        fold and is the model instance that was fitted for that training fold.
+    models_regression: list of objects of type(model_regression)
+        A list of instances of the model_regression object. Each element corresponds to a crossfitting
+        fold and is the model instance that was fitted for that training fold.
+    model_cate(T=t) : object of type(model_final)
```
kbattocchi: This is a method, not an attribute...  Why not put the docstring on the method itself?

Likewise, many of the other things are properties rather than attributes, as well.

econml/drlearner.py
vasilismsr: Good idea. tried it but failed. StatsModelsLinearRegression does not have get and set_params and hence cannot be wrapped by sklearn multioutputregressor. So until we make statsModelsLinearRegression comply with sklearn estimators (which could just be as simple as making the parent of that class be sklearn's BaseEstimator. But I'll leave it for a future PR. Please create an issue.

econml/drlearner.py
vasilismsr: What I wanted to follow here is that properties that can be called only after fitted, I tried to make them attributes.

econml/drlearner.py
vasilismsr: I actually just also tried this but still it fails. So now StatsModelsLinearRegression is a BaseEstimator and can be wrapped by MultiOutputRegressor. But the problem is that I was allowing for StatsModelsLinearRegression to take as input X=None and this is not allowed by the MutliOutputRegressor. This then needs a lot of code refactoring, which I don't think is worth.

kbattocchi: Ok, thanks for giving it a shot.  Can you add a short comment to that effect to the code so that it's easier to know why this approach was taken?

econml/drlearner.py
kbattocchi: I believe that in docs like https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html "Attribute" has a specific technical meaning: attributes are things that have been set directly on an instance's dictionary, like `self.coef_ = ...` (although admittedly this distinction is confusing, because the python spec and other sources describe all members as attributes, so really all properties are attributes, too).

econml/drlearner.py
vasilismsr: Done. Comment added.

econml/drlearner.py
vasilismsr: That makes sense. I'll revert to docstrings for these ones here, except score_.

PR #141, opened 2019-11-07
support multi treatment in meta learners

heimengqi:
1. extend the meta learners support multiple treatments
2. remove DRLearner
3. change tests and notebook accordingly


econml/metalearners.py
```diff
@@ -83,10 +89,16 @@ def effect(self, X):
...
+        if T0 not in self.unique_T or T1 not in self.unique_T:
+            raise ValueError(
+                "T0 and T1 must be scalers in T you passed in at fitting time, please pass in values from {}".format(self.unique_T))
+        ind1, = np.where(self.unique_T == T1)[0]
+        ind0, = np.where(self.unique_T == T0)[0]
+        tau_hat = self.models[ind1].predict(X) - self.models[ind0].predict(X)
         return tau_hat
 
-    def marginal_effect(self, X):
+    def marginal_effect(self, X, T1):
```
kbattocchi: When a method takes only one treatment, probably better to just call it `T`.

```suggestion
    def marginal_effect(self, X, T):
```

econml/metalearners.py
```diff
@@ -95,13 +107,14 @@ def marginal_effect(self, X):
         ----------
         X : matrix, shape (m × dₓ)
             Matrix of features for each sample.
+        T1: scaler
```
kbattocchi:
```suggestion
        T: scalar
```

Again, consider allowing arrays for `T`.

econml/metalearners.py
```diff
@@ -95,13 +107,14 @@ def marginal_effect(self, X):
...
 
         Returns
         -------
         τ_hat : array-like, shape (m, )
```
kbattocchi: This looks wrong to me: the marginal effect should have shape (m, n_t-1), like it does for discrete treatments in DML

econml/metalearners.py
```diff
@@ -95,13 +107,14 @@ def marginal_effect(self, X):
...
             Matrix of heterogeneous treatment effects for each sample.
         """
-        return self.effect(X)
+        return self.effect(X, T0=self.unique_T[0], T1=T1)
```
kbattocchi: I don't think this logic is correct due to the shape requirements.  I think maybe a better approach would be to inherit from `LinearCateEstimator` and implement `const_marginal_effect` (and completely remove `marginal_effect` and `effect`, since those are defined in terms of it).

econml/metalearners.py
```diff
@@ -145,12 +159,11 @@ def fit(self, Y, T, X, inference=None):
...
         Y, T, X, _ = check_inputs(Y, T, X, multi_output_T=False)
-        if not np.array_equal(np.unique(T), [0, 1]):
-            raise ValueError("The treatments array (T) can only contain 0 and 1.")
-        feat_arr = np.concatenate((X, T.reshape(-1, 1)), axis=1)
+        T = self._one_hot_encoder.fit_transform(T.reshape(-1, 1))[:, 1:]
+        feat_arr = np.concatenate((X, T), axis=1)
         self.overall_model.fit(feat_arr, Y)
 
-    def effect(self, X):
+    def effect(self, X, T0, T1):
```
kbattocchi: Please add these to the Parameters section of the docstring.

econml/metalearners.py
```diff
@@ -185,7 +199,8 @@ def marginal_effect(self, X):
         τ_hat : array-like, shape (m, )
```
kbattocchi: Again this has the wrong shape for multiple treatments

econml/metalearners.py
```diff
@@ -165,12 +178,13 @@ def effect(self, X):
...
-        X_controls = np.concatenate((X, np.zeros((X.shape[0], 1))), axis=1)
-        X_treated = np.concatenate((X, np.ones((X.shape[0], 1))), axis=1)
+        m = X.shape[0]
+        X_controls = np.concatenate((X, self._one_hot_encoder.transform(np.ones((m, 1)) * T0)[:, 1:]), axis=1)
+        X_treated = np.concatenate((X, self._one_hot_encoder.transform(np.ones((m, 1)) * T1)[:, 1:]), axis=1)
         tau_hat = self.overall_model.predict(X_treated) - self.overall_model.predict(X_controls)
         return tau_hat
 
-    def marginal_effect(self, X):
+    def marginal_effect(self, X, T1):
```
kbattocchi:
```suggestion
    def marginal_effect(self, X, T):
```

kbattocchi: Add `T` to the parameters

econml/metalearners.py
```diff
@@ -185,7 +199,8 @@ def marginal_effect(self, X):
         τ_hat : array-like, shape (m, )
             Matrix of heterogeneous treatment effects for each sample.
         """
-        return self.effect(X)
+        T0 = self._one_hot_encoder.categories_[0][0]
+        return self.effect(X, T0=T0, T1=T1)
```
kbattocchi: Again this would probably be better defined by the LinearCateEstimator's const_marginal_effect, although it's not obvious that the idea of a marginal effect really makes sense for the SLearner.  Maybe something to dig into during our SDK sync.

econml/metalearners.py
```diff
@@ -185,7 +199,8 @@ def marginal_effect(self, X):
...
 
 
 class XLearner(BaseCateEstimator):
```
kbattocchi: The Parameters section is out of date; please update it to reflect the new models and cate_models parameters

econml/metalearners.py
```diff
@@ -265,20 +263,34 @@ def fit(self, Y, T, X, inference=None):
...
+            self.propensity_models.append(clone(self.propensity_model, safe=False))
+            imputed_effect_on_controls = self.models[ind].predict(X[T == 0]) - Y[T == 0]
+            imputed_effect_on_treated = Y[T == ind] - self.models[0].predict(X[T == ind])
+            self.cate_controls_models[ind - 1].fit(X[T == 0], imputed_effect_on_controls)
+            self.cate_treated_models[ind - 1].fit(X[T == ind], imputed_effect_on_treated)
+            X_concat = np.concatenate((X[T == 0], X[T == ind]), axis=0)
+            T_concat = np.concatenate((T[T == 0], T[T == ind]), axis=0)
+            self.propensity_models[ind - 1].fit(X_concat, T_concat)
+
+    def effect(self, X, T0, T1):
```
kbattocchi: Document parameters

econml/metalearners.py
```diff
@@ -292,13 +304,12 @@ def effect(self, X):
...
-        propensity_scores = self.propensity_func(X)
-        tau_hat = propensity_scores * self.cate_controls_model.predict(X) \
-            + (1 - propensity_scores) * self.cate_treated_model.predict(X)
+        if T0 == self.unique_T[0]:
+            return self.marginal_effect(X, T1)
+        tau_hat = self.marginal_effect(X, T1) - self.marginal_effect(X, T0)
         return tau_hat
 
-    def marginal_effect(self, X):
+    def marginal_effect(self, X, T1):
```
kbattocchi: Again, I don't think the output shape/logic is correct.

econml/metalearners.py
```diff
@@ -313,7 +324,12 @@ def marginal_effect(self, X):
...
-        return self.effect(X)
+        X = check_array(X)
+        ind, = np.where(self.unique_T == T1)[0]
+        propensity_scores = self.propensity_models[ind - 1].predict_proba(X)[:, 1]
+        tau_hat = propensity_scores * self.cate_controls_models[ind - 1].predict(X) \
+            + (1 - propensity_scores) * self.cate_treated_models[ind - 1].predict(X)
+        return tau_hat
 
 
 class DomainAdaptationLearner(BaseCateEstimator):
```
kbattocchi: Docstring needs to be updated to reflect models, final_models parameters.

econml/metalearners.py
```diff
@@ -384,29 +391,39 @@ def fit(self, Y, T, X, inference=None):
...
+            # control units.
+            self._fit_weighted_pipeline(self.models_treated[ind - 1], X[T == ind], Y[T == ind],
+                                        sample_weight=(1 - propensity_scores[T_concat == ind]) / propensity_scores[T_concat == ind])
+            imputed_effect_on_controls = self.models_treated[ind - 1].predict(X[T == 0]) - Y[T == 0]
+            imputed_effect_on_treated = Y[T == ind] - self.models_control[ind - 1].predict(X[T == ind])
+
+            imputed_effects_concat = np.concatenate((imputed_effect_on_controls, imputed_effect_on_treated), axis=0)
+            self.final_models[ind - 1].fit(X_concat, imputed_effects_concat)
+
+    def effect(self, X, T0, T1):
```
kbattocchi: Add new parameters to docstring

econml/metalearners.py
```diff
@@ -439,141 +457,14 @@ def marginal_effect(self, X):
         τ_hat : array-like, shape (m, )
```
kbattocchi: Again, wrong shape for multple treatments

econml/metalearners.py
```diff
@@ -62,19 +64,23 @@ def fit(self, Y, T, X, inference=None):
...
 
-    def effect(self, X):
+    def effect(self, X, T0, T1):
         """Calculate the heterogeneous treatment effect on a vector of features for each sample.
 
         Parameters
         ----------
         X : matrix, shape (m × dₓ)
             Matrix of features for each sample.
+        T0: scaler
```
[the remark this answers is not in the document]
heimengqi: Yes, I didn't update the doc yet. it's a draft, and I push it want to see the code coverage. Yes I will check the DML effect interface as a reference.

heimengqi:
Still need to write a more comprehensive test includes testing multi Y, array Y or column Y, like the test for DML.

econml/metalearners.py
```diff
@@ -51,60 +57,53 @@ def fit(self, Y, T, X, inference=None):
...
         # Check inputs
         Y, T, X, _ = check_inputs(Y, T, X, multi_output_T=False)
-        if not np.array_equal(np.unique(T), [0, 1]):
-            raise ValueError("The treatments array (T) can only contain" +
-                             "0 and 1.")
-        self.controls_model.fit(X[T == 0], Y[T == 0])
-        self.treated_model.fit(X[T == 1], Y[T == 1])
+        T = self._label_encoder.fit_transform(T)
+        self._one_hot_encoder.fit(T.reshape(-1, 1))
+        self._d_t = len(self._label_encoder.classes_)
```
kbattocchi: I think this should be `len(...) - 1`, but it might not matter too much for this particular subclass since it doesn't support the things that make a distinction (like statsmodels inference).

econml/metalearners.py
```diff
@@ -117,6 +116,13 @@ class SLearner(BaseCateEstimator):
...
+        self._label_encoder = LabelEncoder()
+        self._one_hot_encoder = OneHotEncoder(categories='auto', sparse=False)
+        self.transformer = FunctionTransformer(
+            func=(lambda T:
+                  self._one_hot_encoder.transform(
+                      self._label_encoder.transform(T).reshape(-1, 1))[:, 1:]),
+            validate=False)
         super().__init__()
 
     @BaseCateEstimator._wrap_fit
```
kbattocchi: Consider making X optional (and treating it as a 0-column array if it is None).

Also, make inference keyword only.

econml/metalearners.py
```diff
@@ -141,17 +147,19 @@ def fit(self, Y, T, X, inference=None):
...
+        self : an instance of self.
         """
         # Check inputs
         Y, T, X, _ = check_inputs(Y, T, X, multi_output_T=False)
-        if not np.array_equal(np.unique(T), [0, 1]):
-            raise ValueError("The treatments array (T) can only contain 0 and 1.")
-        feat_arr = np.concatenate((X, T.reshape(-1, 1)), axis=1)
+        T = self._label_encoder.fit_transform(T)
+        T = self._one_hot_encoder.fit_transform(T.reshape(-1, 1))
+        self._d_t = T.shape[1:]
```
kbattocchi: Again I would expect this to be `(T.shape[1] - 1,)` for consistency with the rest of the library.

econml/metalearners.py
```diff
@@ -141,17 +147,19 @@ def fit(self, Y, T, X, inference=None):
...
+        self._d_y = Y.shape[1:]
```
kbattocchi: Again `BaseCateEstimator` should already be doing this.

econml/metalearners.py
kbattocchi: (and for T to omit its first column)

econml/metalearners.py
```diff
@@ -160,83 +168,59 @@ def effect(self, X):
...
-    controls_model : outcome estimator for control units
-        Must implement `fit` and `predict` methods.
-
-    treated_model : outcome estimator for treated units
-        Must implement `fit` and `predict` methods.
-
-    cate_controls_model : estimator for pseudo-treatment effects on the controls
+    models : outcome estimators for both control units and treatment units
+        It can be a single estimator applied to all the control and treatment units or a tuple/list of
+        estimators with the same number of treatments (include control).
```
kbattocchi:
```suggestion
        estimators with one estimator per treatment (including control).
```

econml/metalearners.py
```diff
@@ -160,83 +168,59 @@ def effect(self, X):
...
         Must implement `fit` and `predict` methods.
 
-    cate_treated_model : estimator for pseudo-treatment effects on the treated
+    cate_models : estimator for pseudo-treatment effects on control and treatments
+        It can be a single estimator applied to all the control and treatments or a tuple/list of
+        estimators with the same number of treatments (include control).
```
kbattocchi:
```suggestion
        estimators with one estimator per treatment (including control).
```

econml/metalearners.py
```diff
@@ -160,83 +168,59 @@ def effect(self, X):
...
+        If None, it will be same models with outcome estimators.
```
kbattocchi:
```suggestion
        If None, it will be same models as the outcome estimators.
```

econml/metalearners.py
```diff
@@ -255,31 +239,46 @@ def fit(self, Y, T, X, inference=None):
...
-        if not self.has_propensity_func:
-            self.propensity_model.fit(X, T)
-            self.propensity_func = lambda X_score: self.propensity_model.predict_proba(X_score)[:, 1]
-
-    def effect(self, X):
-        """Calculate the heterogeneous treatment effect on a vector of features for each sample.
+        T = self._label_encoder.fit_transform(T)
+        self._one_hot_encoder.fit(T.reshape(-1, 1))
+        self._d_t = len(self._label_encoder.classes_)
+        self._d_y = Y.shape[1:]
```
kbattocchi: Same considerations here as on the other metalearners.

econml/metalearners.py
```diff
@@ -288,74 +287,61 @@ def effect(self, X):
...
     """Meta-algorithm that uses domain adaptation techniques to account for
-       covariate shift (selection bias) between the treatment arms.
+       covariate shift (selection bias) among the treatment arms.
 
     Parameters
     ----------
-    controls_model : outcome estimator for control units
+    models : outcome estimators for both control units and treatment units
+        It can be a single estimator applied to all the control and treatment units or a tuple/list of
+        estimators with the same number of treatments (include control).
```
kbattocchi:
```suggestion
        estimators with one estimator per treatment (including control).
```

econml/metalearners.py
```diff
@@ -288,74 +287,61 @@ def effect(self, X):
...
         The `fit` method must accept the `sample_weight` parameter.
 
-    treated_model : outcome estimator for treated units
-        Must implement `fit` and `predict` methods.
-        The `fit` method must accept the `sample_weight` parameter.
-
-    overall_model : estimator for pseudo-treatment effects
+    final_models : estimators for pseudo-treatment effects for each treatment
+        It can be a single estimator applied to all the control and treatment units or a tuple/list of
+        estimators with the same number of treatments (exclude control).
```
kbattocchi:
```suggestion
        estimators with one estimator per treatment (excluding control).
```

econml/metalearners.py
```diff
@@ -374,40 +360,51 @@ def fit(self, Y, T, X, inference=None):
...
-        X_concat = np.concatenate((X[T == 0], X[T == 1]), axis=0)
-        imputed_effects_concat = np.concatenate((imputed_effect_on_controls, imputed_effect_on_treated), axis=0)
-        self.overall_model.fit(X_concat, imputed_effects_concat)
-
-    def effect(self, X):
-        """Calculate the heterogeneous treatment effect on a vector of features for each sample.
+        T = self._label_encoder.fit_transform(T)
+        self._one_hot_encoder.fit(T.reshape(-1, 1))
+        self._d_t = len(self._label_encoder.classes_)
+        self._d_y = Y.shape[1:]
```
kbattocchi: Same questions as for other metalearners.

econml/metalearners.py
```diff
@@ -416,164 +413,23 @@ def effect(self, X):
...
-            Matrix of heterogeneous treatment effects for each sample.
-        """
-        return self.effect(X)
+        taus = []
+        for model in self.final_models:
+            taus.append(model.predict(X))
+        taus = np.column_stack(taus).reshape((-1, self._d_t - 1,) + self._d_y)  # shape as of m*d_t*d_y
+        if self._d_y:
+            taus = transpose(taus, (0, 2, 1))  # shape as of m*d_y*d_t
+        return taus
```
kbattocchi: Looks like very similar logic to this shows up in a few places.  Would it be worthwhile to create a common base class so that the logic doesn't need to be repeated?

econml/metalearners.py
```diff
@@ -51,60 +57,53 @@ def fit(self, Y, T, X, inference=None):
...
         Y, T, X, _ = check_inputs(Y, T, X, multi_output_T=False)
-        if not np.array_equal(np.unique(T), [0, 1]):
-            raise ValueError("The treatments array (T) can only contain" +
-                             "0 and 1.")
-        self.controls_model.fit(X[T == 0], Y[T == 0])
-        self.treated_model.fit(X[T == 1], Y[T == 1])
+        T = self._label_encoder.fit_transform(T)
+        self._one_hot_encoder.fit(T.reshape(-1, 1))
+        self._d_t = len(self._label_encoder.classes_)
+        self._d_y = Y.shape[1:]
```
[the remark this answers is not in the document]
heimengqi: Yes! I didn't realize that how it worked. Thanks for pointing it out!

econml/metalearners.py
heimengqi: Yes my original thought was d_t represents number of class in multi treatment settings, finally const_marginal_effect will return m*d_y*d_t-1. But you are right in order for consistency with the rest of the library, I will change it accordingly.

econml/metalearners.py
```diff
@@ -150,39 +149,42 @@ def fit(self, Y, T, X, inference=None):
...
         """
         # Check inputs
+        if X is None:
+            X = np.ones((Y.shape[0], 1))
         Y, T, X, _ = check_inputs(Y, T, X, multi_output_T=False)
         T = self._label_encoder.fit_transform(T)
         T = self._one_hot_encoder.fit_transform(T.reshape(-1, 1))
-        self._d_t = T.shape[1:]
-        self._d_y = Y.shape[1:]
+        self._d_t = T.shape[1] - 1
```
kbattocchi: To be consistent with other classes, this should probably always be a 1-tuple rather than a raw value.  That is,
```suggestion
        self._d_t = T.shape[1] - 1,
```

econml/metalearners.py
```diff
@@ -150,39 +149,42 @@ def fit(self, Y, T, X, inference=None):
         self : an instance of self.
         """
         # Check inputs
+        if X is None:
+            X = np.ones((Y.shape[0], 1))
```
kbattocchi: [minor]
I think that in this case, the default could be a 0-column array rather than a column of ones (the columns from T will still be there):

```suggestion
            X = np.empty((Y.shape[0], 0))
```

econml/metalearners.py
```diff
@@ -150,39 +149,42 @@ def fit(self, Y, T, X, inference=None):
...
         -------
-        τ_hat : matrix, shape (m, d_y, n_t-1), n_t is number of treatments
+        τ_hat : matrix, shape (m, d_y, d_t)
             Constant marginal CATE of each treatment on each outcome for each sample X[i].
             Note that when Y is a vector rather than a 2-dimensional array,
             the corresponding singleton dimensions in the output will be collapsed
         """
         # Check inputs
+        if X is None:
+            X = np.ones((1, 1))
```
kbattocchi: As above, leave empty

```suggestion
            X = np.empty((1, 0))
```

econml/metalearners.py
```diff
@@ -70,11 +70,10 @@ def fit(self, Y, T, X, inference=None):
         Y, T, X, _ = check_inputs(Y, T, X, multi_output_T=False)
         T = self._label_encoder.fit_transform(T)
         self._one_hot_encoder.fit(T.reshape(-1, 1))
-        self._d_t = len(self._label_encoder.classes_)
-        self._d_y = Y.shape[1:]
-        self.models = check_models(self.models, self._d_t)
+        self._d_t = len(self._label_encoder.classes_) - 1
```
kbattocchi: To be consistent with other classes, this should probably always be a 1-tuple rather than a raw value. That is,
```suggestion
        self._d_t = len(self._label_encoder.classes_) - 1,
```

econml/metalearners.py
```diff
@@ -251,31 +253,30 @@ def fit(self, Y, T, X, inference=None):
...
```
kbattocchi: To be consistent with other classes, this should probably always be a 1-tuple rather than a raw value. That is,
```suggestion
        self._d_t = len(self._label_encoder.classes_) - 1,
```

econml/metalearners.py
kbattocchi: On the other hand, maybe it's silly to even allow X=None because there is no W (unlike DML)

econml/metalearners.py
```diff
@@ -372,36 +373,36 @@ def fit(self, Y, T, X, inference=None):
         Y, T, X, _ = check_inputs(Y, T, X, multi_output_T=False)
         T = self._label_encoder.fit_transform(T)
         self._one_hot_encoder.fit(T.reshape(-1, 1))
-        self._d_t = len(self._label_encoder.classes_)
-        self._d_y = Y.shape[1:]
-        self.models = check_models(self.models, self._d_t)
-        self.final_models = check_models(self.final_models, self._d_t - 1)
+        self._d_t = len(self._label_encoder.classes_) - 1
```
kbattocchi: To be consistent with other classes, this should probably always be a 1-tuple rather than a raw value. That is,
```suggestion
        self._d_t = len(self._label_encoder.classes_) - 1,
```

econml/metalearners.py
heimengqi: Yes. In the setting of Slearner, X = None is the same with learning the diff of mean(Y) in each class. I will keep it for now.

Issue #157, opened 2019-11-12
KernelDMLCateEstimator Model final


vasilismsr:
The `KernelDMLCateEstimator` should not be a child of `LinearDMLCateEstimator` because this means that the model_final is an unregularized linear model. We should instead have the model final be a hard-coded elasticnet regularized model and the user can specify the alpha and the l1_ratio.

kbattocchi:
Fixed by #159

PR #159, opened 2019-11-14
Vasilis/dml docs

vasilismsr:
Many changes to the dml user guide.

Also fixed bug #157 related to the model_final of KernelDMLCateEstimator.

Also added some docstrings to the statsmodelsinference classes and fixed docstring of the _RLearner class.


azure-pipelines.yml
```diff
@@ -4,148 +4,170 @@
 # https://docs.microsoft.com/azure/devops/pipelines/languages/python
```
kbattocchi: Why are these changes being made?

vasilismsr: No idea! This was the result of merging with your commits. It had some conflicts there and I accepted the incoming change.

vasilismsr: I figured these were the changes related to running linting only once.

kbattocchi: I've pushed to this branch and reverted this back.

kbattocchi:
Looks good.  I'll merge it in.

PR #163, opened 2019-11-15
Switch to macOS 10.13 to mitigate CI delays

kbattocchi:
We've been experiencing long delays with our CI builds, somehow involving the jobs that run on macOS; this PR is a temporary measure to see if we can mitigate the issue by using a different OS version.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,6 +12,8 @@ variables:
 jobs:
 - job: 'EvalChanges'
   displayName: 'Analyze changed files to determine which job to run'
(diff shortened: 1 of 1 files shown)


moprescu:
Looks good to me!

kbattocchi:
@moprescu Didn't work, though :(

kbattocchi:
Closed PR as the root cause on the Azure DevOps side seems to have been resolved.

kbattocchi:
Reopening to deal with the builds once again being slow.

PR #167, opened 2019-11-17
Added SparseLinearDRLearner

vasilismsr:
A DRLearner with a debiased lasso final model that enables inference when the heterogeneity features are high dimensional.

### DIFF
--- a/econml/drlearner.py
+++ b/econml/drlearner.py
@@ -684,13 +684,16 @@ class SparseLinearDRLearner(DebiasedLassoCateEstimatorDiscreteMixin, DRLearner):
     where :math:`\\phi(X)` is the outcome features of the featurizers, or `X` if featurizer is None. :math:`\\beta_t`
     is a an intercept of the CATE, which is included if ``fit_cate_intercept=True`` (Default). It fits this by
     running a debiased lasso regression (i.e. :math:`\\ell_1`-penalized regression with debiasing),
-    regressing the doubly robust outcome differences on X:
+    regressing the doubly robust outcome differences on X: i.e. first solves the penalized square loss problem
 
     .. math ::
         \\min_{\\theta_t, \\beta_t}\
         E_n\\left[\\left(Y_{i, t}^{DR} - Y_{i, 0}^{DR}\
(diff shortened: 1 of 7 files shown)


econml/drlearner.py
```diff
@@ -656,9 +660,228 @@ def multitask_model_cate(self):
...
+
+        If `sample_weight` is not None at fit time, then a weighted average across samples is returned.
+
+    """
+
+    def __init__(self,
+                 model_propensity=LogisticRegressionCV(cv=3, solver='lbfgs', multi_class='auto'),
+                 model_regression=WeightedLassoCV(cv=3),
+                 featurizer=None,
+                 fit_cate_intercept=True,
```
moprescu: Is this the case? I remember we had a conversation about this and you said that the cate intercept is not the same as the final model intercept. For the SparseDML estimator, fit_intercept is always false in the final model.

econml/drlearner.py
```diff
@@ -656,9 +660,228 @@ def multitask_model_cate(self):
...
+        """
+        # Replacing fit from DRLearner, to add debiasedlasso inference in docstring
+        # TODO: support sample_var
+        if sample_weight is not None and inference is not None:
+            warn("This estimator does not yet support sample variances and inference does not take "
+                 "sample variances into account. This feature will be supported in a future release.")
+        self._check_sparsity(X, T)
+        return super().fit(Y, T, X=X, W=W, sample_weight=sample_weight, sample_var=None, inference=inference)
+
+    def _check_sparsity(self, X, T):
```
moprescu: We might want to move this to utilities since it's used in two estimators.

econml/drlearner.py
vasilismsr: Yeap, here it should be true. This is the cate_intercept (i.e. if cate is <theta, phi> + beta or just <theta, phi>), not the final model intercept in the DML (which is a weird thing that has nothing to do with the CATE intercept). It's true that there is a small discrepancy in our interface between DML and DR. Keith is fixing that discrepancy and we will have fit_cate_intercept in DML too, which will be set to True by default.

PR #169, opened 2019-11-18
Fixed minor debiased lasso and sparse DML bugs

moprescu:
Removed `positive` from the list of valid attributes for the debiased lasso and multioutput debiased lasso. Made common base class LinearModelFinalCateEstimatorMixin.

### DIFF
--- a/econml/cate_estimator.py
+++ b/econml/cate_estimator.py
@@ -384,11 +384,6 @@ class TreatmentExpansionMixin(BaseCateEstimator):
 class LinearModelFinalCateEstimatorMixin(BaseCateEstimator):
     """Base class for models where the final stage is a linear model."""
 
-    def _get_inference_options(self):
-        # add statsmodels to parent's options
-        options = super()._get_inference_options()
-        return options
-
     @property
     def coef_(self):
         return self.model_final.coef_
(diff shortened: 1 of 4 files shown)


econml/cate_estimator.py
```diff
@@ -381,12 +381,12 @@ def effect(self, X=None, *, T0=0, T1=1):
     effect.__doc__ = BaseCateEstimator.effect.__doc__
 
 
-class StatsModelsCateEstimatorMixin:
+class LinearModelFinalCateEstimatorMixin(BaseCateEstimator):
+    """Base class for models where the final stage is a linear model."""
 
     def _get_inference_options(self):
```
vasilismsr: I think you can just remove this implementation here.

moprescu: no... we talked about this. Both `StatsModelsCateEstimatorMixin` and `DebiasedLassoCateEstimatorMixin` now inherit from this common base class.

vasilismsr: I mean the implementation of _get_inference_options. It is already implemented in the BaseCateEstimator and you just repeat it here.

moprescu: oh, gotcha. It shall be removed.

Issue #140, opened 2019-11-07
Use doctest to catch regressions via docstring examples


kbattocchi:
The [doctest](https://docs.python.org/3/library/doctest.html) module enables checking that examples in docstrings continue to produce the correct output.  There is also a [corresponding sphinx extension](http://www.sphinx-doc.org/en/master/usage/extensions/doctest.html) for checking examples in other documentation.  We should consider running these tools as part of CI.

kbattocchi:
Closed by #171

PR #171, opened 2019-11-19
Enable doctests as apart of CI loop

kbattocchi:
Enable doctests as part of CI loop


kbattocchi:
Fixes #140.

doc/spec/estimation/forest.rst
```diff
@@ -124,32 +131,35 @@ Similarly, we can call :py:class:`~econml.ortho_forest.DiscreteTreatmentOrthoFor
...
-    >>> est = ContinuousTreatmentOrthoForest()
-    >>> est.fit(Y, T, X, W)
-    >>> X_test = np.linspace(-1, 1, 30).reshape(-1, 1)
-    >>> treatment_effects = est.effect(X_test)
-    >>> plt.plot(X_test, y, label='ORF estimate')
-    >>> plt.plot(X_test[:, 0], np.exp(2*X_test[:, 0]), 'b--', label='True effect')
-    >>> plt.legend()
-    >>> plt.show()
+    .. doctest::
+        :options: +SKIP
```
vasilismsr: why skipping this?

doc/spec/estimation/forest.rst
```diff
@@ -58,10 +58,10 @@ supports fit and predict. The fit function needs to also support sample weights,
 If it does not, then we provided a weighted model wrapper :py:class:`~econml.ortho_forest.WeightedModelWrapper` that
 can wrap any class that supports fit and predict and enables sample weight functionality. This is done either
 by re-sampling the data based on the weights and then calling fit and predict, or, in the case of square losses of
-linear function classes, by re-scaling the features and labels appropriately based on the weights.
+linear function classes, by re-scaling the features and labels appropriately based on the weights::
 
-    >>> est = ContinuousTreatmentOrthoForest(model_y=WeightedModelWrapper(Lasso(), sample_type=sample_type),
-    ...                                      model_t=WeightedModelWrapper(Lasso(), sample_type=sample_type))
+    est = ContinuousTreatmentOrthoForest(model_y=WeightedModelWrapper(Lasso(), sample_type=sample_type),
```
vasilismsr: If we added a testcode below, we should also add one here. Also we should replace the models here with `econml.sklearn_extensions.linear_model.WeightedLasso`

doc/spec/estimation/forest.rst
```diff
@@ -124,32 +131,35 @@ Similarly, we can call :py:class:`~econml.ortho_forest.DiscreteTreatmentOrthoFor
...
+        >>> X = np.random.uniform(-1, 1, size=(4000, 1))
+        >>> W = np.random.normal(size=(4000, 50))
+        >>> support = np.random.choice(50, 4, replace=False)
+        >>> T = np.dot(W[:, support], np.random.normal(size=4)) + np.random.normal(size=4000)
+        >>> Y = np.exp(2*X[:, 0]) * T + np.dot(W[:, support], np.random.normal(size=4)) + .5
+        >>> est = ContinuousTreatmentOrthoForest()
+        >>> est.fit(Y, T, X, W)
+        >>> X_test = np.linspace(-1, 1, 30).reshape(-1, 1)
+        >>> treatment_effects = est.effect(X_test)
+        >>> plt.plot(X_test, y, label='ORF estimate')
```
vasilismsr: What happens to plots in doctests? Can we maybe move the plot out of the doctest and just say in words. If we plot X_test vs the predicted effect, here is the figure we get.

doc/spec/estimation/forest.rst
vasilismsr: from econml.ortho_forest import ContinuousTreatmentOrthoForest
from econml.sklearn_extensions.linear_model import WeightedLasso
np.random.seed(123)
X = np.random.uniform(-1, 1, size=(4000, 1))
W = np.random.normal(size=(4000, 50))
support = np.random.choice(50, 4, replace=False)
T = np.dot(W[:, support], np.random.normal(size=4)) + np.random.normal(size=4000)
Y = np.exp(2*X[:, 0]) * T + np.dot(W[:, support], np.random.normal(size=4)) + .5
est = ContinuousTreatmentOrthoForest(n_trees=100,
                                     max_depth=5,
                                     model_Y=WeightedLasso(alpha=0.01),
                                     model_T=WeightedLasso(alpha=0.01))
est.fit(Y, T, X, W)
X_test = np.linspace(-1, 1, 30).reshape(-1, 1)
treatment_effects = est.effect(X_test)
plt.plot(X_test[:, 0], treatment_effects, label='ORF estimate')
plt.plot(X_test[:, 0], np.exp(2*X_test[:, 0]), 'b--', label='True effect')
plt.legend()
plt.show()

vasilismsr: Then you can also add the resulting image:
ortho_estimate

doc/spec/estimation/forest.rst
kbattocchi: Doctests test that each command runs without error and that any printed output matches.  Since `plt.plot` doesn't print any output we will just verify that it runs without error.

However, that doesn't make it bad to include if it's something that we want users to try out, although we can't automatically test that the output visually matches.

vasilismsr: makes sense. We should still run it (After the mods proposed in the other comment)

doc/spec/estimation/forest.rst
kbattocchi: Okay.  I think the reason I avoided it was that it seemed like perhaps overkill to include all of the necessary imports in this introductory section, although we can include them in a hidden testcode block if we want to test this without adding too much content.

vasilismsr: Yeap, I would include in a hidden code block, just to make sure we have not mistake in an opening paragraph

kbattocchi: Done.  Hilariously, there was a mistake here which was caught by running it: for OrthoForests the parameters are called `model_T` and `model_Y` rather than `model_t` and `model_y`.

doc/spec/estimation/forest.rst
kbattocchi: Done.

doc/spec/estimation/forest.rst
vasilismsr: It's obvious we should be testing every single line :) Next PR we'll add tests for the tests :)

PR #172, opened 2019-11-19
automate the first stage model T and update DML notebook

heimengqi:


econml/dml.py
```diff
@@ -446,7 +453,7 @@ class SparseLinearDMLCateEstimator(DebiasedLassoCateEstimatorMixin, DMLCateEstim
     """
 
     def __init__(self,
-                 model_y=LassoCV(), model_t=LassoCV(),
+                 model_y=LassoCV(), model_t='auto',
```
vasilismsr: Please update all the docstrings accordingly with what auto means!!!

econml/dml.py
```diff
@@ -329,7 +336,7 @@ class LinearDMLCateEstimator(StatsModelsCateEstimatorMixin, DMLCateEstimator):
...
```
vasilismsr: Also model_y should be WeightedLassoCVWrapper.

econml/dml.py
vasilismsr: Also make WeightedLassoCVWrapper the default for model_y

econml/dml.py
```diff
@@ -551,7 +558,7 @@ class KernelDMLCateEstimator(DMLCateEstimator):
         by :mod:`np.random<numpy.random>`.
     """
 
-    def __init__(self, model_y=LassoCV(), model_t=LassoCV(),
+    def __init__(self, model_y=LassoCV(), model_t='auto',
```
vasilismsr: Similarly model_y and docstrings

econml/dml.py
```diff
@@ -170,6 +171,12 @@ def __init__(self,
         # TODO: consider whether we need more care around stateful featurizers,
         #       since we clone it and fit separate copies
 
+        if model_t == 'auto':
```
vasilismsr: Update docstring

econml/dml.py
vasilismsr: Also update docstring

econml/utilities.py
```diff
@@ -1871,3 +1872,42 @@ def fit(self, X, Y):
     def predict(self, X):
         predictions = self.model.predict(X)
         return reshape(predictions, (-1, 1)) if self.needs_unravel else predictions
+
+
+class WeightedLassoCVWrapper:
```
moprescu: This class belongs in `sklearn_extensions.linear_model`

econml/utilities.py
```diff
@@ -1871,3 +1872,42 @@ def fit(self, X, Y):
...
+    """Helper class to wrap either WeightedLassoCV or WeightedMultiTaskLassoCV depending on the shape of the target."""
+
+    def __init__(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
+    def fit(self, X, y):
+        assert shape(X)[0] == shape(y)[0]
```
moprescu: these asserts are not necessary since the underlying `self.model` model does these checks internally. Also, I recommend raising exceptions with messages rather than simple asserts.

econml/utilities.py
```diff
@@ -1871,3 +1872,42 @@ def fit(self, X, Y):
...
+        if ndim(y) == 2 and shape(y)[1] > 1:
+            self.model = WeightedMultiTaskLassoCV(*self.args, **self.kwargs)
+        else:
+            if ndim(y) == 2 and shape(y)[1] == 1:
+                y = np.ravel(y)
+                self.needs_unravel = True
+            self.model = WeightedLassoCV(*self.args, **self.kwargs)
+        self.model.fit(X, y)
+        # set intercept_ attribute
+        self.intercept_ = self.model.intercept_
```
moprescu: These look great! Thanks!

econml/utilities.py
```diff
@@ -1871,3 +1872,42 @@ def fit(self, X, Y):
...
+
+
+class WeightedLassoCVWrapper:
+    """Helper class to wrap either WeightedLassoCV or WeightedMultiTaskLassoCV depending on the shape of the target."""
+
+    def __init__(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
+    def fit(self, X, y):
```
moprescu: SUPER DUPPER IMPORTANT: add weights!!!

econml/utilities.py
```diff
@@ -1871,3 +1872,42 @@ def fit(self, X, Y):
...
+        assert ndim(y) <= 2
+        self.needs_unravel = False
+        if ndim(y) == 2 and shape(y)[1] > 1:
+            self.model = WeightedMultiTaskLassoCV(*self.args, **self.kwargs)
+        else:
+            if ndim(y) == 2 and shape(y)[1] == 1:
+                y = np.ravel(y)
+                self.needs_unravel = True
+            self.model = WeightedLassoCV(*self.args, **self.kwargs)
+        self.model.fit(X, y)
```
moprescu: add `sample_weights`!

moprescu:
Also, add some tests to `test_dml` where the `auto` option is being passed in and dml uses weights to make sure the E2E integration works. These tests would've caught the lack of `sample_weight` argument.

econml/utilities.py
heimengqi: Thanks for pointing it out. I check the test_dml to test 'auto' argument but didn't catch the sample_weight, will add test under test_can_use_sample_weights.

econml/tests/test_dml.py
```diff
@@ -149,7 +149,7 @@ def test_can_use_vectors(self):
     def test_can_use_sample_weights(self):
         """Test that we can pass sample weights to an estimator."""
         dmls = [
-            LinearDMLCateEstimator(LinearRegression(), LinearRegression(), featurizer=FunctionTransformer()),
+            LinearDMLCateEstimator(LinearRegression(), 'auto', featurizer=FunctionTransformer()),
```
vasilismsr: Can we do 'auto' for both here, so that we test that both our auto options can accept sample_weights?

heimengqi: Done!

Issue #119, opened 2019-10-11
Rework featurization in DML, add explicit intercept


kbattocchi:
There are a few issues with the way featurization works in the current version of our DML classes.  For example, given the default featurizer (which just adds a constant feature to the existing features to provide the equivalent of an intercept), if the user passes in None for X we end up with two constant columns (one because we treat no X as the same as a single constant feature and the other from the featurizer).

Instead, we should change the init parameters and their defaults:

1. We should add an explicit `fit_intercept=True` param independent of the featurizer; this will make it easier to present the results of fitting the model, and will additionally make it possible to do things like only shrink the the non-intercept weights when using Lasso (with some additional, separate, work)
2. Then we can make the default featurizer `None`, which will be easier to understand.  The user will be responsible for making sure that if they set the featurizer to something that includes a constant, that they switch `fit_intercept` to False.
3. When `X` is None and the featurizer is not None, we should warn the user and ignore the featurizer.

heimengqi:
Also, when we compute the cross product (ϕ(x);w)⊗(x;w)) we end up with many identical columns (because w_1 * w_2 == w_2 * w_1, for example.  Can we avoid this?  Would it affect, say, confidence intervals?

heimengqi:
Also, if W was generated by a polynomial featurizer, we'll have even more duplicates, because w_i^n*w_i^m is the same for many other combinations of n',m'.

vasilismsr:
#174 addresses this

PR #174, opened 2019-11-19
Add fit_cate_intercept to DML, rework feature generation

kbattocchi:


kbattocchi:
Fixes #119.

econml/dml.py
```diff
@@ -177,22 +182,23 @@ def __init__(self, model, is_Y):
                 self._is_Y = is_Y
 
             def _combine(self, X, W, n_samples, fitting=True):
+                no_x = X is None
+                if no_x:
```
vasilismsr: When X is None, then XW should just be W, not [1, W]. The extra column of ones, should not be appended to XW.

kbattocchi: I simplified the logic, which should now handle this correctly.

econml/dml.py
```diff
@@ -236,6 +239,8 @@ def _combine(self, X, T, fitting=True):
                     else:
                         F = X
                 else:
+                    if not fit_cate_intercept:
```
kbattocchi: I think that if you need `fit_cate_intercept` here, you should store it as an attribute on this model during init.  As I recall, things that are captured from the outside context like this don't get handled correctly when the model is cloned and run on a different process when used with bootstrap.

econml/dml.py
```diff
@@ -271,29 +276,91 @@ def predict(self, X):
...
     @property
     def featurizer(self):
+        # NOTE This is used by the inference methods and has to be the overall featurizer. intended
+        # for internal use by the library
         return super().model_final._featurizer
 
     @property
     def model_final(self):
+        # NOTE This is used by the inference methods and is more for internal use to the library
+        return super().model_final._model
```
kbattocchi: Should these be documented?  Is `original_featurizer` ever used outside of this class?

econml/dml.py
```diff
@@ -271,29 +276,91 @@ def predict(self, X):
...
+
+    def model_cate(self):
```
kbattocchi: 1. Should this be decorated with `@property`?  
2. What's the distinction between `model_final` and `model_cate`?

econml/dml.py
```diff
@@ -271,29 +276,91 @@ def predict(self, X):
...
+        input_feature_names: list of strings of length X.shape[1] or None
+            The names of the input features
+
+        Returns
+        -------
+        out_feature_names: list of strings or None
+            The names of the output features :math:`\\phi(X)`, i.e. the features with respect to which the
+            final constant marginal CATE model is linear. It is the names of the features that are associated
+            with each entry of the :meth:`coef_` parameter. Available only when the featurizer is not None and has
+            a method: `get_feature_names(input_feature_names)`. Otherwise None is returned.
```
kbattocchi: This does not appear to correctly document the behavior - if the original featurizer is None we just return the input features unchanged (which is right, because those are the features we use in the final model)

econml/dml.py
```diff
@@ -595,5 +665,5 @@ def transform(self, X):
         super().__init__(model_y=model_y, model_t=model_t,
                          model_final=ElasticNetCV(),
                          featurizer=RandomFeatures(random_state),
-                         fit_cate_intercept=False,
+                         fit_cate_intercept=fit_cate_intercept,
```
kbattocchi: [minor, mostly out of curiosity]
Shouldn't the random features already approximate a constant well so that an intercept is unnecessary?

econml/utilities.py
```diff
@@ -38,6 +38,24 @@ def transform(self, X):
...
+    if bias_part_of_coef:
+        cate_coef = coef.reshape(d_y + dt + (-1,))[..., 1:]
+        if fit_cate_intercept:
+            cate_intercept = coef.reshape(d_y + dt + (-1,))[..., 0]
+    else:
+        cate_coef = coef.reshape(d_y + dt + (-1,))
+        if fit_cate_intercept:
+            cate_intercept = np.reshape(intercept, d_y + dt)
+    if (cate_intercept is not None) and (np.ndim(cate_intercept) == 0):
+        cate_intercept = np.asscalar(cate_intercept)
```
kbattocchi: I believe `np.asscalar` is deprecated in favor of:
```suggestion
        cate_intercept = cate_intercept.item()
```

econml/utilities.py
```diff
@@ -38,6 +38,24 @@ def transform(self, X):
...
+    cate_intercept = None
+    if bias_part_of_coef:
+        cate_coef = coef.reshape(d_y + dt + (-1,))[..., 1:]
+        if fit_cate_intercept:
+            cate_intercept = coef.reshape(d_y + dt + (-1,))[..., 0]
+    else:
+        cate_coef = coef.reshape(d_y + dt + (-1,))
+        if fit_cate_intercept:
+            cate_intercept = np.reshape(intercept, d_y + dt)
+    if (cate_intercept is not None) and (np.ndim(cate_intercept) == 0):
```
kbattocchi: I don't understand this part - if `np.ndim` is 0, isn't the intercept already a scalar?

econml/utilities.py
```diff
@@ -235,14 +253,13 @@ def cross_product(*XS):
     X2 : n x d2 matrix
         Second matrix of n samples of d2 features
         (or an n-element vector, which will be treated as an n x 1 matrix)
-    …
 
     Returns
     -------
-    A : n x (d1*d2*...) matrix
+    n x (d1*d2*...) matrix
```
kbattocchi: Add the name for the output back so that the docstring is rendered better
```suggestion
    A : n x (d1*d2*...) matrix
```

econml/utilities.py
```diff
@@ -251,9 +268,12 @@ def cross_product(*XS):
     for X in XS:
         assert n == shape(X)[0]
 
+    # TODO: wouldn't making X1 vary more slowly than X2 be more intuitive?
```
kbattocchi: If we're never going to change it, maybe remove the TODO.

econml/dml.py
vasilismsr: Nope these are for internal use. Don't want that these show up in the docs.

econml/dml.py
vasilismsr: consistency with the DR Learner interface

vasilismsr: Done with adding property

econml/dml.py
vasilismsr: Done

econml/dml.py
vasilismsr: Yeah but having an intercept always good for regularization. Also it breaks the X=None case and this was an easy fix.

econml/utilities.py
vasilismsr: No it ends up being e.g. np.array(.5). Behaves like a scalar but has this weird array around it.

econml/utilities.py
vasilismsr: Done

econml/utilities.py
vasilismsr: done

econml/cate_estimator.py
```diff
@@ -405,19 +405,17 @@ def coef_(self):
...
+        coef: (n_x,) or (n_t, n_x) or (n_y, n_t, n_x) array like
             Where n_x is the number of features that enter the final model (either the
             dimension of X or the dimension of featurizer.fit_transform(X) if the CATE
             estimator has a featurizer.), n_t is the number of treatments, n_y is
-            the number of outcomes. The coefficient is flattened in a manner that
-            the first block of n_x columns are the coefficients associated with treatment 0,
-            the next n_x columns are the coefficients associated with treatment 1 etc.
+            the number of outcomes. Dimensions are omitted if the original input was
+            a vector and not a 2D array. For binary treatment the n_t dimension is
+            also omitted.
```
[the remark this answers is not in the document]
kbattocchi: Discussed offline; this seems to make the ergonomics better and maybe we should change `marginal_effect`, etc. in the next release to behave the same way.

Issue #158, opened 2019-11-13
sample_weight in OrthoForest should be passed as kwdarg to sklearn model


vasilismsr:
In this line:
https://github.com/microsoft/EconML/blob/72a7022d99b92149dca8a09735923d1952a102c2/econml/ortho_forest.py#L59

We should be passing `sample_weight` as keyword. Otherwise a weird error is being displayed. If for instance the model does accept extra arguments, but those are not `sample_weight`. e.g. in the following:

```python
from econml.ortho_forest import ContinuousTreatmentOrthoForest
from sklearn.linear_model import Lasso
import numpy as np

n = 5000
d = 5
X = np.random.uniform(-1, 1, size=(n, d))
true_fn = lambda x: x[:, 0]**2
T = X[:, 0] + np.random.normal(size=(n,))
y = true_fn(X) * T + X[:, 0] + np.random.normal(size=(n,))

x_grid = np.linspace(-1, 1, 100)
X_test = np.hstack([x_grid.reshape(-1, 1), np.random.normal(size=(100, d-1))])
est = ContinuousTreatmentOrthoForest(model_Y=Lasso(alpha=0.01),
                                     model_T=Lasso(alpha=0.01),
                                     n_trees=1, subsample_ratio=.5, min_leaf_size=10)
est.fit(y, T, X)
est.effect(X_test)
```

Yields an error:

```
ValueError: The original error: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all() This might be caused by too few sample in the tree leafs. Try increasing the min_leaf_size.
```

Instead of sth along the lines: `sample_weight` is not a valid argument.

PR #175, opened 2019-11-19
Fixed ORF bug and added a test

moprescu:


moprescu:
Fixes #158

econml/ortho_forest.py
```diff
@@ -56,10 +56,17 @@ def _build_tree_in_parallel(Y, T, X, W,
 
 def _fit_weighted_pipeline(model_instance, X, y, sample_weight):
     if not isinstance(model_instance, Pipeline):
-        model_instance.fit(X, y, sample_weight)
+        try:
+            model_instance.fit(X, y, sample_weight=sample_weight)
+        except Exception:
```
kbattocchi: 1.  Can we catch a more specific exception here?  I can imagine that other issues might also hit this code path, making the error message that you generate misleading.
2.  Do we even need to catch and rethrow rather than just relying on the existing error now that you're using a keyword argument?
3.  If you are going to do this, consider preserving the original exception context by using the `raise ... from ...` syntax

econml/ortho_forest.py
```diff
@@ -56,10 +56,17 @@ def _build_tree_in_parallel(Y, T, X, W,
...
+            raise TypeError("Estimators of type {} do not accept weights.".format(model_instance.__class__.__name__))
     else:
-        last_step_name = model_instance.steps[-1][0]
-        model_instance.fit(X, y, **{"{0}__sample_weight".format(last_step_name): sample_weight})
+        try:
+            last_step_name = model_instance.steps[-1][0]
+            model_instance.fit(X, y, **{"{0}__sample_weight".format(last_step_name): sample_weight})
+        except Exception:
```
kbattocchi: Same comment here.

econml/tests/test_orf.py
```diff
@@ -178,6 +178,22 @@ def test_effect_shape(self):
         assert est.effect(X[:3], T0=0, T1=2).shape == (3,), "Effect dimension incorrect"
         assert est.effect(X[:3], T0=1, T1=2).shape == (3,), "Effect dimension incorrect"
 
+    def test_nuissance_model_has_weights(self):
```
kbattocchi: [minor] spelling
```suggestion
    def test_nuisance_model_has_weights(self):
```

econml/ortho_forest.py
kbattocchi: Also, if we're going to emit our own error, it might be nice to suggest using a WeightedModelWrapper as a workaround.

moprescu: Great suggestion! Fixed.

econml/ortho_forest.py
```diff
@@ -55,11 +55,29 @@ def _build_tree_in_parallel(Y, T, X, W,
...
+        "Estimators of type {} do not accept weights. "
+        "Consider using the class WeightedModelWrapper from econml.utilities to build a weighted model."
+    )
+    expected_error_msg = "fit() got an unexpected keyword argument 'sample_weight'"
     if not isinstance(model_instance, Pipeline):
-        model_instance.fit(X, y, sample_weight)
+        try:
+            model_instance.fit(X, y, sample_weight=sample_weight)
+        except TypeError as e:
+            if expected_error_msg in str(e):
```
vasilismsr: Would that not be very sensitive to the exact error msg that sklearn raises here? Is there a more robust way to catch this?

PR #177, opened 2019-11-20
Vasilis/cate interpreters

vasilismsr:
Intepretation of CATE via Trees


kbattocchi:
@vasilismsr I've heavily refactored the tree exporting logic.  Can you take a look and verify you're happy with the general shape of things (maybe just look at the changes from my commits)?

econml/cate_interpreter.py
```diff
@@ -0,0 +1,586 @@
...
+                                                min_samples_leaf=self.min_samples_leaf,
+                                                min_weight_fraction_leaf=self.min_weight_fraction_leaf,
+                                                max_features=self.max_features,
+                                                random_state=self.random_state,
+                                                max_leaf_nodes=self.max_leaf_nodes,
+                                                min_impurity_decrease=self.min_impurity_decrease)
+        y_pred = cate_estimator.const_marginal_effect(X)
+
+        assert (y_pred.ndim == 1) or (y_pred.shape[1] ==
+                                      1), "Only single-dimensional treatment interpretation available!"
```
heimengqi: If it's multi dimensional Y, y_pred.shape[1] is the dimension of Y not T. Or we also don't support multi dimensional Y? What should it be if multi Y and also include uncertainty?

econml/cate_interpreter.py
```diff
@@ -0,0 +1,586 @@
...
+
+        if y_pred.ndim == 1:
+            y_pred = y_pred.reshape(-1, 1)
+
+        if self.include_uncertainty:
+            y_lower, y_upper = cate_estimator.const_marginal_effect_interval(X, alpha=self.uncertainty_level)
```
heimengqi: I think default alpha in other places is 0.1, should we change it to 0.1 as well here?

econml/cate_interpreter.py
```diff
@@ -0,0 +1,586 @@
...
+                                                 max_leaf_nodes=self.max_leaf_nodes,
+                                                 min_impurity_decrease=self.min_impurity_decrease)
+        if self.risk_level is None:
+            y_pred = cate_estimator.const_marginal_effect(X)
+        elif not self.risk_seeking:
+            y_pred, _ = cate_estimator.const_marginal_effect_interval(X, alpha=self.risk_level)
+        else:
+            _, y_pred = cate_estimator.const_marginal_effect_interval(X, alpha=self.risk_level)
+
+        assert (y_pred.ndim == 1) or (y_pred.shape[1] == 1), "Only binary treatment interpretation available!"
```
heimengqi: Same as above, how about multi dimensional Y?

econml/cate_interpreter.py
kbattocchi: I've improved the assertion logic and message to make it clear that we only support single-treatment, single-outcome estimates.

econml/cate_interpreter.py
kbattocchi: Done.

econml/cate_interpreter.py
kbattocchi: Also addressed this assertion.

PR #180, opened 2019-11-21
Update flowchart

kbattocchi:
Update the flowchart and add the classifier matrix


doc/spec/flowchart.rst
```diff
@@ -5,4 +5,163 @@ Library Flow Chart
...
+        The MetaLearner and DRLearner estimators offer the choice of any ML estimation model in all stages
+        and allows for model selection via cross validation. This enhances flexibility, but because the
+        sample data is used to choose among models it is impossible to calculate honest analytic confidence
+        intervals. Moreover, most ML estimation approaches introduce bias for regularization purposes, so as
+        to optimally balance bias and variance. Hence, confidence intervals based on such biased estimates
+        will be invalid. For these models it is still possible to construct bootstrap confidence intervals,
+        but this process is slow, may not be accurate in small samples and these intervals only capture the 
+        variance but not the bias of the model. 
+
++---------------------------------------------+----------------+---------------------+--------------------------+------------------+----------------------+-------------------+---------------------+-------------------------- +
```
vasilismsr: Weirdly this table doesn't show up as a table when I compile, but rather as raw markdown.

doc/spec/flowchart.rst
```diff
@@ -5,4 +5,163 @@ Library Flow Chart
...
+| Estimator                                   | Treatment Type | Requires Instrument | Delivers Conf. Intervals | Linear Treatment | Linear Heterogeneity | Mulitple Outcomes | Multiple Treatments | High-Dimensional Features |
```
vasilismsr: I think because of the long glossary right above, I would make this table a separate rst file with its own heading on the left in the navigation so that it is more visible.

doc/spec/flowchart.rst
kbattocchi: Fixed

doc/spec/flowchart.rst
kbattocchi: I've split the table out to its own "comparison" page

PR #186, opened 2019-11-22
Fix some documentation problems

kbattocchi:
* We forgot to include the `sklearn_extensions` modules.
* Used a more consistent cross-reference style everywhere
* Fixed a few other miscellaneous doc issues (spelling, invalid cross refs, etc.)

### DIFF
--- a/doc/spec/estimation/dml.rst
+++ b/doc/spec/estimation/dml.rst
@@ -29,9 +29,9 @@ provide *valid inference* (confidence interval construction) for measuring the u
 What are the relevant estimator classes?
 ========================================
 
-This section describes the methodology implemented in the classes, :py:class:`~econml._rlearner._RLearner`,
-:py:class:`~econml.dml.DMLCateEstimator`, :py:class:`~econml.dml.LinearDMLCateEstimator`,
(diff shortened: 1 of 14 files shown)


doc/spec/estimation/dml.rst
```diff
@@ -553,26 +586,25 @@ Multiple Outcome, Multiple Treatments
...
 
+    from sklearn.linear_model import MultiTaskElasticNet
     est = LinearDMLCateEstimator(model_y=MultiTaskElasticNet(alpha=0.1),
                                  model_t=MultiTaskElasticNet(alpha=0.1))
     est.fit(Y, T, None, W)
 
     # a_hat[i,j] contains the elasticity of the demand of product i on the price of product j
     a_hat = est.const_marginal_effect()
 
+.. todo:: The following example requires lightning before we can run it as testcode
```
kbattocchi: @vasilismsr 

I've enabled all code in this file to run as doctests, except for this one example, because it requires `lightning`.  Do you think it's worth adding lightning as a test-time dependency just for this one example?

vasilismsr: For some weird reason I really like this example :) I believe lightning is also pip installable

kbattocchi: Okay, added lightning when building docs and re-enabled this example (which had an issue: shape of X passed to const_marginal_effect was not correct)

PR #187, opened 2019-11-22
Switch selective regularization approach away from tensorflow

### LINKED ISSUE #179: Selective regularization is causing sporadic build errors
The short term fix is to temporarily remove selective regularization from utilities and tests.

The longer term fix is to switch to the prototype implementation, which doesn't require tensorflow or work via gradient descent.

kbattocchi:
There's a straightforward general way of selectively regularizing based on any linear model.  We should move to that instead of using the current tensorflow-based gradient descent algorithm.

Fixes #179.


kbattocchi:
> I think we should move this into the linear_model module of the sklearn extensions.

Fine by me.  Other than that does everything look reasonable to you?

econml/selective_regularization.py
```diff
@@ -3,193 +3,98 @@
...
+        coefs[self._penalized_inds] = self._penalized_model.coef_
+        coefs[self._unpenalized_inds] = self._model_X1.coef_
+        return coefs
+
+    @property
+    def intercept_(self):
+        if not self._fit_intercept:
+            raise AttributeError("intercept_ property is only supported when fit_intercept=True")
+        # Note that the penalized model should *not* have an intercept
+        return self._model_X1.intercept_
```
vasilismsr: Might also be good to expose the internal fitted penalized model with some public method.

econml/selective_regularization.py
```diff
@@ -3,193 +3,98 @@
...
     @property
     def coef_(self):
-        """Get the model coefficients."""
-        coef_reg = self.session.run(self.weights_reg.value())
-        coef_ureg = self.session.run(self.weights_ureg.value())
-        full_coef = np.zeros((coef_reg.shape[0] + coef_ureg.shape[0], coef_reg.shape[1]))
-        full_coef[self._subset, :] = coef_reg
-        full_coef[self._subset_c, :] = coef_ureg
-        return full_coef
+        coefs = np.empty(len(self._penalized_inds) + len(self._unpenalized_inds))
```
vasilismsr: This is wrong for multi-dimensional y. The following raises an error here.

```python
from econml.selective_regularization import SelectiveRegularization
from sklearn.linear_model import Lasso, ElasticNet, LassoCV, ElasticNetCV, MultiTaskElasticNetCV
import numpy as np
n = 1000
d = 100
X = np.random.normal(size=(n, d))
y = X[:, [0, 0]] + np.random.normal(size=(n, 2))
est = SelectiveRegularization([0, 1, 2], MultiTaskElasticNetCV(fit_intercept=False))
est.fit(X, y)
est.coef_
```

vasilismsr: e.g. the following would work:

```python
        if np.ndim(self._penalized_model.coef_) == 1:
            coefs = np.empty(len(self._penalized_inds) + len(self._unpenalized_inds))
            coefs[self._penalized_inds] = self._penalized_model.coef_
            coefs[self._unpenalized_inds] = self._model_X1.coef_
        else:
            coefs = np.empty((self._penalized_model.coef_.shape[0], len(
                self._penalized_inds) + len(self._unpenalized_inds)))
            coefs[:, self._penalized_inds] = self._penalized_model.coef_
            coefs[:, self._unpenalized_inds] = self._model_X1.coef_
```

econml/selective_regularization.py
```diff
@@ -3,193 +3,98 @@
...
-        self.optimizer = tf.train.AdagradOptimizer(learning_rate=self._learning_rate)
-        self.train = self.optimizer.minimize(self.cost)
-
-        self.session = tf.Session()
+    def __init__(self, unpenalized_inds, penalized_model, fit_intercept=False):
+        self._unpenalized_inds = unpenalized_inds
+        self._penalized_model = clone(penalized_model)
+        self._fit_intercept = fit_intercept
 
     def fit(self, X, y):
```
vasilismsr: Fit should support sample weights. Should be passed to all the models (both linear regressions and penalized model)

vasilismsr: Would require adding tests that it is doing the right thing and that running this with summarized data and weights is the same as on the unsummarized data without weights.

econml/selective_regularization.py
kbattocchi: Fixed

econml/selective_regularization.py
kbattocchi: Fixed (and test added)

vasilismsr:
> I think we should move this into the linear_model module of the sklearn extensions.

Should we not make it part of `linear_model`? It seems to be a method restricted to linear models.

econml/sklearn_extensions/selective_regularization.py
```diff
@@ -0,0 +1,107 @@
...
+                                 "column in the penalized indices.")
+
+        # now regress X1 on y - X2 * beta2 to learn beta1
+        self._model_X1 = LinearRegression(fit_intercept=self._fit_intercept)
+        self._model_X1.fit(X1, y - self._penalized_model.predict(X2), sample_weight=sample_weight)
+
+        return self
+
+    def predict(self, X):
+        X1 = X[:, self._unpenalized_inds]
```
vasilismsr: This requires running check_fitted. Maybe even check_inputs.

econml/sklearn_extensions/selective_regularization.py
```diff
@@ -0,0 +1,107 @@
...
+        X2 = X[:, self._penalized_inds]
+        return self._model_X1.predict(X1) + self._penalized_model.predict(X2)
+
+    @property
+    def coef_(self):
+        y_dim = self._model_X1.coef_.shape[0:-1]  # get () if y was a vector, or (d_y,) otherwise
```
vasilismsr: Similarly this requires check_fitted.

econml/sklearn_extensions/selective_regularization.py
```diff
@@ -0,0 +1,107 @@
...
+        x_dim = len(self._penalized_inds) + len(self._unpenalized_inds),
+        coefs = np.empty(y_dim + x_dim)
+        coefs[..., self._penalized_inds] = self._penalized_model.coef_
+        coefs[..., self._unpenalized_inds] = self._model_X1.coef_
+        return coefs
+
+    @property
+    def intercept_(self):
+        if not self._fit_intercept:
```
vasilismsr: Similarly requires check_fitted.

econml/sklearn_extensions/selective_regularization.py
```diff
@@ -0,0 +1,107 @@
...
+            raise AttributeError("intercept_ property is only supported when fit_intercept=True")
+        # Note that the penalized model should *not* have an intercept
+        return self._model_X1.intercept_
+
+    def score(self, X, y):
+        """Score the predictions for a set of features to ground truth."""
```
vasilismsr: Similarly requires check_fitted and maybe check_inputs.

econml/sklearn_extensions/selective_regularization.py
```diff
@@ -0,0 +1,107 @@
...
+    and :math:`X_2` when regressed on :math:`X_1` using OLS.
+    """
+
+    def __init__(self, unpenalized_inds, penalized_model, fit_intercept=False):
+        self._unpenalized_inds = unpenalized_inds
+        self._penalized_model = clone(penalized_model)
+        self._fit_intercept = fit_intercept
+
+    def fit(self, X, y, sample_weight=None):
+        self._penalized_inds = np.delete(np.arange(X.shape[1]), self._unpenalized_inds)
```
vasilismsr: We most probably want to add check_inputs here (using for instance the sklearn check_inputs).

econml/sklearn_extensions/selective_regularization.py
```diff
@@ -0,0 +1,107 @@
...
+
+    .. math::
+        \\lVert (M_{X_1} y) - (M_{X_1} X_2) \\beta_2 \\rVert^2 + f(\\beta_2)
+
+    so finding the minimizing :math:`\\beta_2` can be done by regressing :math:`M_{X_1} y` on :math:`M_{X_1} X_2` using
+    the penalized regression method incorporating :math:`f`.  Note that these are just the residual values of :math:`y`
+    and :math:`X_2` when regressed on :math:`X_1` using OLS.
+    """
+
+    def __init__(self, unpenalized_inds, penalized_model, fit_intercept=False):
```
vasilismsr: I think as is convention in scikit-learn, fit_intercept should be set to True.

kbattocchi: Fixed

econml/sklearn_extensions/selective_regularization.py
kbattocchi: Now uses `check_X_y`

econml/sklearn_extensions/selective_regularization.py
kbattocchi: Done.

econml/sklearn_extensions/selective_regularization.py
kbattocchi: Made an attribute set during fit rather than a property.

econml/sklearn_extensions/selective_regularization.py
kbattocchi: Made an attribute set during fit rather than a property

econml/sklearn_extensions/selective_regularization.py
kbattocchi: Done.

PR #200, opened 2019-12-06
Prepare v0.6 release

kbattocchi:


kbattocchi:
Won't merge until #177 is complete.

vasilismsr:
Can we maybe call it .51 

Doesn’t seem a big delta from .5

kbattocchi:
@vasilismsr Doesn't seem so small to me - aren't the forest learners, and non-parametric DML, and CATE interpreters important additions?  I believe that most packages use [semantic versioning](https://semver.org/), so a release like 0.5.1 would indicate only bugfixes, not new features, although I think most people are pretty forgiving of other schemes pre-1.0.

vasilismsr:
Agreed!

PR #208, opened 2019-12-23
Update dml.py

vasilismsr:
Small bug in clipping in DML with weight trick


vasilismsr:
> LGTM. Can you think of a simple test that would be fixed by this?

One easy would be to use a classifier instead of a regressor and forget the discrete_treatment=True flag. Then many T residuals will be zero

PR #209, opened 2020-01-06
Add 2d response array support to DRLearner

kbattocchi:
As requested by the DoWhy team, this adds support for 2D arrays for the response when using DRLearner (though still limited to a single response).

### DIFF
--- a/econml/drlearner.py
+++ b/econml/drlearner.py
@@ -29,15 +29,14 @@ Tsiatis AA (2006).
 
 import numpy as np
 from warnings import warn
-from sklearn.linear_model import LogisticRegressionCV, LinearRegression, LassoCV
-from econml.utilities import inverse_onehot, check_high_dimensional
-from econml.sklearn_extensions.linear_model import WeightedLassoCV, DebiasedLasso
-from econml.sklearn_extensions.ensemble import SubsampledHonestForest
+
 from sklearn.base import clone
+from sklearn.linear_model import LogisticRegressionCV, LinearRegression, LassoCV
+from econml.utilities import inverse_onehot, check_high_dimensional, StatsModelsLinearRegression
+from econml.sklearn_extensions.linear_model import WeightedLassoCVWrapper, DebiasedLasso
+from econml.sklearn_extensions.ensemble import SubsampledHonestForest
 from econml._ortho_learner import _OrthoLearner
 from econml.cate_estimator import StatsModelsCateEstimatorDiscreteMixin, DebiasedLassoCateEstimatorDiscreteMixin
-from econml.utilities import StatsModelsLinearRegression
-from sklearn.preprocessing import PolynomialFeatures
 from econml.inference import GenericModelFinalInferenceDiscrete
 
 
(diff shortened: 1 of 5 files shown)


econml/drlearner.py
```diff
@@ -316,19 +317,21 @@ def predict(self, X=None):
...
+                    preds = np.array([mdl.predict(X) for mdl in self.models_cate])
+                    return np.moveaxis(preds, 0, -1)  # move treatment dim to end
 
             def score(self, Y, T, X=None, W=None, *, nuisances, sample_weight=None, sample_var=None):
                 if (X is not None) and (self._featurizer is not None):
                     X = self._featurizer.transform(X)
                 Y_pred, = nuisances
                 if self._multitask_model_final:
-                    return np.mean(np.average((Y_pred[:, 1:] - Y_pred[:, [0]] - self.model_cate.predict(X))**2,
+                    return np.mean(np.average((Y_pred[..., 1:] - Y_pred[..., [0]] - self.model_cate.predict(X))**2,
```
vasilismsr: Are we sure that rhis will work when we have a multitask model cate? Can we try a unit test where model cate is multi task lasso and we also have a column Y. I’m worried it wont work.

vasilismsr: As I was imagining it fails. I think we should fix this. i.e. the following code fails and it should still be working. In some sense, when y is (-1, 1) then the method should work exactly as if it was (-1,) and just at the end return an extra explicit dimension of the y, in all the quantities we return (according to API). 

```python
import numpy as np
from sklearn.linear_model import LassoCV
from econml.drlearner import LinearDRLearner, DRLearner
from sklearn.linear_model import LogisticRegressionCV, LinearRegression
from sklearn.dummy import DummyClassifier
import scipy.special

np.random.seed(123)
n=2000 # number of raw samples
d=10 # number of binary features + 1

# Generating random segments aka binary features. We will use features 0,...,3 for heterogeneity.
# The rest for controls. Just as an example.
X = np.random.binomial(1, .5, size=(n, d))
# Generating an imbalanced A/B test
T = np.random.binomial(1, scipy.special.expit(X[:, 0]))
# Generating an outcome with treatment effect heterogeneity. The first binary feature creates heterogeneity
# We also have confounding on the first variable. We also have heteroskedastic errors.
y = (-1 + 2 * X[:, 0]) * T + X[:, 0] + (1*X[:, 0] + 1)*np.random.normal(0, 1, size=(n,))

# One can replace model_y and model_t with any scikit-learn regressor and classifier correspondingly
# as long as it accepts the sample_weight keyword argument at fit time.
est = DRLearner(model_regression=LassoCV(cv=3),
                      model_propensity=DummyClassifier(strategy='prior'),
                      model_final=LinearRegression(), multitask_model_final=True)
est.fit(y.reshape(-1, 1), T, X=X[:, :4])
```

kbattocchi: Good catch.  Working on it.

kbattocchi: Fixed.

PR #210, opened 2020-01-07
Enable Python 3.8 support

kbattocchi:


kbattocchi:
Abandoning this for now because tensorflow won't support Python 3.8 until version 2.2 sometime later this year, and we're limiting to tensorflow 1.* for now anyway.

kbattocchi:
Tensorflow 2.2 will soon be out, so I'm reopening this so that we can support Python 3.8, which has been out for several months and is seeing increased use in practice.

azure-pipelines.yml
```diff
@@ -103,40 +103,40 @@ jobs:
...
+#             PYTEST_ADDOPTS: '-m "automl" -n 0'
+#             COVERAGE_PROCESS_START: 'setup.cfg'
+
+#         - task: PublishTestResults@2
+#           displayName: 'Publish Test Results **/test-results.xml'
+#           inputs:
+#             testResultsFiles: '**/test-results.xml'
+#             testRunTitle: 'AutoML'
+#           condition: succeededOrFailed()
+#       package: '.[automl]'
```
kbattocchi, line 139: Had to disable AutoML tests, because it seems like they were already broken by something changing server-side for AutoML and I couldn't figure out the necessary changes after a small amount of debugging.  I've opened a new issue (#270) to fix this.

doc/spec/estimation/dml.rst
```diff
@@ -696,9 +696,7 @@ the case where this matrix has low rank: all the products can be embedded in som
 space and the cross-price elasticities is a linear function of these low dimensional embeddings. This corresponds
 to well-studied latent factor models in pricing. Our framework can easily handle this by using 
 a nuclear norm regularized multi-task regression in the final stage. For instance the 
-lightning package implements such a class:
-
-.. testcode::
+lightning package implements such a class::
```
kbattocchi, line 699: Released versions of lightning are currently not compatible with sklearn 0.23 (though this problem has been fixed in the lightning GitHub repo - see https://github.com/scikit-learn-contrib/lightning/pull/142).  We can revert this small change if there is a new PyPI release.

PR #211, opened 2020-01-07
Changed the alpha grid search in the DebiasedLasso.

moprescu:
* Changed the alpha grid for the alpha=`auto` setting to the default
LassoCV alpha path
* Cleaned up tests for sklearn extensions of linear models


moprescu:
> Would it be possible to add a test that fails with the old selection strategy but passes now? Otherwise LGTM.

@kbattocchi 
This is not actually a bug, it's just that the alpha grid I was choosing was too coarse and sometimes for scenarios where smaller alpha was more optimal, there could be some bias in the debiasedLasso compared with the LassoCV for example. 

I am now using the LassoCV path search for the alphas which takes a bit longer, but gives more optimal results in these corner cases. But it's hard to test because we don't know the true best alpha and small bias can be hard to measure.

kbattocchi:
@moprescu I thought Vasilis had a concrete example where the CIs were way off because of the chosen alpha, and hoped maybe that could be adapted into a useful test to make sure we don't regress.

kbattocchi:
Nevermind, discussed offline; I was misunderstanding.

PR #213, opened 2020-01-22
AutomatedML for EconML

v-keacqu:
Tests and Code for AutomatedML addition to EconML. Test will fail as credentials are not yet configured on Azures Pipeline service. In addition there are is not concrete testing for performance. But this will serve as a draft PR.


econml/automated_ml.py
```diff
@@ -0,0 +1,360 @@
...
+    Add AutomatedMLMixin to specified base class.
+
+    In addition to a mixin that creates automatedML models from AutoMLConfig
+    objects, another mixin is added specifying modifications to the
+    AutoMLConfig. The business logic that this mixin runs is detailed below:
+
+    If the baseClass is of type NonParamDMLCateEstimator
+    the automatedML configuration must support sample weights for
+    correctness, and addAutomatedML will add functionality inject the
+    following models to the ``model_final_automl_config`` blacklist.
```
kbattocchi: Wouldn't it be better to use a whitelist in case new regressors are added in the future?

econml/automated_ml.py
```diff
@@ -0,0 +1,360 @@
...
+
+def addAutomatedML(baseClass):
+    """
+    Add AutomatedMLMixin to specified base class.
+
+    In addition to a mixin that creates automatedML models from AutoMLConfig
+    objects, another mixin is added specifying modifications to the
+    AutoMLConfig. The business logic that this mixin runs is detailed below:
+
+    If the baseClass is of type NonParamDMLCateEstimator
```
kbattocchi: 1. I don't see any logic that actually does this - am I missing something?
2. Both here and for what you say about linearity below, only the final stage model has this requirement; the first stage models do not need to support weights (or be linear, in the other case)

econml/automated_ml.py
```diff
@@ -0,0 +1,360 @@
...
+        kwargs: Dict, optional
+           kwargs that are passed in order to initiate the final automatedML run.
+           Any kwarg, that is an AutoMLConfig, will be converted into as
+           AutomatedMLModel.
+       """
+        # Loop through the kwargs and args if any of them is an AutoMLConfig file, pass them
+        # create model and pass model into final.
+        new_args = ()
+        for var in args:
+            new_args += (var,) if var is not None else (None,)
```
kbattocchi: 1. It doesn't look like the check for `None` accomplishes anything.
2. Shouldn't positional arguments also be wrapped via `_get_automated_ml_model`?

econml/tests/test_automated_ml.py
```diff
@@ -0,0 +1,235 @@
...
+# all solutions to underdetermined (or exactly determined) Ax=b are given by A⁺b+(I-A⁺A)w for some arbitrary w
+# note that if Ax=b is overdetermined, this will raise an assertion error
+
+
+def rand_sol(A, b):
+    """Generate a random solution to the equation Ax=b."""
+    assert np.linalg.matrix_rank(A) <= len(b)
+    A_plus = np.linalg.pinv(A)
+    x = A_plus @ b
+    return x + (np.eye(x.shape[0]) - A_plus @ A) @ np.random.normal(size=x.shape)
```
kbattocchi: This isn't used, so remove it.

econml/tests/test_automated_ml.py
```diff
@@ -0,0 +1,235 @@
...
+from econml.data.dgps import ihdp_surface_B
+from azureml.train.automl.exceptions import ClientException
+from azureml.core.authentication import AzureCliAuthentication
+import os
+
+AutomatedTLearner = addAutomatedML(TLearner)
+AutomatedSLearner = addAutomatedML(SLearner)
+AutomatedXLearner = addAutomatedML(XLearner)
+AutomatedDomainAdaptationLearner = \
+    addAutomatedML(DomainAdaptationLearner)
```
kbattocchi: These don't need to be split across lines, and would probably look cleaner without this.

econml/tests/test_automated_ml.py
```diff
@@ -0,0 +1,235 @@
...
+    'experiment_timeout_minutes': 1,
+    'enable_early_stopping': True,
+    'iteration_timeout_minutes': 1,
+    'max_cores_per_iteration': 1,
+    'n_cross_validations': 2,
+    'preprocess': False,
+    'featurization': 'off',
+    'enable_stack_ensemble': False,
+    'enable_voting_ensemble': False,
+    'primary_metric': 'normalized_mean_absolute_error',
```
kbattocchi: Why MAE instead of MSE?  I believe that for correctness at least DML should be using MSE in the final stage.

econml/tests/test_automated_ml.py
```diff
@@ -0,0 +1,235 @@
...
+                                     debug_log='automl_errors.log',
+                                     enable_onnx_compatible_models=True, model_explainability=True,
+                                     **AUTOML_SETTINGS_REG)
+
+AUTOML_CONFIG_CLF = EconAutoMLConfig(task='classification',
+                                     debug_log='automl_errors.log',
+                                     enable_onnx_compatible_models=True, model_explainability=True,
+                                     **AUTOML_SETTINGS_CLF)
+
+AUTOML_CONFIG_LINEAR_REG = EconAutoMLConfig(task='regression',
```
kbattocchi: There appears to be no difference between the LINEAR and SAMPLE_WEIGHT variants defied here...

econml/tests/test_automated_ml.py
```diff
@@ -0,0 +1,235 @@
...
+        auth = AzureCliAuthentication()
+
+        setAutomatedMLWorkspace(auth=auth,
+                                subscription_id=subscription_id,
+                                resource_group=resource_group, workspace_name=workspace_name)
+
+    def test_nonparam(self):
+        Y, T, X, _ = ihdp_surface_B()
+        est = \
+            AutomatedNonParamDMLCateEstimator(model_y=automl_model_reg(),
```
kbattocchi: Final should be using sample weights, if that's not automatically enforced.

econml/tests/test_automated_ml.py
```diff
@@ -0,0 +1,235 @@
...
+            AutomatedDMLCateEstimator(model_y=automl_model_reg(),
+                                      model_t=GradientBoostingClassifier(),
+                                      model_final=RandomForestRegressor(),
+                                      featurizer=None,
+                                      discrete_treatment=True)
+        est.fit(Y, T, X)
+        _ = est.effect(X)
+
+    def test_forest_dml(self):
+        """Testing accuracy of forest DML is reasonable"""
```
kbattocchi: This doesn't test accuracy in any way; change the comment to explain what the test actually does.

econml/tests/test_automated_ml.py
```diff
@@ -0,0 +1,235 @@
...
+                                            verbose=0, min_weight_fraction_leaf=.01)
+        est.fit(Y, T, X)
+        _ = est.effect(X)
+
+
+@pytest.mark.automl
+class TestMetalearners(unittest.TestCase):
+
+    def test_TLearner(self):
+        """Tests whether the TLearner can accurately estimate constant and heterogeneous
```
kbattocchi: Again, this isn't testing accuracy.

econml/tests/test_automated_ml.py
```diff
@@ -0,0 +1,235 @@
...
+        Y, T, X, _ = ihdp_surface_B()
+        est = AutomatedTLearner(models=automl_model_reg())
+
+        # Test constant and heterogeneous treatment effect, single and multi output y
+
+        est.fit(Y, T, X)
+        _ = est.effect(X)
+
+    def test_SLearner(self):
+        """Tests whether the SLearner can accurately estimate constant and heterogeneous
```
kbattocchi: Same here.

econml/tests/test_automated_ml.py
```diff
@@ -0,0 +1,235 @@
...
+        Y, T, X, _ = ihdp_surface_B()
+        est = AutomatedSLearner(overall_model=automl_model_reg())
+
+        est.fit(Y, T, X)
+        _ = est.effect(X)
+
+        # Test heterogeneous treatment effect with multi output Y
+
+    def test_DALearner(self):
+        """Tests whether the DomainAdaptationLearner can accurately estimate constant and
```
kbattocchi: And here.

econml/tests/test_automated_ml.py
```diff
@@ -0,0 +1,235 @@
...
+                                              discrete_treatment=True)
+        est.fit(Y, T, X)
+        _ = est.effect(X)
+
+    def test_param(self):
+        Y, T, X, _ = ihdp_surface_B()
+        est = \
+            AutomatedDMLCateEstimator(model_y=automl_model_reg(),
+                                      model_t=GradientBoostingClassifier(),
+                                      model_final=RandomForestRegressor(),
```
kbattocchi: DMLCateEstimator requires a linear final model, so don't use RandomForestRegressor.  Consider using LinearDMLCateEstimator instead and not overriding the final model.

azure-pipelines-steps.yml
```diff
@@ -23,7 +23,7 @@ steps:
   condition: and(succeeded(), eq(variables['Agent.OS'], 'Linux'))
 
 # Install the package
-- script: 'python -m pip install --upgrade pip && pip install --upgrade setuptools && pip install .'
+- script: 'python -m pip install --upgrade pip && pip install --upgrade setuptools && pip install -e .[automl]'
```
kbattocchi: I don't think we need the install to be editable.
```suggestion
- script: 'python -m pip install --upgrade pip && pip install --upgrade setuptools && pip install .[automl]'
```

econml/automated_ml.py
v-keacqu: Yes I apologize for this, this was an old implementation, now, the way that users can specify whether they want to constrain their model space happens in the initialization of an EconAutoMLConfig object.

econml/tests/test_automated_ml.py
v-keacqu: Unfortunately, the only "evaluation metrics" I am able to user are spearman_correlation, normalized_root_mean_squared_error, r2_score, normalized_mean_absolute_error. I spoke to @vasilismsr about where using a specific error function should effect results and he suggest it should not. Would be interesting to investigate if we could get better performance that way however.

kbattocchi: I think minimizing normalized_root_mean_squared_error should be equivalent to minimizing MSE (it's the square root, but will have the same minimizer).  I thought that the weight trick in NonParamDML relied on MSE in a non-trivial way, but @vasilismsr would know for sure.

v-keacqu: I'll change it, I'll also rerun the notebooks with RMSE and see if it performs better.

Issue #188, opened 2019-11-23
Bad output shape for ortho_forest effect


vasilismsr:
Take the example from Section 1.1 and 1.2 from the ortho forest notebook:
https://github.com/microsoft/EconML/blob/master/notebooks/Orthogonal%20Random%20Forest%20Examples.ipynb

When T is a vector of (n,), then the effect(X) of the ContinuousOrthoForest returns a (n, 1) vector. 

When T is (n, 2) (e.g. if we just duplicate T in cell 6) then effect(X) correctly a (n,) vector.

When T is (n, 1) (e.g. if we reshape T in cell 6) it runs super slow and had to be cancelled.

PR #214, opened 2020-01-28
Added `blb` inference option to the OrthoForest

moprescu:
* Added Bootstrap of Little Bags inference to the ORF classes
* Added tests and updated notebook
* Fixed the marginal effect shape when T is a vector
* Fixed bugs and reorganized class functionality


moprescu:
Also addresses issue #188

doc/spec/estimation/forest.rst
```diff
@@ -363,8 +363,7 @@ and the `ForestLearners Jupyter notebook <https://github.com/microsoft/EconML/bl
     >>> est.fit(Y, T, W, W)
     <econml.ortho_forest.ContinuousTreatmentOrthoForest object at 0x...>
     >>> print(est.effect(W[:2]))
-    [[1. ]
-     [1.2]]
+    [1.00...  1.19...]
```
kbattocchi: If I understand correctly, the shape changing here is good because it was wrong before.

Should I be concerned that the values themselves have changed as well?

econml/ortho_forest.py
```diff
@@ -305,12 +344,17 @@ def _fit_forest(self, Y, T, X, W=None):
                 self.min_leaf_size, self.max_depth,
                 self.random_state.randint(MAX_RAND_SEED)) for s in subsample_ind)
 
-    def _get_weights(self, X_single):
-        # Calculates weights
+    def _get_weights(self, X_single, tree_slice=None):
+        # Calculates weights for an input feature vector for a slice of trees
+        # defined by the `tree_slice` tuple (start, end)
```
kbattocchi: Is it obvious that start is inclusive and end is exclusive?

(not a rhetorical question - maybe in python this is obvious)

econml/ortho_forest.py
```diff
@@ -330,7 +374,26 @@ def _get_weights(self, X_single):
...
+        return (w1 / len(tree_range), w2 / len(tree_range))
+
+    def _get_blb_indices(self, X):
+        """Get  data indices for every tree under the little bags split."""
+        # Define subsample size
+        subsample_size = X.shape[0] // 2
+        if not self.bootstrap:
+            if self.subsample_ratio > 1.0:
+                # Safety check
+                self.subsample_ratio = 1.0
```
kbattocchi: Does this merit a warning, or could it happen for reasons that a user can't control?

econml/ortho_forest.py
```diff
@@ -330,7 +374,26 @@ def _get_weights(self, X_single):
...
+            subsample_size = int(self.subsample_ratio * subsample_size)
+        subsample_ind = []
+        # Draw points to create little bags
+        for it in range(self.n_slices):
+            half_sample_inds = self.random_state.choice(
+                X.shape[0], X.shape[0] // 2, replace=False)
```
kbattocchi:
```suggestion
                X.shape[0], subsample_size, replace=False)
```

econml/ortho_forest.py
```diff
@@ -330,7 +374,26 @@ def _get_weights(self, X_single):
...
+            for _ in np.arange(it * self.slice_len, min((it + 1) * self.slice_len, self.n_trees)):
+                subsample_ind.append(half_sample_inds[self.random_state.choice(
+                    X.shape[0] // 2, subsample_size, replace=False)])
```
kbattocchi: Is this right - it's just a permutation?

```suggestion
                    subsample_size, subsample_size, replace=False)])
```

kbattocchi: If so, consider using `permutation` directly instead of `choice` to make the intent clear.

econml/ortho_forest.py
```diff
@@ -509,7 +568,8 @@ def second_stage_parameter_estimator_gen(lambda_reg):
         """
         def parameter_estimator_func(Y, T, X,
                                      nuisance_estimates,
-                                     sample_weight=None):
+                                     sample_weight=None,
+                                     X_single=None):
             """Calculate the parameter of interest for points given by (Y, T) and corresponding nuisance estimates."""
```
kbattocchi: Consider adding a description of X_single to the docstring

econml/ortho_forest.py
```diff
@@ -526,11 +586,13 @@ def parameter_estimator_func(Y, T, X,
...
             # Ridge regression estimate
-            param_estimate = np.linalg.lstsq(np.matmul(weighted_XT_res.T, XT_res) + reg,
-                                             np.matmul(weighted_XT_res.T, Y_res.reshape(-1, 1)),
-                                             rcond=None)[0].flatten()
-            # Parameter returned by LinearRegression is (d_T, )
-            return param_estimate
+            linear_coef_estimate = np.linalg.lstsq(np.matmul(weighted_XT_res.T, XT_res) + reg,
+                                                   np.matmul(weighted_XT_res.T, Y_res.reshape(-1, 1)),
+                                                   rcond=None)[0].flatten()
+            X_aug = np.append([1], X_single)
```
kbattocchi: Is this legit if X_single is None?

econml/ortho_forest.py
```diff
@@ -807,10 +857,13 @@ def parameter_estimator_func(Y, T, X,
...
             reg = lambda_reg * np.diag(diagonal)
             # Ridge regression estimate
-            param_estimate = np.linalg.lstsq(np.matmul(weighted_X_aug.T, X_aug) + reg,
-                                             np.matmul(weighted_X_aug.T, pointwise_params), rcond=None)[0].flatten()
-            # Parameter returned by LinearRegression is (d_T, )
-            return param_estimate
+            linear_coef_estimate = np.linalg.lstsq(np.matmul(weighted_X_aug.T, X_aug) + reg,
+                                                   np.matmul(weighted_X_aug.T, pointwise_params),
+                                                   rcond=None)[0].flatten()
+            X_aug = np.append([1], X_single)
```
kbattocchi: Same question about case when X_single is None.

econml/ortho_forest.py
```diff
@@ -855,3 +908,113 @@ def _check_treatment(self, T):
...
+        """
+        Fits the inference model.
+
+        This is called after the estimator's fit.
+        """
+        self._estimator = estimator
+        # Test expansion of treatment
+        # If expanded treatments are a vector, flatten const_marginal_effect_interval
+        _, T0, _ = self._estimator._expand_treatments(None, 0, 1)
+        self._T_vec = False
```
kbattocchi: I don't think you need the `if` below, this whole block could just be

```suggestion
        self._T_vec = T0.ndim == 1
```

econml/ortho_forest.py
```diff
@@ -855,3 +908,113 @@ def _check_treatment(self, T):
...
+
+    This class can only be used for inference with any estimator derived from :class:`BaseOrthoForest`.
+
+    Parameters
+    ----------
+    estimator : :class:`BaseOrthoForest`
+        Estimator to perform inference on. Must be a child class of :class:`BaseOrthoForest`.
+    """
+
+    def fit(self, estimator, *args, **kwargs):
```
kbattocchi: Should you return self at the end?

kbattocchi: (I can't remember whether we care about this at all for inference objects)

econml/ortho_forest.py
```diff
@@ -855,3 +908,113 @@ def _check_treatment(self, T):
...
+        """
+        Fits the inference model.
+
+        This is called after the estimator's fit.
+        """
+        self._estimator = estimator
```
kbattocchi: Consider asserting that the estimator is supported.  For example you might check for a `_pointwise_effect` method, which seems to be the real requirement.

econml/tests/test_orf.py
```diff
@@ -64,23 +63,24 @@ def test_continuous_treatments(self):
                           TestOrthoForest.X, TestOrthoForest.W)
         # Check that outputs have the correct shape
         out_te = est.const_marginal_effect(TestOrthoForest.x_test)
-        self.assertSequenceEqual((TestOrthoForest.x_test.shape[0], 1), out_te.shape)
+        self.assertEqual(TestOrthoForest.x_test.shape[0], out_te.shape[0])
```
kbattocchi: Note that this now tests only a single entry rather than the entire shape.

doc/spec/estimation/forest.rst
moprescu: Yes, the shape change is intentional. The values have changed because the trees are partitioned into segments ("little bags") that share data, so the same random seed gives slightly different results now.

econml/ortho_forest.py
moprescu: This change is not quite right because `subsample_size` is smaller than `X.shape[0] // 2` if subsampling is enabled (which is the default). I did make a modification for the `replace=False` argument in line 395 since it would not have done the right thing for `bootstrap=True`.

econml/ortho_forest.py
moprescu: Fixed the comments!

econml/ortho_forest.py
moprescu: No, because `subsample_size` gets updated if bootstrap=False (which is the default): `subsample_size = int(self.subsample_ratio * subsample_size)`

econml/tests/test_orf.py
moprescu: Not really, this tests that the lengths are equal since these are now flat arrays rather than column vectors.

econml/ortho_forest.py
moprescu: I think's it's good practice to return `self`. Fixed!

econml/ortho_forest.py
moprescu: Fixed!

doc/spec/estimation/forest.rst
kbattocchi: Hm, so the trees are partitioned even when `fit`'s `inference` option is `None` rather than `'blb'`?

econml/ortho_forest.py
moprescu: Added! Thanks for the suggestion!

econml/ortho_forest.py
moprescu: The correct way to handle this is to not set default values for `sample_weight` and `X_single` since those arguments must always be passed in the base class. I fixed the code to reflect that. I also added some explanations around what `X_single` means.

econml/ortho_forest.py
moprescu: Fixed! See other comment.

econml/ortho_forest.py
moprescu: Added! Thanks!

econml/ortho_forest.py
moprescu: Added a warning to reflect this change.

doc/spec/estimation/forest.rst
moprescu: Yes, otherwise we'd have two ways of partitioning the tress and it would be hard to keep track of. It doesn't affect the estimate much since that's done using all of the `2*n_trees` leaves. The only difference is which data samples a particular tree uses, i.e. groups `slice_len` trees subsample from the same `n_samples/2` samples but which `n_samples/2` are used differs from group to group.

PR #215, opened 2020-01-29
Run tests under Powershell

kbattocchi:


kbattocchi:
This is not a real PR -  it is being used to test how to modify our build for the auotml work.

PR #216, opened 2020-01-30
Basic monte carlo framework for simulating dgps and methods 

v-asatha:
Built on top of https://github.com/vsyrgkanis/plugin_regularized_estimation
Changes made to to fit econml's structure 
Documentation upcoming

### DIFF
--- a/montecarlo/mc_from_config.py
+++ b/montecarlo/mc_from_config.py
@@ -1,0 +1,15 @@
+import sys
+import argparse
+from mcpy.monte_carlo import MonteCarlo
+import importlib
+
+def monte_carlo_main():
+    parser = argparse.ArgumentParser(description='Process some integers.')
+    parser.add_argument('--config', type=str, help='config file')
+    args = parser.parse_args(sys.argv[1:])
+
+    config = importlib.import_module(args.config, __name__)
+    MonteCarlo(config.CONFIG).run()
+    
+if __name__=="__main__":
+    monte_carlo_main()
--- a/montecarlo/mcpy/monte_carlo.py
+++ b/montecarlo/mcpy/monte_carlo.py
@@ -1,0 +1,169 @@
+import os
+import sys
+import numpy as np
+from joblib import Parallel, delayed
+import joblib
+import argparse
+import importlib
+from itertools import product
+import collections
+from copy import deepcopy
+from mcpy.utils import filesafe
+from mcpy import plotting
+
+def _check_valid_config(config):
+    assert 'dgps' in config, "config dict must contain dgps"
+    assert 'dgp_opts' in config, "config dict must contain dgp_opts"
+    assert 'method_opts' in config, "config dict must contain method_opts"
+    assert 'mc_opts' in config, "config dict must contain mc_opts"
+    assert 'metrics' in config, "config dict must contain metrics"
+    assert 'methods' in config, "config dict must contain methods"
+    assert 'plots' in config, "config dict must contain plots"
+    assert 'target_dir' in config, "config must contain target_dir"
+    assert 'reload_results' in config, "config must contain reload_results"
+    assert 'n_experiments' in config['mc_opts'], "config[mc_opts] must contain n_experiments"
+    assert 'seed' in config['mc_opts'], "config[mc_opts] must contain seed"
+
+class MonteCarlo:
+
+    def __init__(self, config):
+        self.config = config
+        _check_valid_config(self.config)
+        config['param_str'] = '_'.join(['{}_{}'.format(filesafe(k), v) for k,v in self.config['mc_opts'].items()])
+        config['param_str'] += '_' + '_'.join(['{}_{}'.format(filesafe(k), v) for k,v in self.config['dgp_opts'].items()])
+        config['param_str'] += '_' + '_'.join(['{}_{}'.format(filesafe(k), v) for k,v in self.config['method_opts'].items()])
+        return
+
+    # TODO: add dgp-specific, method-specific parameters and metrics
+    def experiment(self, instance_params, seed):
+        ''' Runs an experiment on a single randomly generated instance and sample and returns
+        the parameter estimates for each method and the evaluated metrics for each method
+        '''
+        np.random.seed(seed)
+
+        param_estimates = {}
+        true_params = {}
+        for dgp_name, dgp_fn in self.config['dgps'].items():
+            data, true_param = dgp_fn(self.config['dgp_opts'][dgp_name], instance_params[dgp_name], seed)
+            true_params[dgp_name] = true_param
+            param_estimates[dgp_name] = {}
+
+            for method_name, method in self.config['methods'].items():
+                param_estimates[dgp_name][method_name] = method(data, self.config['method_opts'][method_name], seed)
+
+        return param_estimates, true_params
+
+    def run(self):
+        ''' Runs multiple experiments in parallel on randomly generated instances and samples and returns
+        the parameter estimates for each method and the evaluated metrics for each method across all
+        experiments
+        '''
+        random_seed = self.config['mc_opts']['seed']
+
+        if not os.path.exists(self.config['target_dir']):
+            os.makedirs(self.config['target_dir'])
+
+        instance_params = {}
+        for dgp_name in self.config['dgps']:
+            instance_params[dgp_name] = self.config['dgp_instance_fns'][dgp_name](self.config['dgp_opts'][dgp_name], random_seed)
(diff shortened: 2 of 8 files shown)


montecarlo/mcpy/monte_carlo.py
```diff
@@ -0,0 +1,169 @@
...
+        for plot_name, plot_fn in self.config['sweep_plots'].items():
+            if isinstance(plot_fn, dict):
+                plotting.sweep_plot(plot_key, sweep_keys, sweep_params, sweep_metrics, self.config, plot_fn)
+            else:
+                plot_fn(plot_name, sweep_keys, sweep_params, sweep_metrics, sweep_true_params, self.config)
+
+        return sweep_keys, sweep_params, sweep_metrics, sweep_true_params
+
+def monte_carlo_main():
+    parser = argparse.ArgumentParser(description='Process some integers.')
```
kbattocchi: That description seems inaccurate.

montecarlo/mc_from_config.py
```diff
@@ -0,0 +1,15 @@
+import sys
+import argparse
+from mcpy.monte_carlo import MonteCarlo
+import importlib
+
+def monte_carlo_main():
+    parser = argparse.ArgumentParser(description='Process some integers.')
```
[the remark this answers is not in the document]
kbattocchi: Also, the description here seems wrong.

montecarlo/mcpy/plotting.py
```diff
@@ -0,0 +1,244 @@
...
+                    x, y = metric_results[dgp_name][method_name][metric_name]
+                    plt.plot(x, y, label=method_name)
+                plt.xlabel("X_test")
+                plt.ylabel(metric_name)
+                plt.legend()
+                plt.show()
+                # mean_rmse = np.mean(metric_results[dgp_name][method_name]['rmse'])
+                # mean_conf_length = np.mean(metric_results[dgp_name][method_name]['conf_length'])
+                # mean_coverage = np.mean(metric_results[dgp_name][method_name]['coverage'])
+                # print("means for {}: rmse: {}, conf_length: {}, coverage: {}".format(method_name, mean_rmse, mean_conf_length, mean_coverage))
```
kbattocchi: Delete dead code.

(Unless this is super valuable, in which case add a comment explaining why it is dead and what would need to happen for it to be enabled)

montecarlo/mcpy/monte_carlo.py
```diff
@@ -0,0 +1,169 @@
...
+        random_seed = self.config['mc_opts']['seed']
+
+        if not os.path.exists(self.config['target_dir']):
+            os.makedirs(self.config['target_dir'])
+
+        instance_params = {}
+        for dgp_name in self.config['dgps']:
+            instance_params[dgp_name] = self.config['dgp_instance_fns'][dgp_name](self.config['dgp_opts'][dgp_name], random_seed)
+
+        # results_file = os.path.join(self.config['target_dir'], 'results_{}.jbl'.format(self.config['param_str']))
```
kbattocchi: Throughout this file there is lots of dead code; delete all of it unless it serves some valuable purpose, in which case add comments explaining why.

montecarlo/mcpy/plotting.py
```diff
@@ -0,0 +1,244 @@
...
+            for i in range(pdgp[m_name].shape[1]):
+                plt.subplot(n_methods, n_params, it * n_params + i + 1)
+                plt.hist(pdgp[m_name][:, i])
+                plt.title("{}[{}]. $\\mu$: {:.2f}, $\\sigma$: {:.2f}".format(m_name, i, np.mean(pdgp[m_name][:, i]), np.std(pdgp[m_name][:, i])))
+        plt.tight_layout()
+        plt.savefig(os.path.join(config['target_dir'], 'dist_dgp_{}_{}.png'.format(dgp_name, config['param_str'])), dpi=300)
+        plt.close()
+    return
+
+def plot_metrics2(param_estimates, metric_results, config):
```
kbattocchi: This isn't an intuitive name to me.

Also, add docstrings to all functions here.

montecarlo/mcpy/plotting.py
```diff
@@ -0,0 +1,244 @@
...
+        n_methods = len(list(mdgp.keys()))
+        for metric_name in next(iter(mdgp.values())).keys():
+            plt.figure(figsize=(1.5 * n_methods, 2.5))
+            plt.violinplot([mdgp[method_name][metric_name] for method_name in mdgp.keys()], showmedians=True)
+            plt.xticks(np.arange(1, n_methods + 1), list(mdgp.keys()))
+            plt.ylabel(metric_name)
+            plt.tight_layout()
+            plt.savefig(os.path.join(config['target_dir'], '{}_dgp_{}_{}.png'.format(filesafe(metric_name), dgp_name, config['param_str'])), dpi=300)
+            plt.close()
+    return
```
kbattocchi: Is there any benefit to these empty returns (here and in the other functions in this file?

If not, delete them.

montecarlo/mcpy/utils.py
```diff
@@ -0,0 +1,12 @@
+import numpy as np
+
+def cross_product(X1, X2):
```
kbattocchi: How does this differ from the existing cross_product in our utilities?  Is this just so that montecarlo doesn't directly reference anything else in econml?

montecarlo/mcpy/utils.py
```diff
@@ -0,0 +1,12 @@
+import numpy as np
```
kbattocchi: Add copyright header

montecarlo/sweep_config_dml_te.py
```diff
@@ -0,0 +1,90 @@
+import os
```
kbattocchi: Add copyright header

montecarlo/sweep_config_dml_te.py
```diff
@@ -0,0 +1,90 @@
...
+            'std': metrics.std,
+            'coverage_band': metrics.coverage_band
+        },
+        "plots": {
+            'plot1': plotting.plot_metrics,
+            'plot2': plotting.plot_visualization,
+            'plot3': plotting.plot_violin
+        },
+        "per_plots": ['coverage_band'],
+        "sweep_plots": {
```
kbattocchi: Delete dead code

montecarlo/sweep_mc_from_config.py
```diff
@@ -0,0 +1,15 @@
+import sys
```
kbattocchi: Add copyright header

montecarlo/sweep_mc_from_config.py
```diff
@@ -0,0 +1,15 @@
...
+import argparse
+from mcpy.monte_carlo import MonteCarloSweep
+import importlib
+
+def monte_carlo_main():
```
kbattocchi: Do we need this many similar entrypoints (here vs. mc_from_config vs. monte_carlo.py itself)?  Could you consolidate them?

v-asatha: Makes sense to take out the one in monte_carlo.py because the methods from there are meant to be imported, so I'll take out the main() function in that.

montecarlo/sweep_config_dml_te.py
v-asatha: I commented this out so that the experiments wouldn't take a long time to run on my laptop - so I'd just be calculating metrics for two estimators instead of four. I can uncomment things under method_opts {} but I'll still leave the third and fourth methods commented for now under methods {}

montecarlo/mcpy/utils.py
v-asatha: Vestigial piece from metrics the previous framework calculating I think; we can remove it.

montecarlo/mcpy/plotting.py
v-asatha: Many of these are from previous monte carlo code that isn't exactly extended to here,because the dgps and methods we're working with require the return types of the results to be specific to the dgp/method/metric. Thus many of these plotting functions are somewhat "deprecated". However, if someone's dgp/method/metrics fit a more basic form, these plotting functions can be used. I think I'll take them out for now and they can be added back later. It doesn't make sense to have them in here unless we have example dgps/methods/metrics that use them.  In general there's some more thinking that needs to be done for how to keep the framework general but still allow for metrics and plots to specific forms of dgp/method. The best way to do that is to just implement dgps/methods/metrics until there's a clash and one realizes the framework needs to be slightly adjusted.

PR #225, opened 2020-02-24
Fixed inference results when `cate_feature_names` is not defined

moprescu:


moprescu:
> This is great. But could you please also add a test that covers the `else` branch so that we don't regress this change in the future?

@kbattocchi I thought about that, but for that to work, we need a CATE estimator with linear final stage that will never have `cate_feature_names` defined. If you have a better idea for testing, let me know!

kbattocchi:
> > This is great. But could you please also add a test that covers the `else` branch so that we don't regress this change in the future?
> 
> @kbattocchi I thought about that, but for that to work, we need a CATE estimator with linear final stage that will never have `cate_feature_names` defined. If you have a better idea for testing, let me know!

I think that one fairly straightforward approach would be to create a very simple wrapper like:

```python3
class NoFtNamesEst:
    def __init__(self, cate_est):
         self.cate_est = clone(cate_est, safe=False)

    def __get_attr__(self, name):
        if name != 'cate_feature_names`:
            return getattr(self.cate_est, name)
        else:
            return self.__getattribute__(name)
```

and then test that you can get the inference results with either a plain DML estimator or a wrapped one.

kbattocchi:
> > This is great. But could you please also add a test that covers the `else` branch so that we don't regress this change in the future?
> 
> @kbattocchi I thought about that, but for that to work, we need a CATE estimator with linear final stage that will never have `cate_feature_names` defined. If you have a better idea for testing, let me know!

I think that one fairly straightforward approach would be to create a very simple wrapper like:

```python3
class NoFtNamesEst:
    def __init__(self, cate_est):
         self.cate_est = clone(cate_est, safe=False)

    def __get_attr__(self, name):
        if name != 'cate_feature_names`:
            return getattr(self.cate_est, name)
        else:
            return self.__getattribute__(name)
```

and then test that you can get the inference results with either a plain DML estimator or a wrapped one.

econml/tests/test_inference.py
```diff
@@ -0,0 +1,53 @@
...
+        cls.n = 1000
+        cls.d_w = 30
+        cls.d_x = 5
+        # Generate data
+        cls.X = np.random.uniform(0, 1, size=(cls.n, cls.d_x))
+        cls.W = np.random.normal(0, 1, size=(cls.n, cls.d_w))
+        cls.T = np.random.normal(0, 1, size=(cls.n, ))
+        cls.Y = np.random.normal(0, 1, size=(cls.n, ))
+
+    def test_inference_results(self):
```
kbattocchi: I'd rename to something like `test_inference_no_feature_names` or something.

moprescu: I named it this way so we can add more tests under the same method in the future. And this is just one subtest.

kbattocchi: Okay; I generally prefer to have a larger number of fine grained tests than a smaller number of broader tests where possible because it makes it quicker to track down the nature of the error when there's a failure, but I don't feel super strongly about it.

PR #226, opened 2020-02-25
Fix minor issues

kbattocchi:
Fix a few small issues raised by @moprescu, plus make matplotlib requirement more flexible where possible.

### DIFF
--- a/econml/_ortho_learner.py
+++ b/econml/_ortho_learner.py
@@ -132,7 +132,7 @@ def _crossfit(model, folds, *args, **kwargs):
             nuisances = (nuisances,)
 
         first_arr = args[0] if args else kwargs.items()[0][1]
-        return nuisances, model_list, range(first_arr.shape[0])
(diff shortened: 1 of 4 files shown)


econml/_ortho_learner.py
```diff
@@ -132,7 +132,7 @@ def predict(self, X, y, W=None):
             nuisances = (nuisances,)
 
         first_arr = args[0] if args else kwargs.items()[0][1]
-        return nuisances, model_list, range(first_arr.shape[0])
+        return nuisances, model_list, np.arange(first_arr.shape[0])
```
moprescu: I noticed that the inputs of the children classes don't get transformed to numpy arrays and this still fails when one of the inputs is a Python list (for e.g. IntentToTreatDRIV). Can we add some simple input transformation in the `_OrthoLearner` class?

kbattocchi: Done - see latest push.

PR #230, opened 2020-03-06
Add two customer scenarios case study

heimengqi:


notebooks/CustomerScenarios/Case Study - Customer Segmentation at An Online Media Company.ipynb
```diff
@@ -0,0 +1,856 @@
...
+      "text/plain": [
+       "<Figure size 720x432 with 1 Axes>"
+      ]
+     },
+     "metadata": {
+      "needs_background": "light"
+     },
+
```
kbattocchi:
```suggestion
    "From the plot above, it's clear to see that the true treatment effect is a **nonlinear** function of income, with elasticity around -1.75 when income is smaller than 1 and a small negative value when income is larger than 1. The model fits a quadratic treatment effect, which is not a great fit. But it still captures the overall trend: the elasticity is negative and people are less sensitive to the price change if they have higher income."
```

kbattocchi:
```suggestion
    "`LinearDMLCateEstimator` estimator can also return the summary of the coefficients and intercept for the final model, including point estimates, p-values and confidence intervals. From the table above, we notice that $income$ has positive effect and ${income}^2$ has negative effect, and both of them are statistically significant."
```

kbattocchi:
```suggestion
    "We notice that this model fits much better than the `LinearDMLCateEstimator`, the 90% confidence interval correctly covers the true treatment effect and captures the variation when income is around 1. Overall, the model shows that people with low income are much more sensitive to the price changes than higher income people."
```

kbattocchi:
```suggestion
    "    & = \\exp^{log(Y)} \\cdot T \\\\\n",
    "    & = \\exp^{(\\theta(X) \\cdot log(T) + f(X,W) + \\epsilon)} \\cdot T \\\\\n",
    "    & = \\exp^{(f(X,W) + \\epsilon)} \\cdot \\T^{(\\theta(X)+1)}\n",
```

kbattocchi: Maybe mention that raising prices might not be feasible in a real setting.

kbattocchi:
```suggestion
    "With the decrease of price, revenue will increase only if $\\theta(X)+1<0$. Thus, we set `sample_treatment_cost=-1` here to learn **what kinds of customers we should give a small discount to maximize the revenue**.\n",
```

kbattocchi:
```suggestion
    "Now, let us compare our policy with other baseline policies! Our model says which customers to give a small discount to, and for this experiment, we will set a discount level of 10% for those users.  Because the model is misspecified we would not expect good results with large discounts.  Here, because we know the ground truth, we can evaluate the value of this policy."
```

kbattocchi: I don't know why this and the following suggestions are targeted at the wrong lines.  Please don't commit them without verifying that they can be applied to the right place...

notebooks/CustomerScenarios/Case Study - Customer Segmentation at An Online Media Company.ipynb
```diff
@@ -0,0 +1,856 @@
...
+    "**days_visited** |W| the average number of days user visited the website per week in the past\n",
+    "**friend_count** |W| number of friends user connected in the account \n",
+    "**has_membership** |W| whether the user had membership\n",
+    "**is_US** |W| whether the user accesses the website from the US \n",
+    "**songs_purchased** |W| the average songs user purchased per week in the past\n",
+    "**income** |X| user's income\n",
+    "**price** |T| the price user was exposed during the discount season (baseline price * samll discount)\n",
+    "**demand** |Y| songs user purchased during the discount season\n",
+    "\n",
+    "**To protect the privacy of the company, we use the simulated data as an example here. The data is synthetically generated and the feature distributions don't correspond to real distributions. However, the feature names have preserved their names and meaning.*\n",
```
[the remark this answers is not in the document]
heimengqi: Nope, there is no real company is based on, but we assume this is a real case?

PR #232, opened 2020-03-06
Fixed SparseDML and bootstrap bug.

moprescu:


econml/tests/test_dml.py
```diff
@@ -117,7 +117,8 @@ def make_random(is_discrete, d):
                                                                    fit_cate_intercept=fit_cate_intercept,
                                                                    discrete_treatment=is_discrete),
                                       True,
-                                      [None, 'debiasedlasso']),
+                                      [None, 'debiasedlasso'] +
+                                      [BootstrapInference(n_bootstrap_samples=20)] if not is_discrete else []),
```
kbattocchi: Why only use bootstrap for discrete treatments?  Also, I think the grouping here is wrong: `[a] + [b] if cond else []` evaluates to either `[a;b]` or `[]` so you need to use parentheses.

moprescu: The intention was to add bootstrap only when the treatment is continuous, i.e. `not is_discrete`. It fails otherwise for DML methods and this is encoded in a TODO above. I will fix the parentheses!

PR #235, opened 2020-03-06
Add mirror for statsmodels docs

kbattocchi:


kbattocchi:
Looks like we're not the only ones having problems, although it's not widespread: https://github.com/statsmodels/statsmodels/issues/6572.

PR #236, opened 2020-03-25
Enable stratification in bootstrap

### LINKED ISSUE #89: Enable a greater variety of bootstrap estimates
As @vasilismsr pointed out in his review of #75, we are currently using percentile rather than pivot bootstrap intervals.  We should implement pivot intervals as the default instead, and consider adding additional methods (e.g. bootstrap t-intervals as described in https://web.as.uky.edu/statistics/users/pbreheny/621/F12/notes/9-18.pdf)

### LINKED ISSUE #125: Bootstrap must be stratified for discrete treatments
If we have a discrete treatment and we use bootstrap, then it's possible that some subsamples will not have the full set of treatments, in which case some of our code will break.  This becomes more likely as the number of replicates increases or the number of samples (or number of samples with one particular treatment) decreases.

Instead, we need to implement a variant of bootstrap that stratifies by treatment, so that we are sure to keep some examples with each treatment in each bootstrap sample.

kbattocchi:
Fixes #125.  Fixes #89.

### DIFF
--- a/econml/_ortho_learner.py
+++ b/econml/_ortho_learner.py
@@ -494,6 +494,22 @@ class _OrthoLearner(TreatmentExpansionMixin, LinearCateEstimator):
                 non_none_kwargs[key] = value
         return non_none_kwargs
 
+    def _strata(self, Y, T, X=None, W=None, Z=None, sample_weight=None, sample_var=None):
+        if self._discrete_instrument:
+            Z = LabelEncoder().fit_transform(np.ravel(Z))
+
+        if self._discrete_treatment:
+            enc = LabelEncoder()
+            T = enc.fit_transform(np.ravel(T))
+            if self._discrete_instrument:
+                return T + Z * len(enc.classes_)
+            else:
+                return T
+        elif self._discrete_instrument:
+            return Z
+        else:
+            return None
+
     @BaseCateEstimator._wrap_fit
     def fit(self, Y, T, X=None, W=None, Z=None, sample_weight=None, sample_var=None, *, inference=None):
         """
@@ -540,6 +556,9 @@ class _OrthoLearner(TreatmentExpansionMixin, LinearCateEstimator):
     def _fit_nuisances(self, Y, T, X=None, W=None, Z=None, sample_weight=None):
         # use a binary array to get stratified split in case of discrete treatment
         stratify = self._discrete_treatment or self._discrete_instrument
+        strata = self._strata(Y, T, X=X, W=W, Z=Z, sample_weight=sample_weight)
+        if strata is None:
+            strata = T  # always safe to pass T as second arg to split even if we're not actually stratifying
 
         if self._discrete_treatment:
             T = self._one_hot_encoder.fit_transform(reshape(T, (-1, 1)))
@@ -547,20 +566,12 @@ class _OrthoLearner(TreatmentExpansionMixin, LinearCateEstimator):
         if self._discrete_instrument:
             z_enc = LabelEncoder()
             Z = z_enc.fit_transform(Z.ravel())
-
-            if self._discrete_treatment:  # need to stratify on combination of Z and T
-                to_split = inverse_onehot(T) + Z * len(self._one_hot_encoder.categories_[0])
-            else:
-                to_split = Z  # just stratify on Z
-
             z_ohe = OneHotEncoder(categories='auto', sparse=False, drop='first')
             Z = z_ohe.fit_transform(reshape(Z, (-1, 1)))
             self.z_transformer = FunctionTransformer(
                 func=_EncoderWrapper(z_ohe, z_enc).encode,
                 validate=False)
         else:
-            # stratify on T if discrete, and fine to pass T as second arg to KFold.split even when not
-            to_split = inverse_onehot(T) if self._discrete_treatment else T
             self.z_transformer = None
 
         if self._n_splits == 1:  # special case, no cross validation
@@ -575,9 +586,9 @@ class _OrthoLearner(TreatmentExpansionMixin, LinearCateEstimator):
             all_vars = [var if np.ndim(var) == 2 else var.reshape(-1, 1) for var in [Z, W, X] if var is not None]
             if all_vars:
                 all_vars = np.hstack(all_vars)
-                folds = splitter.split(all_vars, to_split)
+                folds = splitter.split(all_vars, strata)
             else:
-                folds = splitter.split(np.ones((T.shape[0], 1)), to_split)
+                folds = splitter.split(np.ones((T.shape[0], 1)), strata)
 
         if self._discrete_treatment:
             self._d_t = shape(T)[1:]
--- a/econml/bootstrap.py
+++ b/econml/bootstrap.py
@@ -46,7 +46,8 @@ class BootstrapEstimator:
         This option only affects behavior if `compute_means` is set to ``True``.
     """
 
-    def __init__(self, wrapped, n_bootstrap_samples=1000, n_jobs=None, compute_means=True, prefer_wrapped=False):
+    def __init__(self, wrapped, n_bootstrap_samples=1000, n_jobs=None,
+                 compute_means=True, prefer_wrapped=False):
         self._instances = [clone(wrapped, safe=False) for _ in range(n_bootstrap_samples)]
         self._n_bootstrap_samples = n_bootstrap_samples
         self._n_jobs = n_jobs
@@ -55,21 +56,49 @@ class BootstrapEstimator:
 
     # TODO: Add a __dir__ implementation?
 
+    @staticmethod
+    def __stratified_indices(arr):
+        assert 1 <= np.ndim(arr) <= 2
+        unique = np.unique(arr, axis=0)
+        indices = []
+        for el in unique:
+            ind, = np.where(np.all(arr == el, axis=1) if np.ndim(arr) == 2 else arr == el)
+            indices.append(ind)
+        return indices
+
     def fit(self, *args, **named_args):
         """
         Fit the model.
 
         The full signature of this method is the same as that of the wrapped object's `fit` method.
         """
-        n_samples = np.shape(args[0] if args else named_args[(*named_args,)[0]])[0]
-        indices = np.random.choice(n_samples, size=(self._n_bootstrap_samples, n_samples), replace=True)
+        from .cate_estimator import BaseCateEstimator  # need to nest this here to avoid circular import
+
+        index_chunks = None
+        if isinstance(self._instances[0], BaseCateEstimator):
+            index_chunks = self._instances[0]._strata(*args, **named_args)
+            if index_chunks is not None:
+                index_chunks = self.__stratified_indices(index_chunks)
+        if index_chunks is None:
+            n_samples = np.shape(args[0] if args else named_args[(*named_args,)[0]])[0]
+            index_chunks = [np.arange(n_samples)]  # one chunk with all indices
+
+        indices = []
+        for chunk in index_chunks:
+            n_samples = len(chunk)
+            indices.append(chunk[np.random.choice(n_samples,
+                                                  size=(self._n_bootstrap_samples, n_samples),
+                                                  replace=True)])
+
+        indices = np.hstack(indices)
 
         def fit(x, *args, **kwargs):
             x.fit(*args, **kwargs)
             return x  # Explicitly return x in case fit fails to return its target
 
         def convertArg(arg, inds):
-            return arg[inds] if arg is not None else None
+            return np.asarray(arg)[inds] if arg is not None else None
+
         self._instances = Parallel(n_jobs=self._n_jobs, prefer='threads', verbose=3)(
             delayed(fit)(obj,
                          *[convertArg(arg, inds) for arg in args],
@@ -84,6 +113,11 @@ class BootstrapEstimator:
 
         Additionally, the suffix "_interval" is supported for getting an interval instead of a point estimate.
         """
+
+        # don't proxy special methods
+        if name.startswith('__'):
+            raise AttributeError(name)
+
         def proxy(make_call, name, summary):
             def summarize_with(f):
                 return summary(np.array(Parallel(n_jobs=self._n_jobs, prefer='threads', verbose=3)(
--- a/econml/bootstrap.py
+++ b/econml/bootstrap.py
@@ -5,6 +5,9 @@
 import numpy as np
 from joblib import Parallel, delayed
 from sklearn.base import clone
+from scipy.stats import norm
+from collections import OrderedDict
+import pandas as pd
 
 
 class BootstrapEstimator:
@@ -44,15 +47,23 @@ class BootstrapEstimator:
         In case a method ending in '_interval' exists on the wrapped object, whether
         that should be preferred (meaning this wrapper will compute the mean of it).
         This option only affects behavior if `compute_means` is set to ``True``.
+
+    bootstrap_type: 'percentile', 'pivot', or 'normal', default 'pivot'
+        Bootstrap method used to compute results.  'percentile' will result in using the empiracal CDF of
+        the replicated computations of the statistics.   'pivot' will also use the replicates but create a pivot
+        interval that also relies on the estimate over the entire dataset.  'normal' will instead compute an interval
+        assuming the replicates are normally distributed.
     """
 
-    def __init__(self, wrapped, n_bootstrap_samples=1000, n_jobs=None,
-                 compute_means=True, prefer_wrapped=False):
+    def __init__(self, wrapped, n_bootstrap_samples=1000, n_jobs=None, compute_means=True, prefer_wrapped=False,
+                 bootstrap_type='pivot'):
         self._instances = [clone(wrapped, safe=False) for _ in range(n_bootstrap_samples)]
         self._n_bootstrap_samples = n_bootstrap_samples
         self._n_jobs = n_jobs
         self._compute_means = compute_means
         self._prefer_wrapped = prefer_wrapped
+        self._bootstrap_type = bootstrap_type
+        self._wrapped = wrapped
 
     # TODO: Add a __dir__ implementation?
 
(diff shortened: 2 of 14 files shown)


kbattocchi:
Fixes #89

vsyrgkanis:
The point estimate in the bootstrap should be the estimate on the original set of samples, not the mean of the bootstrap subsamples.

econml/bootstrap.py
```diff
@@ -5,6 +5,254 @@
 import numpy as np
 from joblib import Parallel, delayed
 from sklearn.base import clone
+from scipy.stats import norm
+from collections import OrderedDict
+import pandas as pd
+
+
+class BootstrapInferenceResults:
```
vsyrgkanis, line 13: Why isn't this a child of the InferenceResults and re-use a lot of the code there?

vsyrgkanis: In fact InferenceResults already contain an input param pred_dist in anticipation of filling in the bootstrap. Not sure what the best merging is, but the two classes need to be merged somehow or be under the same inheritance family to re-use a lot of the logic.

econml/bootstrap.py
```diff
@@ -97,16 +387,40 @@ def call(*args, **kwargs):
...
             def call_with_bounds(can_call, lower, upper):
-                return proxy(can_call, prefix,
-                             lambda arr: (np.percentile(arr, lower, axis=0), np.percentile(arr, upper, axis=0)))
+                def percentile_bootstrap(arr, _):
+                    return np.percentile(arr, lower, axis=0), np.percentile(arr, upper, axis=0)
+
+                def pivot_bootstrap(arr, _):
+                    # TODO: do we want the central estimate to be the average of all bootstrap estimates,
+                    #       or the original estimate over the entire non-bootstrapped population?
+                    est = np.mean(arr, axis=0)
```
vsyrgkanis, line 409: Again here the point estimate is the original estimate. This should be calling sth not re-doing the point estimate.

econml/bootstrap.py
```diff
@@ -97,16 +387,40 @@ def call(*args, **kwargs):
...
+                    return 2 * est - np.percentile(arr, upper, axis=0), 2 * est - np.percentile(arr, lower, axis=0)
+
+                def normal_bootstrap(arr, _):
+                    est = np.mean(arr, axis=0)
```
vsyrgkanis, line 413: Same here about point estimate

econml/inference.py
```diff
@@ -45,26 +45,69 @@ class BootstrapInference(Inference):
...
+            def wrapped(*args, alpha=0.1, **kwargs):
+                return m(*args, lower=100 * alpha / 2, upper=100 * (1 - alpha / 2), **kwargs)
+            return wrapped
+        else:
+            return m
 
-        def wrapped(*args, alpha=0.1, **kwargs):
-            return m(*args, lower=100 * alpha / 2, upper=100 * (1 - alpha / 2), **kwargs)
-        return wrapped
+    def summary(self, alpha=0.1, value=0, decimals=3, feat_name=None):
```
vsyrgkanis, line 84: Isn't this piece of code already implemented in LinearModelFinalInference?

vsyrgkanis (changes requested):

econml/bootstrap.py
```diff
@@ -5,6 +5,254 @@
...
+
+        Returns
+        -------
+        prediction : array-like, shape (m, d_y, d_t) or (m, d_y)
+            The point estimate of each treatment on each outcome for each sample X[i].
+            Note that when Y or T is a vector rather than a 2-dimensional array,
+            the corresponding singleton dimensions in the output will be collapsed
+            (e.g. if both are vectors, then the output of this method will also be a vector)
+        """
+        return np.mean(self.pred_dist, axis=0)
```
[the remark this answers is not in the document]
vsyrgkanis: For instance calling coef_ of the estimator should give the same result as the point estimate printed by the summary. 
```
X = np.random.normal(0, 1, size=(100, 1))
T = np.random.normal(0, 1, size=(100,))
Y = X[:, 0]*T + np.random.normal(0, 1, size=(100,))
est = LinearDMLCateEstimator(model_y=RandomForestRegressor(),
                             model_t=RandomForestRegressor(),
                             random_state=123)
est.fit(Y, T, X, None, inference='bootstrap')
print(est.coef_)
est.summary()
```

And similarly for all other quantities, the point estimate should be matching what is produced by calling effect(X), const_marginal_effect(X), marginal_effect(X), coef_, intercept_ of the original estimator.

econml/bootstrap.py
```diff
@@ -120,19 +434,62 @@ def call(lower=5, upper=95):
                     return call_with_bounds(can_call, lower, upper)
                 return call
 
+        def get_inference():
+            # can't import from econml.inference at top level without creating mutual dependencies
+            from .inference import InferenceResults
+
+            prefix = name[: - len("_inference")]
+            if prefix in ['const_marginal_effect', 'effect']:
```
vsyrgkanis, line 442: Why not also marginal_effect? Is there another way of doing this that avoids keeping track of all the methods that each cate estimator is implementing. Maybe for now it's ok to just add marginal_effect to this list.

vsyrgkanis (changes requested):

econml/bootstrap.py
```diff
@@ -5,6 +5,254 @@
...
+            Note that when Y or T is a vector rather than a 2-dimensional array,
+            the corresponding singleton dimensions in the output will be collapsed
+            (e.g. if both are vectors, then the output of this method will also be a vector)
+        """
+
+        if self.kind == 'percentile':
+            dist = self.pred_dist
+        elif self.kind == 'pivot':
+            est = np.mean(self.pred_dist, axis=0)
+            dist = 2 * est - pred_dist
```
vsyrgkanis, line 140: This should be self.pred_dist.  Also this means we need some extra test to capture this flow path, since the following piece of code raises an error:
```
from econml.inference import BootstrapInference
X = np.random.normal(0, 1, size=(100, 1))
T = np.random.normal(0, 1, size=(100,))
Y = .01 * X[:, 0]*T + np.random.normal(0, 1, size=(100,))
est = LinearDMLCateEstimator(model_y=RandomForestRegressor(),
                             model_t=RandomForestRegressor(),
                             random_state=123)
est.fit(Y, T, X, None, inference=BootstrapInference(bootstrap_type='pivot'))
est.coef__inference().summary_frame()
```
Or similarly if you call: ```est.summary()``` in the end.

vsyrgkanis (changes requested):

econml/bootstrap.py
```diff
@@ -5,6 +5,254 @@
...
+        Number of outputs
+    inf_type: string
+        The type of inference result.
+        It could be either 'effect', 'coefficient' or 'intercept'.
+    fname_transformer: None or predefined function
+        The transform function to get the corresponding feature names from featurizer
+    """
+
+    def __init__(self, pred_dist, kind, d_y, d_t, inf_type, fname_transformer):
+        self.pred_dist = pred_dist
```
vsyrgkanis, line 37: Wouldnt' it suffice if you check whether kind='pivot' or 'percentile' and simply changed the pred_dist here? i.e. if pivot then set `self.pred_dist = 2*point_estimate - pred_dist`. Then you don't need to check for pivot or percentile in each of the calculations. 

Also it seems that similar to inference results, point_estimate should just be another input parameter provided by the cate estimator class as input to the inference results and not computed here as a function of pred_dist.

vsyrgkanis (changes requested):

econml/inference.py
```diff
@@ -45,26 +45,69 @@ class BootstrapInference(Inference):
...
 
-    def __init__(self, n_bootstrap_samples=100, n_jobs=-1):
+    def __init__(self, n_bootstrap_samples=100, n_jobs=-1, bootstrap_type='percentile'):
         self._n_bootstrap_samples = n_bootstrap_samples
         self._n_jobs = n_jobs
+        self._bootstrap_type = bootstrap_type
 
     def fit(self, estimator, *args, **kwargs):
-        est = BootstrapEstimator(estimator, self._n_bootstrap_samples, self._n_jobs, compute_means=False)
+        discrete_treatment = estimator._discrete_treatment if hasattr(estimator, '_discrete_treatment') else False
```
vsyrgkanis, line 62: This seems to be an unused variable?

econml/bootstrap.py
```diff
@@ -5,6 +5,254 @@
...
+        Returns
+        -------
+        lower, upper: tuple of arrays, shape (m, d_y, d_t) or (m, d_y)
+            The lower and the upper bounds of the confidence interval for each quantity.
+            Note that when Y or T is a vector rather than a 2-dimensional array,
+            the corresponding singleton dimensions in the output will be collapsed
+            (e.g. if both are vectors, then the output of this method will also be a vector)
+        """
+        lower = alpha / 2
+        upper = (1 - alpha) / 2
```
vsyrgkanis, line 108: This should be 1 - alpha/2, without the parenthesis. Otherwise currently the confidence intervals produced by inference results are too narrow and wrong. Another reason why we might want to re-use logic between the existing implementation of inferenceresults and this one.

vsyrgkanis (changes requested):

econml/inference.py
```diff
@@ -45,26 +45,69 @@ class BootstrapInference(Inference):
...
+    bootstrap_type: 'percentile', 'pivot', or 'normal', default 'percentile'
+        Bootstrap method used to compute results.
+        'percentile' will result in using the empiracal CDF of the replicated copmutations of the statistics.
+        'pivot' will also use the replicates but create a pivot interval that also relies on the estimate
+        over the entire dataset.
+        'normal' will instead compute a pivot interval assuming the replicates are normally distributed.
     """
 
-    def __init__(self, n_bootstrap_samples=100, n_jobs=-1):
+    def __init__(self, n_bootstrap_samples=100, n_jobs=-1, bootstrap_type='percentile'):
```
vsyrgkanis, line 56: I think default should be pivot.

vsyrgkanis (changes requested):

econml/_ortho_learner.py
```diff
@@ -494,6 +494,22 @@ def _filter_none_kwargs(self, **kwargs):
                 non_none_kwargs[key] = value
         return non_none_kwargs
 
+    def _strata(self, Y, T, X=None, W=None, Z=None, sample_weight=None, sample_var=None):
```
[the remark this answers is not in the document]
kbattocchi: Yes, but there's a default implementation that just returns None, so it is only overridden here in _ortho_learner

econml/inference.py
kbattocchi: That would be better, but is not backwards compatible.  Since it's not a big behavior change (and it's almost certainly an improvement), I'm happy to make the change if you think that's more important.

vsyrgkanis: Yeah I would switch to pivot as it's a better approach.

econml/_ortho_learner.py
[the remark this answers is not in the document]
vsyrgkanis: Sounds good!

moprescu (approved):

econml/bootstrap.py
```diff
@@ -5,6 +5,254 @@
...
+        """
+
+        if self.kind == 'percentile':
+            dist = self.pred_dist
+        elif self.kind == 'pivot':
+            est = np.mean(self.pred_dist, axis=0)
+            dist = 2 * est - pred_dist
+        else:
+            raise ValueError("Unrecognized bootstrap kind; valid kinds are 'percentile' and 'pivot'")
+        return min((dist < value).sum(), (dist > value).sum()) / dist.shape[0]
```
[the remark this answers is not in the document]
kbattocchi: That should only be the case if we choose to make a parametric assumption (that is, if bootstrap_type is 'normal'); otherwise we want to use a non-parametric calculation since the statistic we're bootstrapping may not be normally distributed even asymptotically.

econml/_ortho_learner.py
```diff
@@ -540,27 +556,22 @@ def fit(self, Y, T, X=None, W=None, Z=None, sample_weight=None, sample_var=None,
     def _fit_nuisances(self, Y, T, X=None, W=None, Z=None, sample_weight=None):
         # use a binary array to get stratified split in case of discrete treatment
         stratify = self._discrete_treatment or self._discrete_instrument
+        strata = self._strata(Y, T, X=X, W=W, Z=Z, sample_weight=sample_weight)
```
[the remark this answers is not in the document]
kbattocchi: It could be.  For example, if DML's fit method is called with bootstrap inference and sample_var is passed there, then it will also be passed through to _strata by the bootstrap code (which just gets a set of args and kwargs and passes them through directly).

econml/bootstrap.py
kbattocchi: This is fixed in the latest update.

econml/bootstrap.py
kbattocchi: Yes, this is fixed in the latest update.  There is now an `EmpiricalInferenceResults` subclass that takes the distribution as an input and uses it as-is; the calling code in bootstrap is now responsible for doing the pivot logic itself before passing the array in.

econml/bootstrap.py
kbattocchi: The hierarchy is now much improved and shares most logic.

econml/bootstrap.py
kbattocchi: Fixed

econml/bootstrap.py
```diff
@@ -5,6 +5,254 @@
...
+            The p value of of each treatment on each outcome for each sample X[i].
+            Note that when Y or T is a vector rather than a 2-dimensional array,
+            the corresponding singleton dimensions in the output will be collapsed
+            (e.g. if both are vectors, then the output of this method will also be a vector)
+        """
+
+        if self.kind == 'percentile':
+            dist = self.pred_dist
+        elif self.kind == 'pivot':
+            est = np.mean(self.pred_dist, axis=0)
```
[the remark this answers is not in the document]
kbattocchi: Fixed

econml/bootstrap.py
```diff
@@ -5,6 +5,254 @@
...
+        -------
+        zstat : array-like, shape (m, d_y, d_t) or (m, d_y)
+            The z statistic of the metric of each treatment on each outcome for each sample X[i].
+            Note that when Y or T is a vector rather than a 2-dimensional array,
+            the corresponding singleton dimensions in the output will be collapsed
+            (e.g. if both are vectors, then the output of this method will also be a vector)
+        """
+        return (self.point_estimate - value) / self.stderr
+
+    def summary_frame(self, alpha=0.1, value=0, decimals=3, feat_name=None):
```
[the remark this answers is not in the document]
kbattocchi: Fixed

econml/bootstrap.py
```diff
@@ -44,32 +292,69 @@ class BootstrapEstimator:
         In case a method ending in '_interval' exists on the wrapped object, whether
         that should be preferred (meaning this wrapper will compute the mean of it).
         This option only affects behavior if `compute_means` is set to ``True``.
+
+    bootstrap_type: 'percentile', 'pivot', or 'normal', default 'percentile'
+        Bootstrap method used to compute results.  'percentile' will result in using the empiracal CDF of
+        the replicated copmutations of the statistics.   'pivot' will also use the replicates but create a pivot
```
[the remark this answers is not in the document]
kbattocchi: Fixed

econml/bootstrap.py
kbattocchi: Fixed.

econml/bootstrap.py
kbattocchi: Fixed by adding marginal effect explicitly; the issue is that the logic for how to display coefficients vs. effects vs. intercepts is different and we don't have any good way to know which one to use generically.

econml/inference.py
kbattocchi: This will be fixed in an update I'm about to push.

econml/inference.py
kbattocchi: Yes, this is not actually needed for stratification (because we peek at the estimator's strata so that we can also handle discrete instruments).  It will be removed in the next update.

econml/bootstrap.py
kbattocchi: Will be fixed in next update.

econml/bootstrap.py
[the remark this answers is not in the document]
kbattocchi: Fixed.

econml/inference.py
kbattocchi: Fixed by adding a mixin that both classes use.

econml/cate_estimator.py
```diff
@@ -392,8 +392,11 @@ def marginal_effect_inference(self, T, X=None):
         if X is None:
             pred = np.repeat(pred, shape(T)[0], axis=0)
             pred_stderr = np.repeat(pred_stderr, shape(T)[0], axis=0)
-        return InferenceResults(d_t=cme_inf.d_t, d_y=cme_inf.d_y, pred=pred,
-                                pred_stderr=pred_stderr, inf_type='effect', pred_dist=None, fname_transformer=None)
+        # TODO: It seems wrong to return inference results based on a normal approximation
+        #       even in the case where const_marginal_effect_inference was actually generated
+        #       using bootstrap
+        return NormalInferenceResults(d_t=cme_inf.d_t, d_y=cme_inf.d_y, pred=pred,
```
vsyrgkanis, line 398: This should be checking what the inference type is and return the same type as what cme_inf is. So if it is EmpiricalInference, return EmpiricalInference.

vsyrgkanis (approved):
Small change in marginal_effect_inference logic at the LinearCateEstimator level is required.

econml/cate_estimator.py
kbattocchi: Fixed in the latest commit; now `const_marginal_effect_interval` and `const_marginal_effect_inference` both take an additional optional named argument `n_row` and they are responsible for sizing the output accordingly; `marginal_effect_interval` (and inference) then just defer to that implementation and don't need to worry about any resizing themselves.

vsyrgkanis (commented):
Are these really the only places where const_marginal_effect_interval and inference are defined, or are there many other places where the signature needs to change??

vsyrgkanis (approved):
Looks great!

PR #247, opened 2020-04-09
[WIP] Safe Upgrade to Modern TensorFlow 2.x

arose13:
By the way, is it really the case that there are no tests in this project for the neural network components or am I missing something

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,7 +10,7 @@ jobs:
 - job: 'EvalChanges'
   displayName: 'Analyze changed files to determine which job to run'
   pool:
-    vmImage: 'macOS-10.13'
+    vmImage: 'macOS-10.15'
   steps:
   # We want to enforce the following rules for PRs:
   # * if all modifications are to README.md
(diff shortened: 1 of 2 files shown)


arose13:
Nevermind I found the tests. I guess you guys have a good reason why the test are located inside the `econml` rather than a `test` folder.

kbattocchi:
> Nevermind I found the tests. I guess you guys have a good reason why the test are located inside the `econml` rather than a `test` folder.

Not sure about "good" reason, but it's one of the two major ways people organize python tests (see, e.g. https://docs.pytest.org/en/latest/goodpractices.html#tests-as-part-of-application-code).  I can't remember if we had a specific reason to choose this rather than putting the tests beside the package when starting the project, but it has worked fine for us so far.

By the way, please cherry pick 0f5ddfef0c52764 into your branch so that the automated tests can run - Azure Pipelines has deprecated one of VM images that we were using before, which is why the checks are failing immediately.

kbattocchi:
I apologize for leaving this hanging - it turns out that many of our other dependencies were also pretty limited (e.g. we had capped sklearn < 0.22 and didn't support Python 3.8 at all), so I incorporated loosening our tensorflow restriction into that set of changes, which has now been merged (as #210).  We hope to get a new release out with these changes as well as a few other PRs within the next few weeks.  Thanks for suggesting the change and contributing the PR, but I'm going to close it now since the other one supersedes it.

PR #256, opened 2020-06-05
Fix build issues

kbattocchi:
Some changes to the Azure Pipelines environment, sphinx, and pycodestyle have necessitated a few changes to our build.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,7 +10,7 @@ jobs:
 - job: 'EvalChanges'
   displayName: 'Analyze changed files to determine which job to run'
   pool:
-    vmImage: 'macOS-10.13'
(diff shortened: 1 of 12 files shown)


kbattocchi:
This PR does 3 things to fix our build:
* update our pipeline definition to avoid deprecated machine types
* fix some linting issues newly flagged by recent versions of pycodestyle
* fix some issues newly flagged by recent versions of sphinx
  * One is that intersphinx now wants links to sklearn's `cv splitter` term to use exactly this casing
  * Another is that in two places we explicitly documented an attribute that we inherited from an sklearn parent class, but which was listed as a property there.  I have removed our documentation for now but will investigate if there is an easy way to prefer our own content instead.

PR #258, opened 2020-06-09
Fixed SubsampledHonestForest randomness

moprescu:


econml/sklearn_extensions/ensemble.py
```diff
@@ -284,6 +284,12 @@ class SubsampledHonestForest(ForestRegressor, RegressorMixin):
     estimators_ : list of DecisionTreeRegressor
         The collection of fitted sub-estimators.
 
+    feature_importances_ : array of shape = [n_features]
+        The feature importances (the higher, the more important the feature).
+        Feature importances are calculate based on the reduction in the
+        splitting criterion among the split samples (not the estimation samples).
+        So it might contain some upwards bias.
+
```
kbattocchi, line 292: This was removed as part of my big bugfix PR because it conflicts with the parent's property documentation.  I believe it will cause a sphinx issue if not removed again.

moprescu: Fixed!

PR #283, opened 2020-09-10
Vasilis/dmlcoef

vsyrgkanis:
Added coef_ and intercept_ to DMLCateEstimator. This can also enable bootstrap inference of the coef and the intercept.

### DIFF
--- a/econml/tests/test_dml.py
+++ b/econml/tests/test_dml.py
@@ -180,8 +180,7 @@ class TestDML(unittest.TestCase):
                                             eff = est.effect(X, T0=T0, T1=T)
                                             self.assertEqual(shape(eff), effect_shape)
 
-                                            if isinstance(est, (DMLCateEstimator,
-                                                                LinearDMLCateEstimator, SparseLinearDMLCateEstimator)):
+                                            if not isinstance(est, KernelDMLCateEstimator):
                                                 self.assertEqual(shape(est.coef_), coef_shape)
                                                 if fit_cate_intercept:
                                                     self.assertEqual(shape(est.intercept_), intercept_shape)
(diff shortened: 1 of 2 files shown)


econml/tests/test_dml.py
```diff
@@ -171,7 +180,8 @@ def make_random(n, is_discrete, d):
                                             eff = est.effect(X, T0=T0, T1=T)
                                             self.assertEqual(shape(eff), effect_shape)
 
-                                            if isinstance(est, LinearDMLCateEstimator) or\
+                                            if isinstance(est, DMLCateEstimator) or\
+                                                    isinstance(est, LinearDMLCateEstimator) or\
                                                     isinstance(est, SparseLinearDMLCateEstimator):
```
kbattocchi, line 185: Minor refactoring: `isinstance`'s second argument can be a tuple, so we can get rid of the `or`s:
```suggestion
                                            if isinstance(est, (DMLCateEstimator, LinearDMLCateEstimator, SparseLinearDMLCateEstimator)):
```

econml/tests/test_dml.py
```diff
@@ -190,6 +200,8 @@ def make_random(n, is_discrete, d):
                                                 self.assertEqual(shape(est.effect_interval(X, T0=T0, T1=T)),
                                                                  (2,) + effect_shape)
                                                 if (isinstance(est,
+                                                               DMLCateEstimator) or
+                                                    isinstance(est,
                                                                LinearDMLCateEstimator) or
                                                     isinstance(est,
                                                                SparseLinearDMLCateEstimator)):
```
kbattocchi, line 207: Same here.
```suggestion
                                                if isinstance(est,
                                                               (DMLCateEstimator, LinearDMLCateEstimator, SparseLinearDMLCateEstimator)):
```

econml/tests/test_dml.py
```diff
@@ -268,6 +280,8 @@ def make_random(n, is_discrete, d):
 
                                                 # test coef__inference and intercept__inference
                                                 if (isinstance(est,
+                                                               DMLCateEstimator) or
+                                                    isinstance(est,
                                                                LinearDMLCateEstimator) or
                                                     isinstance(est,
                                                                SparseLinearDMLCateEstimator)):
```
kbattocchi, line 287: And here:
```suggestion
                                                if isinstance(est, (DMLCateEstimator, LinearDMLCateEstimator, SparseLinearDMLCateEstimator)):
```

kbattocchi (changes requested):
This change looks straightforward, but the `test_cate_api` test is now failing because an output shape is unexpected - can you investigate why this might be the case?

kbattocchi (approved):

PR #290, opened 2020-10-13
Added new customer scenario notebooks

moprescu:
* Added multi-investment attribution customer scenario, EconML only
* Added multi-investment attribution customer scenario, EconML + DoWhy
* Updated images in the other customer scenario notebooks.

### DIFF
--- a/notebooks/CustomerScenarios/Case Study - Customer Segmentation at An Online Media Company - EconML + DoWhy.ipynb
+++ b/notebooks/CustomerScenarios/Case Study - Customer Segmentation at An Online Media Company - EconML + DoWhy.ipynb
@@ -4,14 +4,14 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "<img src=\"https://www.productboard.com/wp-content/uploads/2019/11/screen-shot-2019-11-07-at-4-18-51-pm.png\" />"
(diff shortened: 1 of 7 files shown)


GregLewis (approved):
This all looks good, Miruna.  Some thoughts:

1) I'd put robustness at the end, after interpretability.  Better to show them how to get treatment effects first, before we show that they're robust.  I'd highlight EconML before DoWhy (but the causal graph is neat).

2) What is DoWhy doing, exactly, when they "add an unobserved confounder"? I'm curious...

moprescu:
> This all looks good, Miruna. Some thoughts:
> 
> 1. I'd put robustness at the end, after interpretability.  Better to show them how to get treatment effects first, before we show that they're robust.  I'd highlight EconML before DoWhy (but the causal graph is neat).
> 2. What is DoWhy doing, exactly, when they "add an unobserved confounder"? I'm curious...

@GregLewis 

1. Good idea about changing the ordering. It's done.
2. DoWhy modifies the treatment and outcome according to an unobserved continuous confounder. For binary treatment/outcome, it performs random binary flips with some probability (given by the strength parameter) and for continuous treatment/outcome, the confounder is added to the treatment/outcome with a scaling constant given by the strength parameter.

PR #299, opened 2020-11-02
Fix bootstrap inference

kbattocchi:


econml/bootstrap.py
```diff
@@ -212,21 +220,31 @@ def get_dist(est, arr):
...
                                               pred_stderr=stderr, inf_type=inf_type,
-                                              pred_dist=None, fname_transformer=None)
+                                              fname_transformer=fname_transformer)
 
-            return {'normal': get_inference_parametric,
-                    'percentile': lambda: get_inference_nonparametric('percentile'),
-                    'pivot': lambda: get_inference_nonparametric('pivot')}[self._bootstrap_type]
+            return {'normal': get_inference_parametric if can_call else lambda: get_inference_parametric(),
+                    'percentile': get_inference_nonparametric('percentile'),
+                    'pivot': get_inference_nonparametric('pivot')}[self._bootstrap_type]
```
vsyrgkanis, line 247: I thought we said we should avoid the dictionary as it seems to be actually creating all the entries in an eager manner and then choosing the correct one.

vsyrgkanis (approved):

econml/bootstrap.py
```diff
@@ -187,14 +187,20 @@ def call(lower=5, upper=95):
...
+            from .inference import EmpiricalInferenceResults, NormalInferenceResults
 
             prefix = name[: - len("_inference")]
+            fname_transformer = None
+
             if prefix in ['const_marginal_effect', 'marginal_effect', 'effect']:
                 inf_type = 'effect'
             elif prefix == 'coef_':
                 inf_type = 'coefficient'
+                if (hasattr(self._instances[0], 'cate_feature_names') and
```
vsyrgkanis, line 200: Currently we have no test that calls coef__inference or summary() but such that the estimator has no cate_feature_names method!

We either need to add such a test and make sure the behavior is as we want: 
1) I see this here being problematic because we would ideally want when I call summary(feat_name=..), even if the method has no cate_feature_names method, that the feat_name list that I give, will appear in the summary table.

Or maybe even better: we make cate_feature_names a mandatory method for any estimator that inherits from LinearModelFinalInference class and the LinearModelFinalInferenceDiscrete class. In this case, summary should perform ok as we will always have cate_feature_names implemented, whenever summary is available.

econml/bootstrap.py
```diff
@@ -187,14 +187,20 @@ def call(lower=5, upper=95):
...
+                        callable(self._instances[0].cate_feature_names)):
+                    def fname_transformer(x):
+                        return self._instances[0].cate_feature_names(x)
             elif prefix == 'intercept_':
                 inf_type = 'intercept'
             else:
```
vsyrgkanis, line 206: We should add a test that checks that the attribute error is raised in this loop and nothing else happens.

econml/bootstrap.py
```diff
@@ -212,21 +220,31 @@ def get_dist(est, arr):
                         return 2 * est - arr
                     else:
                         raise ValueError("Invalid kind, must be either 'percentile' or 'pivot'")
```
vsyrgkanis, line 222: Maybe here too we should add a test that the correct attribute error is raised.

econml/bootstrap.py
```diff
@@ -212,21 +220,31 @@ def get_dist(est, arr):
...
 
-            return {'normal': get_inference_parametric,
-                    'percentile': lambda: get_inference_nonparametric('percentile'),
-                    'pivot': lambda: get_inference_nonparametric('pivot')}[self._bootstrap_type]
+            return {'normal': get_inference_parametric if can_call else lambda: get_inference_parametric(),
+                    'percentile': get_inference_nonparametric('percentile'),
+                    'pivot': get_inference_nonparametric('pivot')}[self._bootstrap_type]
 
         caught = None
         m = None
```
vsyrgkanis, line 250: Currently no test reaches the branch where self._compute_means=True and self._prefer_wrapped=True.

Do we want to maintain these flags? If so then what functionality are they serving and if they are serving some functionality we should add a test that verifies it. Alternatively we can simply remove these flags and remove this branch from here if that's the only place they are used, just for the shake of maintaining a clean code.

econml/cate_estimator.py
```diff
@@ -784,7 +809,30 @@ def summary(self, T, *, alpha=0.1, value=0, decimals=3, feat_name=None):
...
+            print("Coefficient Results: ", e)
+        try:
+            intercept_table = self.intercept__inference(T).summary_frame(
+                alpha=alpha, value=value, decimals=decimals, feat_name=None)
+            intercept_array = intercept_table.values
+            intercept_headers = intercept_table.columns.tolist()
+            intercept_stubs = intercept_table.index.tolist()
+            intercept_title = 'Intercept Results'
+            smry.add_table(intercept_array, intercept_headers, intercept_stubs, intercept_title)
+        except Exception as e:
```
vsyrgkanis, line 831: Currently no Discrete estimator has fit_cate_intercept=False in the tests and this exception is not tested. We should add tests for instance for LinearDRLearner with fit_cate_intercept=False.

econml/inference.py
```diff
@@ -1216,11 +1161,16 @@ def _mixture_ppf(self, alpha, mean, stderr, tol):
         """
         Helper function to get the confidence interval of mixture gaussian distribution
         """
-        done = False
+        # if stderr is zero, ppf will return nans and the loop below would never terminate
+        # so bail out early; note that it might be possible to correct the algorithm for
+        # this scenario, but since scipy's cdf returns nan whenever scale is zero it won't
+        # be clean
+        if (np.any(stderr == 0)):
```
vsyrgkanis, line 1168: This branch is never tested. So we never test population summary with zero stderrs.

vsyrgkanis (approved):

econml/inference.py
```diff
@@ -949,7 +893,8 @@ def pvalue(self, value=0):
             the corresponding singleton dimensions in the output will be collapsed
             (e.g. if both are vectors, then the output of this method will also be a vector)
         """
-        return min((self.pred_dist < value).sum(), (self.pred_dist > value).sum()) / self.pred_dist.shape[0]
+        return np.minimum((self.pred_dist <= value).sum(axis=0),
```
vsyrgkanis, line 896: The pvalue calculation has been fixed for empirical bootstrap when point=0 and std=0, but for NormalInference, if point=0 and std=0, we still return a NaN instead of 1 for p-value.

vsyrgkanis (approved):

vsyrgkanis:
``
est2 = LinearDRLearner(model_regression=RandomForestRegressor(n_estimators=10),
                       model_propensity=RandomForestClassifier(n_estimators=10, min_samples_leaf=50),
                 featurizer=None, min_propensity=1e-3,
                 n_splits=6)
est2.fit(Y, T, X, W, inference=BootstrapInference(n_bootstrap_samples=8))
est2.coef__inference(T=1).summary_frame()
``

Should be returning the inference only for the coefficient associated with T=1. Currently it returns inference for all models. 

Moreover, `est2.summary(T=1)` raises an error. Weird since, I think we do have a test for summary with drlearner and bootstrap.

econml/bootstrap.py
```diff
@@ -212,21 +220,31 @@ def get_dist(est, arr):
...
-            def get_inference_parametric():
+                result = proxy(can_call, prefix,
+                               lambda arr, est: EmpiricalInferenceResults(d_t=d_t, d_y=d_y,
+                                                                          pred=est, pred_dist=get_dist(est, arr),
+                                                                          inf_type=inf_type,
+                                                                          fname_transformer=fname_transformer))
+                # Note that inference results are always methods even if the inference is for a property
+                # (e.g. coef__inference() is a method but coef_ is a property)
+                # Therefore we must insert a lambda if getting inference for a non-callable
+                return result if can_call else lambda: result
```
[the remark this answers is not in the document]
kbattocchi: If you mean, why are these methods instead of properties, it's to match the signature on the CATE estimator class.

econml/bootstrap.py
kbattocchi: Yes, I will still clean this up (and address a few outstanding corner cases), but wanted to make sure the logic was correct before optimizing.

econml/bootstrap.py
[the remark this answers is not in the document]
vsyrgkanis: Oh I see the comment above actually explains it!

econml/bootstrap.py
```diff
@@ -212,21 +220,31 @@ def get_dist(est, arr):
...
+                    stderr = stderr(*args, **kwargs)
                 return NormalInferenceResults(d_t=d_t, d_y=d_y, pred=pred,
                                               pred_stderr=stderr, inf_type=inf_type,
-                                              pred_dist=None, fname_transformer=None)
+                                              fname_transformer=fname_transformer)
 
-            return {'normal': get_inference_parametric,
-                    'percentile': lambda: get_inference_nonparametric('percentile'),
-                    'pivot': lambda: get_inference_nonparametric('pivot')}[self._bootstrap_type]
+            return {'normal': get_inference_parametric if can_call else lambda: get_inference_parametric(),
```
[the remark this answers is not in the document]
kbattocchi: Addressed by refactoring

econml/bootstrap.py
kbattocchi: Addressed by refactoring

econml/bootstrap.py
kbattocchi: Refactored so that it's impossible to hit this line

econml/bootstrap.py
kbattocchi: Removed that code since it added complexity for no benefit.

econml/cate_estimator.py
kbattocchi: Added such a test.

econml/inference.py
kbattocchi: Fixed edge cases and added a test.

econml/inference.py
kbattocchi: Added a test.

PR #302, opened 2020-11-06
Fix Pandas compatibility

moprescu:
* Added array checks to ensure compatibility with Pandas
* Fixed failures when passing in column vectors for Y and T


econml/cate_estimator.py
```diff
@@ -5,8 +5,10 @@
 
 import abc
 import numpy as np
+import pandas as pd
 from functools import wraps
 from copy import deepcopy
+from sklearn.utils import check_array
```
kbattocchi, line 11: The new imports here shouldn't be needed if there are no other changes to this file, right?
```suggestion
from functools import wraps
from copy import deepcopy
```

econml/dml.py
```diff
@@ -696,6 +696,7 @@ def fit(self, Y, T, X=None, W=None, sample_weight=None, sample_var=None, inferen
         if sample_var is not None and inference is not None:
             warn("This estimator does not yet support sample variances and inference does not take "
                  "sample variances into account. This feature will be supported in a future release.")
+        Y, T, X, W, sample_weight, sample_var = check_input_arrays(Y, T, X, W, sample_weight, sample_var)
```
kbattocchi, line 699: Why only add the check for SparseLinearDML, but not the other classes in this file?

econml/drlearner.py
```diff
@@ -928,6 +928,7 @@ def fit(self, Y, T, X=None, W=None, *, sample_weight=None, sample_var=None, infe
         if sample_weight is not None and inference is not None:
             warn("This estimator does not yet support sample variances and inference does not take "
                  "sample variances into account. This feature will be supported in a future release.")
+        Y, T, X, W, sample_weight, sample_var = check_input_arrays(Y, T, X, W, sample_weight, sample_var)
```
kbattocchi, line 931: Likewise, why only the SparseLinear version?  If it's because it's needed for the check on the next line, then maybe only run check_input_arrays on X and T and add a comment explaining.

econml/metalearners.py
```diff
@@ -262,6 +262,8 @@ def fit(self, Y, T, X, *, inference=None):
         """
         # Check inputs
         Y, T, X, _ = check_inputs(Y, T, X, multi_output_T=False)
```
kbattocchi, line 264: Is check_inputs equivalent to check_input_arrays?  Would it be beneficial to switch things to the latter everywhere for consistency?

econml/metalearners.py
```diff
@@ -262,6 +262,8 @@ def fit(self, Y, T, X, *, inference=None):
...
+        if Y.ndim == 2 and Y.shape[1] == 1:
+            Y = Y.flatten()
```
kbattocchi, line 266: Why is this necessary?  Please add a comment explaining, and a test that covers this scenario.

econml/utilities.py
```diff
@@ -510,6 +511,12 @@ def check_inputs(Y, T, X, W=None, multi_output_T=True, multi_output_Y=True):
     return Y, T, X, W
 
 
+def check_input_arrays(*args):
+    # Check arrays for all sequence-like objects
```
kbattocchi, line 515: Add proper docstring

econml/utilities.py
```diff
@@ -4,6 +4,7 @@
 """Utility methods."""
 
 import numpy as np
+import pandas as pd
```
kbattocchi, line 7: Doesn't appear necessary unless there are other changes.
```suggestion
```

kbattocchi (changes requested):
Mostly looks fine, but definitely needs corresponding tests.

econml/dml.py
moprescu: Because the sparse methods call other methods (e.g. `check_high_dimensional`) before calling the super's fit. All other DML and DR estimators call the parent's fit (`_ortho_learner` in this case).

econml/drlearner.py
moprescu: Same as above.

econml/utilities.py
moprescu: Done!

kbattocchi (approved):

econml/metalearners.py
moprescu: The `sklearn` models flatten the output and the difference between an `(n, 1)` and an `(n, )`  is an `(n, n)` array. Since there are several models where we compute differences this way and it's difficult to keep track of the shapes, it's more straightforward to just flatten the input `Y`, I'll make sure to add a comment that reflects this.

Issue #303, opened 2020-11-06
Logic in summary_frame


vsyrgkanis:
The logic here:
https://github.com/microsoft/EconML/blob/a1d6aa81b512e2312a08fecac849db8cbbea03b9/econml/inference.py#L640

might be better in terms of user experience to change slightly. In particular, cate_feature_names can handle None feat_names as input and do something fancy even when feat_names is None (e.g. returning the metadata info or the featurizer's output (which can also accept None). 

So if fname_transformer is not None, we should always call it as fname_transformer(feat_name), even if feat_name is None. 

Moreover, if fname_transformer is None, but feat_name is not None, then we should use the raw feat_names as feature names without transforming them.

vsyrgkanis:
Fixed by #305

shrabasteebanerjee:
Wanted to point out a small thing related to this in case someone runs into a similar issue: following the DRLearner notebook, I was trying to pass "feature_names" to summary_frame() which it did not recognise. Then I explicitly imported InferenceResults from econml.inference, and discovered the feat_name input!

PR #305, opened 2020-11-07
Changed logic in summary_frame for feature names

vsyrgkanis:
Changed logic of feature names in summary_frame and added corresponding tests. Fixed a small bug in LinearModelFinalCateEstimatorDiscreteMixin where we werenot checking in intercept_ whether an intercept was fitted and raising an Attribute error if so. This lead to a weird error being raised because we were trying to reshape an integer.

Addressing #303

### DIFF
--- a/econml/inference.py
+++ b/econml/inference.py
@@ -637,8 +637,10 @@ class InferenceResults(metaclass=abc.ABCMeta):
         if self.d_y == 1:
             res.index = res.index.droplevel(1)
         if self.inf_type == 'coefficient':
+            if self.fname_transformer is not None:
+                feat_name = self.fname_transformer(feat_name)
             if feat_name is not None:
-                ind = self.fname_transformer(feat_name)
+                ind = feat_name
             else:
                 ct = res.shape[0] // self.d_y
                 ind = ['X' + str(i) for i in range(ct)]
--- a/econml/tests/test_inference.py
+++ b/econml/tests/test_inference.py
@@ -5,6 +5,7 @@ import numpy as np
 import unittest
 from sklearn.base import clone
 from sklearn.preprocessing import PolynomialFeatures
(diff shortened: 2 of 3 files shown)


econml/tests/test_inference.py
```diff
@@ -93,15 +261,16 @@ def test_degenerate_cases(self):
             pop.print()  # verify that we can access all attributes even in degenerate case
 
     def test_can_summarize(self):
-        LinearDMLCateEstimator().fit(
+        LinearDMLCateEstimator(model_t=LinearRegression(), model_y=LinearRegression()).fit(
```
kbattocchi, line 264: What was wrong with just using the defaults here?

econml/inference.py
```diff
@@ -637,8 +637,10 @@ def summary_frame(self, alpha=0.1, value=0, decimals=3, feat_name=None):
         if self.d_y == 1:
             res.index = res.index.droplevel(1)
         if self.inf_type == 'coefficient':
+            if self.fname_transformer is not None:
```
kbattocchi, line 640: I thought we had discussed using an identity lambda as the default for `fname_transformer` so that we'd never need to check for `None` (and I made that change to the default in several places in the last bootstrap PR, though perhaps I didn't cover them all).

kbattocchi (approved):

econml/tests/test_inference.py
vsyrgkanis: Just faster ;)

econml/inference.py
vsyrgkanis: Just in case an “internal” user passes fname_transformer=None explicitly.

vsyrgkanis: I should say: the main change here is not checking that transformer is not none. Rather it is that we are always applying the fname transformer even if feat_name is none. Previously we were applying the transformer only when feat_name was not None.

PR #306, opened 2020-11-07
Enabled feature_importances_ for our ForestDML and ForestDRLearner estimators

### LINKED ISSUE #297: Shap interpretability with SubsampledHonestForest
The subsampled honest forest currently is not exactly compatible with Shap, due to some internal re-workings, even though it inherits from sciki-learn's random forest regressor. Thus getting Shap values for the learned model is not trivial. We should enable shap interpretability of the subsampled  honest forest, so that we can interpret ForestDML, ForestDRLearner and ForestDRIV.

vsyrgkanis:
This required changing the subsampled honest forest code a bit so that it does not alter the arrays of the tree structures of sklearn but rather stores two additional arrays required for prediction. This does add around 1.5 times the original running time, so makes it slightly slower due to the extra memory allocation.

However this enables correct feature_importance calculation and also in the future correct SHAP calculation (fixes #297), as now the tree entries are consistent with a tree in a randomforestregressor and so shap logic can be applied if we recast the subsampled honest forest as a randomforestregressor (additivity of shap will still be violated since the prediction of the subsample honest forest is not just the aggregation of the predictions across the trees but more complex weighted average). But we can still call shap and still get meaningful shap numbers. One discrepancy is that shap is explaining a different value that what effect returns, since it explains the value that corresponds to the average of the predictions of each honest tree regressor. however, the prediction of an honest forest is not the average of the tree predictions. For a full solution to this small discrepancy, one would need a full re-working of Shap's tree explainer and the tree explainer algorithm to account for such alternative aggregations of tree predictors.

This enables the following two uses:

```python
from econml.dml import ForestDMLCateEstimator
import shap
import sklearn.ensemble
import copy
est3 = ForestDMLCateEstimator(model_y=RandomForestRegressor(),
                              model_t=RandomForestRegressor(),
                              n_estimators=1000,
                              subsample_fr=.8,
                              min_samples_leaf=10,
                              min_impurity_decrease=0.001,
                              verbose=0, min_weight_fraction_leaf=.01)
est3.fit(Y, T, X, W)
print(est3.feature_importances_)
model = copy.deepcopy(est3.model_cate)
model.__class__ = sklearn.ensemble.RandomForestRegressor
explainer = shap.Explainer(model, X)
shap_values = explainer(X)
shap.plots.beeswarm(shap_values, X)
```

```python
from econml.drlearner import ForestDRLearner
import shap
import sklearn.ensemble
import copy
est3 = ForestDRLearner(model_regression=RandomForestRegressor(),
                       model_propensity=RandomForestClassifier(min_samples_leaf=10),
                       min_propensity=1e-3,
                       n_estimators=1000,
                       subsample_fr=.8,
                       min_samples_leaf=10,
                       min_impurity_decrease=0.001,
                       verbose=0, min_weight_fraction_leaf=.01)
est3.fit(Y, T, X, W)
for t in np.unique(T):
    if t > 0:
        print(est3.feature_importances_(T=t))
        model = copy.deepcopy(est3.model_cate(T=t))
        model.__class__ = sklearn.ensemble.RandomForestRegressor
        explainer = shap.Explainer(model, X)
        shap_values = explainer(X)
        shap.plots.beeswarm(shap_values, X)
```

Side benefits: 
1) 6x speed up of SubsampledHonestForest by pre-transforming dense but sparsely represented matrices to dense representation before a loop with many slicing operations on them.


econml/dml.py
```diff
@@ -866,7 +867,7 @@ def __init__(self,
                          random_state=random_state)
 
 
-class ForestDMLCateEstimator(NonParamDMLCateEstimator):
+class ForestDMLCateEstimator(NonParamDMLCateEstimator, ForestModelFinalCateEstimatorMixin):
```
kbattocchi, line 870: Generally speaking I believe the mixin should come first in the set of inherited classes to avoid having method lookup happen in an unexpected order.  It may not matter at the moment in this case because the mixin's logic never calls into anythingn from `super` but I think it should be switched around in case changes are made in the future.

econml/drlearner.py
```diff
@@ -949,7 +950,7 @@ def fitted_models_final(self):
         return super().model_final.models_cate
 
 
-class ForestDRLearner(DRLearner):
+class ForestDRLearner(DRLearner, ForestModelFinalCateEstimatorDiscreteMixin):
```
kbattocchi, line 953: Same comment applies here.

econml/sklearn_extensions/ensemble.py
```diff
@@ -523,21 +532,27 @@ def fit(self, X, y, sample_weight=None, sample_var=None):
...
-                             **_joblib_parallel_args(prefer='threads'))(
+            res = Parallel(n_jobs=self.n_jobs, verbose=self.verbose,
+                           **_joblib_parallel_args(prefer='threads'))(
                 delayed(_parallel_add_trees)(
                     t, self, X, y, sample_weight, s_inds[i], i, len(trees),
                     verbose=self.verbose)
                 for i, t in enumerate(trees))
+            trees = [t[0] for t in res]
+            numerators = [t[1] for t in res]
+            denominators = [t[2] for t in res]
```
kbattocchi, line 543: I think this is the same as:
```suggestion
            trees, numerators, denominators = zip(*res)
```
which is cleaner looking to me.

kbattocchi (changes requested):

econml/dml.py
vsyrgkanis: Done

econml/drlearner.py
vsyrgkanis: Done

econml/sklearn_extensions/ensemble.py
```diff
@@ -756,3 +761,38 @@ def predict_interval(self, X, alpha=.1, normal=True):
...
+        def unnormalized_importances(tree):
+            return tree.tree_.compute_feature_importances(normalize=False)
+
+        all_importances = Parallel(n_jobs=self.n_jobs,
+                                   **_joblib_parallel_args(prefer='threads'))(
+            delayed(unnormalized_importances)(tree)
+            for tree in self.estimators_ if tree.tree_.node_count > 1)
+
+        if not all_importances:
+            return np.zeros(self.n_features_, dtype=np.float64)
```
kbattocchi, line 793: This isn't covered by any tests.

kbattocchi (approved):

econml/sklearn_extensions/ensemble.py
vsyrgkanis: I have added tests that call feature importances. Not sure why it isnt covered. Maybe you looked at the build before the last commit. This was added in the last commit

vsyrgkanis: Oh you mean the if not all_importances??

vsyrgkanis: I implemented the TODO which now makes the feature_importances_ more correct. So now this piece of code is actually erased. But it is replaced with more code in the parallel_add_trees function.

econml/sklearn_extensions/ensemble.py
```diff
@@ -65,14 +59,36 @@ def _parallel_add_trees(tree, forest, X, y, sample_weight, s_inds, tree_idx, n_t
     # Calculate the total weight of estimation samples on each tree node:
     # \sum_i sample_weight[i] * 1{i \\in node}
     weight_est = scipy.sparse.csr_matrix(
-        sample_weight_est.reshape(1, -1)).dot(path_est)
+        sample_weight_est.reshape(1, -1)).dot(path_est).todense()
```
kbattocchi, line 62: So `path_est` is sparse, but `sample_weight_est` is initially dense and so we first make it sparse to compute the dot product, and then densify the result?  It looks like there's an existing scipy function [safe_sparse_dot](https://scikit-learn.org/stable/modules/generated/sklearn.utils.extmath.safe_sparse_dot.html) that does a perhaps more efficient version of that already, and [the implementation](https://github.com/scikit-learn/scikit-learn/blob/0fb307bf3/sklearn/utils/extmath.py#L119) just boils down to `a @ b` for the not-higher-than-2-dimensional case, which I assume we're in here.

econml/sklearn_extensions/ensemble.py
```diff
@@ -65,14 +59,36 @@ def _parallel_add_trees(tree, forest, X, y, sample_weight, s_inds, tree_idx, n_t
...
+    # Calculate the criterion on each node based on the estimation sample and for each output dimension,
+    # summing the impurity across dimensions.
+    # First we calculate the difference of observed label y of each node and predicted value for each
+    # node that the sample falls in: y[i] - value_est[node]
+    impurity_est = np.zeros((1, path_est.shape[1]))
+    for i in range(tree.n_outputs_):
+        diff = path_est.multiply(y_est[:, [i]]) - path_est.multiply(value_est[[i], :])
+        if tree.criterion == 'mse':
+            # If criterion is mse then calculate weighted sum of squared differences for each node
+            impurity_est_i = scipy.sparse.csr_matrix(sample_weight_est.reshape(1, -1)).dot(diff.power(2))
```
kbattocchi, line 83: Looks like the `scipy.sparse.csr_matrix(sample_weight_est.reshape(1, -1))` calculation happens each time through the loop, and also happens in the computation of `weight_est`; just compute once outside the loop and reuse.

econml/sklearn_extensions/ensemble.py
```diff
@@ -65,14 +59,36 @@ def _parallel_add_trees(tree, forest, X, y, sample_weight, s_inds, tree_idx, n_t
...
+        sample_weight_est.reshape(1, -1)).dot(path_est).todense()
     # Calculate the total number of estimation samples on each tree node:
     # |node| = \sum_{i} 1{i \\in node}
-    count_est = np.sum(path_est, axis=0)
+    count_est = path_est.sum(axis=0)
     # Calculate the weighted sum of responses on the estimation sample on each node:
     # \sum_{i} sample_weight[i] 1{i \\in node} Y_i
-    value_est = scipy.sparse.csr_matrix(
-        (sample_weight_est.reshape(-1, 1) * y_est).T).dot(path_est)
+    num_est = scipy.sparse.csr_matrix(
```
kbattocchi, line 68: Same comment applies here.

kbattocchi (approved):
Mostly looks fine, though I'm not really familiar with the tree internals to vouch for correctness.  Please consider addressing the comments I left before merging.

econml/sklearn_extensions/ensemble.py
vsyrgkanis: Weirdly when I tried storing this matrix and re-using it it was slower. Maybe there is some weirdness with memory persistence? Don't know.

econml/sklearn_extensions/ensemble.py
vsyrgkanis: Cool! Yeah this is much cleaner.

econml/sklearn_extensions/ensemble.py
vsyrgkanis: Sweet!

PR #307, opened 2020-11-09
Vasilis/autoinference

### LINKED ISSUE #205: inference='auto'
We should add a "inference='auto'"  or "inference=True" option to all estimators, which would use some default option for inference. For instance, this could be bootstrap or None (if we want to disallow inference in some estimators that we think bootstrap will be deceiving) or the more tailored methods.

This will allow the user to not require to know which one is the most appropriate method. Also it would allow one to not have to change the fit call whenever they change the estimator class.

vsyrgkanis:
1) Made default='auto' for many estimators, so that default inference is enabled (fixes #205).
2) Enabled LinearModelFinalInference at the DMLCateEstimator level, so that if the user uses any model_final that has predict_interval and prediction_stderr they will get inference. This enables use of RLM as final stage with conf intervals.
3) Added an RLM scikit-learn wrapper.

This enables the following usecase:
```python 
from econml.sklearn_extensions.linear_model import StatsModelsRLM
est = DMLCateEstimator(model_y=LassoCV(),
                        model_t=LassoCV(),
                        model_final=StatsModelsRLM(t=1, maxiter=10000, tol=1e-12, fit_intercept=False),
                        featurizer=PolynomialFeatures(degree=1, include_bias=False),
                        random_state=123)
est.fit(Y, T, X, W)
te_pred = est.effect(X_test)
lb, ub= est.effect_interval(X_test)
est.summary()
```


econml/cate_estimator.py
```diff
@@ -662,6 +668,7 @@ def _get_inference_options(self):
         # add debiasedlasso to parent's options
         options = super()._get_inference_options()
         options.update(debiasedlasso=LinearModelFinalInference)
+        options.update(auto=LinearModelFinalInference)
```
kbattocchi, line 671: Should be unnecessary given the change to the mixin, no?

econml/cate_estimator.py
```diff
@@ -876,6 +890,7 @@ def _get_inference_options(self):
         # add statsmodels to parent's options
         options = super()._get_inference_options()
         options.update(debiasedlasso=LinearModelFinalInferenceDiscrete)
+        options.update(auto=LinearModelFinalInferenceDiscrete)
```
kbattocchi, line 893: Also appears to be redundant.

kbattocchi (approved):
Love the idea of enabling inference by default when it's efficient.  I don't see which changes enable arbitrary linear models in this set of commits - what am I missing?

econml/cate_estimator.py
vsyrgkanis: Just in case we ever change the default in the base class, I thought it'd be good that the default is hard coded and controlled here.

econml/cate_estimator.py
vsyrgkanis: Same on this.

econml/drlearner.py
```diff
@@ -1,1188 +1,1188 @@
-# Copyright (c) Microsoft Corporation. All rights reserved.
```
kbattocchi, line 1: The entire file is showing up as modified on GitHub; when I do a diff locally and set it to ignore whitespace I see the expected changes.  Is it possible that the line endings have changed here?  Normally git automatically handles this, but please take a look and fix on your end.

kbattocchi (changes requested):
Mostly fine, except for a weird whitespace issue.

econml/cate_estimator.py
kbattocchi: That's fine, but if the logic is going to be set explicitly in each subclass does having it in the base serve any purpose?

vsyrgkanis: Yeap: dml is not of any of these subclasses and this sets the “auto” for that class for instance

econml/drlearner.py
vsyrgkanis: Weird! What can I do?

kbattocchi: Looks like you could try `git add --renormalize` on the file, assuming your version of git is relatively recent.

vsyrgkanis: fixed!

kbattocchi (approved):
LGTM

econml/dml.py
```diff
@@ -447,6 +447,34 @@ def __init__(self,
                          n_splits=n_splits,
                          random_state=random_state)
 
+    # override only so that we can update the docstring to indicate support for `StatsModelsInference`
+    def fit(self, Y, T, X=None, W=None, sample_weight=None, sample_var=None, inference='auto'):
```
kbattocchi, line 451: This override is missing the `groups` argument.

econml/dml.py
```diff
@@ -447,6 +447,34 @@ def __init__(self,
...
+        inference: string, :class:`.Inference` instance, or None
+            Method for performing inference.  This estimator supports 'bootstrap'
+            (or an instance of :class:`.BootstrapInference`) and 'auto'
+            (or an instance of :class:`.LinearModelFinalInference`)
+
+        Returns
+        -------
+        self
+        """
+        return super().fit(Y, T, X=X, W=W, sample_weight=sample_weight, sample_var=sample_var, inference=inference)
```
kbattocchi, line 476: And `groups` should be passed through here.

econml/dml.py
```diff
@@ -529,7 +557,7 @@ def __init__(self,
                          random_state=random_state)
 
     # override only so that we can update the docstring to indicate support for `StatsModelsInference`
-    def fit(self, Y, T, X=None, W=None, *, sample_weight=None, sample_var=None, groups=None, inference=None):
+    def fit(self, Y, T, X=None, W=None, sample_weight=None, sample_var=None, groups=None, inference='auto'):
```
kbattocchi, line 560: Continue to require that the final arguments are passed only by name.
```suggestion
    def fit(self, Y, T, X=None, W=None, *, sample_weight=None, sample_var=None, groups=None, inference='auto'):
```

econml/dml.py
```diff
@@ -675,7 +703,7 @@ def __init__(self,
                          n_splits=n_splits,
                          random_state=random_state)
 
-    def fit(self, Y, T, X=None, W=None, *, sample_weight=None, sample_var=None, groups=None, inference=None):
+    def fit(self, Y, T, X=None, W=None, sample_weight=None, sample_var=None, groups=None, inference='auto'):
```
kbattocchi, line 706:
```suggestion
    def fit(self, Y, T, X=None, W=None, *, sample_weight=None, sample_var=None, groups=None, inference='auto'):
```

econml/dml.py
```diff
@@ -1076,7 +1104,7 @@ def __init__(self,
                          categories=categories,
                          n_splits=n_crossfit_splits, random_state=random_state)
 
-    def fit(self, Y, T, X=None, W=None, *, sample_weight=None, sample_var=None, groups=None, inference=None):
+    def fit(self, Y, T, X=None, W=None, sample_weight=None, sample_var=None, groups=None, inference='auto'):
```
kbattocchi, line 1107:
```suggestion
    def fit(self, Y, T, X=None, W=None, *, sample_weight=None, sample_var=None, groups=None, inference='auto'):
```

kbattocchi (changes requested):

econml/dml.py
vsyrgkanis: done

econml/dml.py
vsyrgkanis: weird!

vsyrgkanis: oooh yeah I got it! It's because now I added this definition of fit to change to 'auto'!

kbattocchi (approved):
Looks good, thanks!

PR #309, opened 2020-11-11
Deprecate wordy class names

kbattocchi:


vsyrgkanis (changes requested):
All looks great!

One thing is that we definitely need to change our notebooks and the README to use the new names. Similarly also the docstrings in the documentation and within code.

vsyrgkanis:
Also maybe as you are changing the names in the notebooks and in the readme, it would be great if you could also change the inference='statsmodels' etc keyword to omit the keyword, when the "auto" option is good enough. This would make the code much shorter and will use the new proposed syntax with the auto inference.

vsyrgkanis (approved):
Looks good! I guess if any test or notebook is left to use the old names we’ll also get it flagged when we deprecate in the next release.

vsyrgkanis:
Can we also change the examples in the readme?

doc/spec/inference.rst
```diff
@@ -153,7 +154,7 @@ estimating the uncertainty of the honest causal forests underlying the OrthoFore
                                          min_leaf_size=3,
                                          model_T=WeightedLasso(alpha=0.01),
                                          model_Y=WeightedLasso(alpha=0.01))
-    est.fit(y, t, X, W, inference='blb')
+    est.fit(y, t, X, W)
```
kbattocchi, line 157: @vsyrgkanis This change is causing a doctest failure because ContinuousTreatmentOrthoForest doesn't use 'auto' as a default (as opposed to DiscreteTreatmentOrthoForest, which does).  Is there a reason that it doesn't, or is that just an oversight?

vsyrgkanis: Oversight! Please change

vasilismsr:
One small thing if you have time: with the replacing of old with new names, the tab alignment is broken in our notebooks and they look slightly ugly.

doc/spec/inference.rst
kbattocchi: I've addressed this, although the fix has highlighted how weird our `wrap_fit` logic is; I think we should probably completely revamp it as part of addressing the second-stage-only bootstrap changes.

PR #314, opened 2020-11-13
Deprecate positional arguments to fit

kbattocchi:
This change deprecates passing arguments to fit (other than Y and T) by position, rather than by keyword.  So, for example `est.fit(Y, T, X)` will warn, while `est.fit(Y, T, X=None)` will not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,13 +140,13 @@ To install from source, see [For Developers](#for-developers) section below.
 
   est = LinearDML(model_y=LassoCV(), model_t=LassoCV())
   ### Estimate with OLS confidence intervals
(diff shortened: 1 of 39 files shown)


kbattocchi:
@heimengqi @vsyrgkanis I've added the attribute pass-through for SelectiveRegularization to this PR - please take a look.

vsyrgkanis (changes requested):
Since we are deprecating non-positional shouldn't we also be changing again all our notebooks and examples to use the new notation for demonstration purposes?

vsyrgkanis (approved):
LGTM

PR #316, opened 2020-11-13
Vasilis/orf speed

vsyrgkanis:
* Speed up prediction time of ORF. Replaced in-class pointwise_effect function with static method that takes the subset of arrays for each point. Also using processes instead of threading. Leads to at around a 10x speed-up.
* Enabled global residualization in continuous treatment orthoforest. This then replicates exactly the functionality of grf and brings parity with grf. Added class causal_forest.CausalForest, that essentially is a continuous treatment ortho forest with global residualization. Global residualization is computationally much faster, but might lose statistical power especially with W not none, as it fits a global residualizer and not a local residualizer for each target test point. This now allows us to do sth like ForestDML but with many treatments. Moreover, unlike ForestDML this uses exactly the splitting criterion of the causal forest.
* Enabled _inference in all ortho forests by adding inference methods in the BLBInference class in orthoforest.

# Speed up example
For instance, here is a code example with 1000 trees and 1000 prediction points:
```python
np.random.seed(123)
n = 2000
p = 10
X = np.random.normal(size=(n, p))
def true_propensity(x): return .4 + .2 * (x[:, 0] > 0)
def true_effect(x): return (x[:, 0] * (x[:, 0] > 0))
def true_conf(x): return x[:, 1] + np.clip(x[:, 2], - np.inf, 0)

T = np.random.binomial(1, true_propensity(X))
Y = true_effect(X) * T + true_conf(X) + np.random.normal(size=(n,))

X_test = np.zeros((1000, p))
X_test[:, 0] = np.linspace(-2, 2, 1000)


est3 = DiscreteTreatmentOrthoForest(model_Y=Lasso(alpha=0.01),
                                    propensity_model=LogisticRegression(C=1),
                                    model_Y_final=WeightedLassoCV(cv=3),
                                    propensity_model_final=LogisticRegressionCV(cv=3),
                                    n_trees=1000, min_leaf_size=10)
est3.fit(Y, T, X)
lb, ub = est3.effect_interval(X_test)
pred3 = est3.effect(X_test)
```

Here is the running time of the new code:
image

and here is the running time of the current master code:
image

Moreover, now there is a big benefit for parallelism as with 8 cores we get around 5x speed-up, compared to non-parallel.
image


econml/sklearn_extensions/model_selection.py
```diff
@@ -190,3 +201,183 @@ def split(self, X, y, sample_weight=None):
             Weights associated with the training data.
         """
         return _split_weighted_sample(self, X, y, sample_weight, is_stratified=True)
+
+
+def cross_val_predict(estimator, X, y=None, *, groups=None, cv=None,
+                      n_jobs=None, verbose=0, fit_params=None,
```
kbattocchi, line 207: This looks extremely similar to sklearn's version - what's different?  Also, be sure to add a comment giving attribution to sklearn so that it's obvious we're not claiming this is entirely our work.

econml/utilities.py
```diff
@@ -1577,3 +1577,40 @@ def transpose_dictionary(d):
         for key2, val in value.items():
             output[key2][key1] = val
     return output
+
+
+class RegWrapper:
+    """
```
kbattocchi, line 1583: I think `RegressionWrapper` would be more appropriate for a public component.

kbattocchi: Should this also pass attributes through to the underlying classifier?

kbattocchi (changes requested):

econml/ortho_forest.py
```diff
@@ -461,6 +475,25 @@ class ContinuousTreatmentOrthoForest(BaseOrthoForest):
...
+        Whether to perform a prior residualization of Y and T using the model_Y_final and model_T_final
+        estimators, or whether to perform locally weighted residualization at each target point.
+        Global residualization is computationally less intensive, but could lose some statistical
+        power, especially when W is not None.
+
+    global_res_cv : int, cross-validation generator or an iterable, optional (default=2)
+        The specification of the cv splitter to be used for cross-fitting, when constructing
+        the global residuals of Y and T.
+
+    discrete_treatment : bool, optional (default=False)
```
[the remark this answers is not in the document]
vsyrgkanis: I know. I was worried about this too, but thought it's fine to do it this way. Alternatively we can move all the discretization of treatment stuff in the causal forest fit function and not allow for discrete treatment here. Though this limits functionality: what if I want to use local residualization and the dml moments (not the DR moments that are used in DiscreteTreatmentORF)? This is done here with discrete_treamtent=True. So I thought we could add it.

I was alternatively compensating renaming to:
DMLOrthoForest
DROrthoForest

instead of ContinousTreatmentOrthoForest and DiscreteTreatmentOrthoForest and deprecating the old names in the next release. 

What do you think of this??

econml/sklearn_extensions/model_selection.py
vsyrgkanis: It is a copy-paste, with the only change in passing the "safe" keyword arg, that allows for safe=False, cloning of the models for each fold. Otherwise it wont be able to accept models that are not sklearn models.

econml/utilities.py
vsyrgkanis: I was having a problem with cloning if I enable that and for now we don't need it, it's just for internal use. If we use it further we can enable this in a future PR.

vsyrgkanis: I changed the name

econml/ortho_forest.py
```diff
@@ -105,6 +108,17 @@ def _cross_fit(model_instance, X, y, split_indices, sample_weight=None, predict_
     return np.concatenate((pred_1, pred_2))[sorted_split_indices]
 
 
+def _group_predict(X, n_groups, predict_func):
```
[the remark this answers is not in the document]
vsyrgkanis: Done

econml/ortho_forest.py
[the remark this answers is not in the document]
kbattocchi: I thought we were going to have just a single OrthoForest with the discrete_treatment flag, and then we could add deprecated aliases for both Continuous and Discrete that inherit from it and set the flag appropriately; is there a reason to keep both classes separate?

econml/utilities.py
kbattocchi: If it's just for internal use, put an underscore at the beginning so it doesn't end up in the public docs.

kbattocchi (approved):

econml/ortho_forest.py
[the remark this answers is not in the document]
vsyrgkanis: I'm changing them to DML and DR orthoforest and decprecating the old ones

econml/ortho_forest.py
```diff
@@ -252,96 +320,58 @@ def const_marginal_effect(self, X):
...
-
-        Parameters
-        ----------
-        X_single : array-like, shape (d_x, )
-            Feature vector that captures heterogeneity for one sample.
-
-        stderr : boolean (default=False)
-            Whether to calculate the covariance matrix via bootstrap-of-little-bags.
-        """
+        def pw_effect_inputs(mask_w1, mask_w2, w_nonzero, split_inds, slice_weights_list):
```
moprescu, line 331: This can be a private helper function outside of `_predict` since it's not using anything from `_predict` so it doesn't have to be redefined every time we call `_predict`. Alternatively, this can be grouped with `_nonzero_weights` since `_nonzero_weights` is only used inside this function.

econml/ortho_forest.py
```diff
@@ -548,28 +630,67 @@ def fit(self, Y, T, X, W=None, inference='auto'):
...
-            return effects.flatten()
-        return effects
+        effects = super().const_marginal_effect(X=X)
+        return effects.reshape((-1,) + self._d_y + self._d_t)
     const_marginal_effect.__doc__ = BaseOrthoForest.const_marginal_effect.__doc__
 
     @staticmethod
-    def nuisance_estimator_generator(model_T, model_Y, random_state=None, second_stage=True):
+    def nuisance_estimator_generator(model_T, model_Y, random_state=None, second_stage=True,
+                                     global_residualization=False, discrete_treatment=False):
```
moprescu, line 666: `global_residualization` is only used when calling `_get_conforming_residuals`. Instead of having this parameter in the `nuisance` or `parameter` estimators, you can instead modify `get_conforming_residuals`:

```
@staticmethod
def _get_conforming_residuals(Y, T, nuisance_estimates, global_residualization):
    if nuisance_estimates==0:
        return reshape_Y_T(Y, T)
    # returns shape-conforming residuals
    Y_hat, T_hat = reshape_Y_T(*nuisance_estimates)
    Y, T = reshape_Y_T(Y, T)
    Y_res, T_res = Y - Y_hat, T - T_hat
    return Y_res, T_res
```

econml/ortho_forest.py
```diff
@@ -646,32 +771,37 @@ def parameter_estimator_func(Y, T, X,
...
-        # Return moments and gradients
-        # Compute residuals
-        Y_res, T_res = ContinuousTreatmentOrthoForest._get_conforming_residuals(Y, T, nuisance_estimates)
-        # Compute moments
-        # Moments shape is (n, d_T)
-        moments = (Y_res - np.matmul(T_res, parameter_estimate)).reshape(-1, 1) * T_res
-        # Compute moment gradients
-        mean_gradient = - np.matmul(T_res.T, T_res) / T_res.shape[0]
-        return moments, mean_gradient
+    def moment_and_mean_gradient_estimator_func_gen(global_residualization=False):
```
moprescu, line 774: This would also go away if you get rid of `global_residualization`

econml/ortho_forest.py
```diff
@@ -1065,25 +1226,53 @@ def effect_interval(self, X=None, *, T0=0, T1=1, alpha=0.1):
...
         upper = 1 - alpha / 2
-        # Calculate the effects
-        eff = np.asarray([np.dot(params_and_cov[i][0], dT[i]) for i in range(X.shape[0])])
-        # Calculate the standard deviations for the effects
-        scales = [np.sqrt(dT[i] @ params_and_cov[i][1] @ dT[i]) for i in range(X.shape[0])]
         effect_lower = eff + np.apply_along_axis(lambda s: norm.ppf(lower, scale=s), 0, scales)
         effect_upper = eff + np.apply_along_axis(lambda s: norm.ppf(upper, scale=s), 0, scales)
         return effect_lower, effect_upper
 
+    def effect_inference(self, X=None, *, T0=0, T1=1, alpha=0.1):
```
moprescu, line 1236: This has the same docstring as `effect_interval`

econml/sklearn_extensions/model_selection.py
```diff
@@ -190,3 +201,186 @@ def split(self, X, y, sample_weight=None):
             Weights associated with the training data.
         """
         return _split_weighted_sample(self, X, y, sample_weight, is_stratified=True)
+
+
+def cross_val_predict(estimator, X, y=None, *, groups=None, cv=None,
```
moprescu, line 206: I would make this a private function and remove most of the docstring and instead add a reference to the sklearn method. This signals to the user that this method might change in the future (if, say, the sklearn one changes) and they shouldn't use it as is.

econml/tests/test_orf.py
```diff
@@ -20,9 +21,9 @@ class TestOrthoForest(unittest.TestCase):
     def setUpClass(cls):
         np.random.seed(123)
         # DGP constants
-        cls.n = 1000
-        cls.d_w = 30
-        cls.support_size = 5
+        cls.n = 5000
```
moprescu, line 24: Why change the dimensions?

moprescu (approved):

econml/ortho_forest.py
vsyrgkanis: good point! I just merged with nonzero weights

econml/ortho_forest.py
vsyrgkanis: I had tried to do this but failed to do in a simple manner. The problem is that _get_conforming_residuals is a static function and cannot access the self.globa_residualization. So one would need to write a generator in a similar fashion that generates the right stattic function. But then it's weird how to merge this with init etc. I found the above cleaner, as we already have generators for these other methods.

econml/ortho_forest.py
vsyrgkanis: same reasoning as above

econml/ortho_forest.py
vsyrgkanis: Good point!!

econml/sklearn_extensions/model_selection.py
vsyrgkanis: Good point!

econml/tests/test_orf.py
vsyrgkanis: Just to make the tests more stable. I think with the previous nsamples and d_w some tests were flaky

econml/ortho_forest.py
moprescu: Sorry, in the example above, global_residualization is not necessary since it just checks whether the nuisances are `0`. It's a sort of contract, when residualization is global, the nuisances are 0 and that can be used by `_get_conforming_residuals`

vsyrgkanis: what if nuisance estimates just happen to be 0 (weird corner case), but weird to use a numeric value to have a special meaning.

vsyrgkanis: We could change the nuisance estimator to return the string 'global' and then check for 'global' here. What do you think?

econml/ortho_forest.py
```diff
@@ -254,96 +321,54 @@ def const_marginal_effect(self, X):
...
-
-        Parameters
-        ----------
-        X_single : array-like, shape (d_x, )
-            Feature vector that captures heterogeneity for one sample.
-
-        stderr : boolean (default=False)
-            Whether to calculate the covariance matrix via bootstrap-of-little-bags.
-        """
+        results = Parallel(n_jobs=self.n_jobs, verbose=3)(
```
kbattocchi, line 331: Out of curiosity, have you compared performance vs. adding a `prefer="threads"` hint?  Given that most of the computation on each thread looks to be done via numpy, this might be faster since it could avoid copying large arrays across processes (but might be slower due to the GIL)

econml/sklearn_extensions/model_selection.py
```diff
@@ -190,3 +201,148 @@ def split(self, X, y, sample_weight=None):
...
+            y = le.fit_transform(y)
+        elif y.ndim == 2:
+            y_enc = np.zeros_like(y, dtype=np.int)
+            for i_label in range(y.shape[1]):
+                y_enc[:, i_label] = LabelEncoder().fit_transform(y[:, i_label])
+            y = y_enc
+
+    # We clone the estimator to make sure that all the folds are
+    # independent, and that it is pickle-able.
+    parallel = Parallel(n_jobs=n_jobs, verbose=verbose,
```
kbattocchi, line 312: Consider also comparing the performance with `prefer="threads"` here, too.

econml/ortho_forest.py
vsyrgkanis: it's slower. The point is that each process uses a sub-slice, which essentially is a new numpy array.

vsyrgkanis: My only left over is that processes have many times a big overhead and there is something related to the auto-batching that parallel does that slows down the first few periods. I will leave that to someone else to figure out if there is a better way to call parallel.

econml/ortho_forest.py
vsyrgkanis: Let me know if you think I should change to something like this. Happy either way.

moprescu: Note that `nuisance_estimates` is usually a tuple of two arrays, so being `0` is a very special case. We can also do `global` or `np.nan`. If none of these seem appealing, we can also just make the method non-static. Either of these will remove a lot of overhead.

vsyrgkanis: Sounds good. I'll hcange to global. Though still we need the global_residualization in the generator of the nuisance estimator, as it needs to return the 0 or the global key when global residualziation is enabled. But indeed we are removing the other generators

econml/ortho_forest.py
vsyrgkanis: Made changes.

econml/ortho_forest.py
vsyrgkanis: Made changes. Let me know

moprescu: Looks good!

PR #318, opened 2020-11-16
LinearModelFinalInference was not subtracting intercept when model_final has one

vsyrgkanis:
subtracted intercept from pointwise estimate in intercetp in linear model final inference and raised extra warning. added corresponding tests that intercept is ignored.


kbattocchi (commented):
Should there be corresponding changes to `GenericFinalInferenceDiscrete`?

vsyrgkanis:
> Should there be corresponding changes to `GenericFinalInferenceDiscrete`?

Nope in the Discrete setting, intercept is the cate intercept as bias_part_of_coef=False.

kbattocchi (approved):
LGTM

PR #319, opened 2020-11-17
Move StatsModelsLinearRegression

### LINKED ISSUE #118: Restructuring utilities
Utilities currently contain a large set of classes and hard to maintain the code. We might want to split some of these classes into other python files or even better in some folder structure. 

For instance there is a large set of classes related to sklearn. Among those classes there is a large set of wrapper style classes that wrap sklearn estimators. There are also other classes that enhance sklearn estimators such as the `WeightedLasso`. As a first we could move all these sklearn related utilities in a separate structure/folder/file. For instance, we could move into a structure of the form: 
sklearn_ext.wrappers.WeightedModelWrapper
sklearn_ext.extensions.WeightedLasso
etc.

kbattocchi:
This moves `StatsModelsLinearRegression` from the `econml.utilities` package to `econml.sklearn_extensions.linear_model` package, deprecates the unused `LassoCVWrapper`, and cleans up some of our imports a bit (fixes #118).

Given that `econml.sklearn_extensions.linear_model` imports from `econml.utilities`, I couldn't find a clean way to leave a deprecated version of this class in `utilities`, but it seems unlikely that any user directly depends on this class in a way that will break.

### DIFF
--- a/econml/utilities.py
+++ b/econml/utilities.py
@@ -14,7 +14,6 @@ from sklearn.base import TransformerMixin, BaseEstimator
 from sklearn.linear_model import LassoCV, MultiTaskLassoCV, Lasso, MultiTaskLasso
 from functools import reduce, wraps
 from sklearn.utils import check_array, check_X_y
-from statsmodels.tools.tools import add_constant
 import warnings
 from warnings import warn
 from sklearn.model_selection import KFold, StratifiedKFold, GroupKFold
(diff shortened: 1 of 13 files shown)


econml/utilities.py
```diff
@@ -988,379 +988,6 @@ def _safe_norm_ppf(q, loc=0, scale=1):
...
-
-        Returns
-        -------
-        self : StatsModelsLinearRegression
-        """
-        # TODO: Add other types of covariance estimation (e.g. Newey-West (HAC), HC2, HC3)
-        X, y, sample_weight, sample_var = self._check_input(X, y, sample_weight, sample_var)
-
-        if self.fit_intercept:
-            X = add_constant(X, has_constant='add')
```
heimengqi, line 1077: Since you also clean imports in this PR, this import could also be deleted

heimengqi (approved):
Looks good to me!

econml/utilities.py
kbattocchi: Sorry, it's not clear to me what you're referring to here - could you explain?

heimengqi: ohh sorry I should point to the function name, `add_constant` can be deleted from import.

kbattocchi: I removed it, thanks for pointing it out.

PR #324, opened 2020-11-19
Update README.md

vsyrgkanis:
Updated some examples and also updated bibliography to include some papers at the core of the library that we were missing.


kbattocchi (approved):
LGTM

PR #327, opened 2020-11-20
Prepare bugfix release

kbattocchi:
This ought to fix reliance on newer sklearn versions, and also cleans up a few documentation issues.

### DIFF
--- a/econml/sklearn_extensions/ensemble.py
+++ b/econml/sklearn_extensions/ensemble.py
@@ -10,6 +10,7 @@ import scipy.sparse
 import threading
 import sparse as sp
 import itertools
+from joblib import effective_n_jobs, Parallel, delayed
 from sklearn.utils import check_array, check_X_y, issparse
 from sklearn.ensemble.forest import ForestRegressor, _accumulate_prediction
 from sklearn.tree import DecisionTreeRegressor
@@ -17,12 +18,8 @@ from sklearn.model_selection import train_test_split
 from sklearn.base import RegressorMixin
 from warnings import catch_warnings, simplefilter, warn
 from sklearn.exceptions import DataConversionWarning, NotFittedError
-from sklearn.tree._tree import DTYPE, DOUBLE
 from sklearn.utils import check_random_state, check_array, compute_sample_weight
-from sklearn.utils._joblib import Parallel, delayed
-from sklearn.utils.fixes import _joblib_parallel_args
 from sklearn.utils.validation import check_is_fitted
-from sklearn.ensemble.base import _partition_estimators
 
 MAX_INT = np.iinfo(np.int32).max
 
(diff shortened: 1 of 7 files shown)


econml/sklearn_extensions/ensemble.py
```diff
@@ -590,10 +586,9 @@ def _mean_fn(self, X, fn, acc, slice=None):
             n_estimators = slice[1] - slice[0]
 
         # Assign chunk of trees to jobs
-        n_jobs, _, _ = _partition_estimators(n_estimators, self.n_jobs)
+        n_jobs = min(effective_n_jobs(self.n_jobs), n_estimators)
         lock = threading.Lock()
-        Parallel(n_jobs=n_jobs, verbose=self.verbose,
-                 **_joblib_parallel_args(require="sharedmem"))(
+        Parallel(n_jobs=n_jobs, verbose=self.verbose, require="sharedmem")(
```
vsyrgkanis, line 591: I believe joblib args is doing sth more than just setting require=sharedmem. I believe it also sets backend='threading' when require='sharedmem' is set. Can we verify that the same functionality is eanbled.

econml/sklearn_extensions/ensemble.py
```diff
@@ -555,8 +552,7 @@ def fit(self, X, y, sample_weight=None, sample_var=None):
                                                                        int(np.ceil(self.subsample_fr_ *
                                                                                    (X.shape[0] // 2))),
                                                                        replace=False)])
-            res = Parallel(n_jobs=self.n_jobs, verbose=self.verbose,
-                           **_joblib_parallel_args(prefer='threads'))(
+            res = Parallel(n_jobs=self.n_jobs, verbose=self.verbose, prefer='threads')(
```
vsyrgkanis, line 555: Similarly here, can we verify that _joblib_parallel_args only returns prefer='threads' and not some other keyword too.

econml/sklearn_extensions/ensemble.py
```diff
@@ -462,7 +459,7 @@ def fit(self, X, y, sample_weight=None, sample_var=None):
         """
 
         # Validate or convert input data
-        X = check_array(X, accept_sparse="csc", dtype=DTYPE)
+        X = check_array(X, accept_sparse="csc", dtype=np.float32)
```
vsyrgkanis, line 462: We are using the _tree backend of sklearn. Would it be a problem if they change the native dtype in their _tree implementation. I suspect it's benign and we'll discover it.

vsyrgkanis (approved):
Looks good. Two minor comments

econml/sklearn_extensions/ensemble.py
vsyrgkanis: this is handled in the joblib versions we require

econml/sklearn_extensions/ensemble.py
vsyrgkanis: this is handled in the joblib versions we require

econml/sklearn_extensions/ensemble.py
kbattocchi: There is some extra logic in the sklearn private method, but it's only applicable to versions of joblib<=0.11 while we require joblib 0.13 (which does not need their fix)

econml/sklearn_extensions/ensemble.py
kbattocchi: Yes, to some extent we have just moved the problem around somewhat; however as you say it seems like we would discover the problem if they change datatypes and it's one less internal thing that we're referencing this way.

PR #328, opened 2020-11-22
Added GridSearchCV list that can help auto select among multiple models

vsyrgkanis:
Added an sklearn extension in model_selection GridSearchCVList that you can pass a list of esitmators and list of parameter grids. Can be helpful for cross_val model selection outside of econml for first stages. Showed how to use that outside of econml in the Forest Learners basic example notebook.


kbattocchi (approved):
LGTM

kbattocchi (commented):
But consider covering the new code with tests

PR #336, opened 2020-12-08
add shap value features on each estimator (depends on master branch of shap)

heimengqi:
- add shap_values method on each of our estimators except for OrthoIV classes, which plan to add after we have a more complete version of subclasses
- Currently this PR depends on the master branch of shap package, not the released version
- Fix the comparison table


econml/cate_estimator.py
```diff
@@ -660,6 +751,52 @@ def summary(self, alpha=0.1, value=0, decimals=3, feat_name=None, treatment_name
...
+        if output_names is None:
+            output_names = [f"Y{i}" for i in range(d_y)]
+
+        Xs, T = broadcast_unit_treatments(X, d_t)
+        if self.featurizer is not None:
+            F = self.featurizer.transform(Xs)
+        else:
+            F = Xs
+        d_x = F.shape[1]
+        X_new = cross_product(F, T)
```
kbattocchi, line 769: This looks sketchy - what if the estimator doesn't have a featurizer attribute?  E.g. in GenericModelFinalInference we check for the attribute and use None if there isn't one.

Also, in the const_marginal_effect_inference method there we first call featurizer.transform and then call broadcast_unit_treatments rather than the other way around - I believe the results should be the same, but might be nice to be more consistent.

econml/cate_estimator.py
```diff
@@ -660,6 +751,52 @@ def summary(self, alpha=0.1, value=0, decimals=3, feat_name=None, treatment_name
...
+        if self.fit_cate_intercept and feature_names is not None:
+            feature_names = ["Intercept"] + feature_names
+        if feature_names is not None:
+            feature_names = np.tile(feature_names, d_t)
+
+        shap_outs = defaultdict(dict)
+        for i in range(d_t):
+            X_sub = X_new[T[:, i] == 1]
+            ind_x = list(range(i * d_x, (i + 1) * d_x)
+                         )[1:] if self.fit_cate_intercept else list(range(i * d_x, (i + 1) * d_x))
```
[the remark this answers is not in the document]
kbattocchi: But actually it might be cleaner to define ind_x as a 2D array outside the loop:

```
ind_x = np.arange(d_t * d_x).reshape(d_t, d_x)
if self.fit_cate_intercept:
    ind_x = ind_x[:,1:]
```

and then inside the loop just take the `i`th row

econml/dml.py
```diff
@@ -1162,6 +1176,13 @@ def fit(self, Y, T, X=None, W=None, *, sample_weight=None, sample_var=None, grou
                            sample_weight=sample_weight, sample_var=None, groups=groups,
                            inference=inference)
 
+    def shap_values(self, X, *, feature_names=None, treatment_names=None, output_names=None):
+        model = copy.deepcopy(self.model_cate)
+        model.__class__ = RandomForestRegressor
```
kbattocchi, line 1181: This looks hacky - at least add a comment explaining why it works.

econml/drlearner.py
```diff
@@ -558,6 +562,46 @@ def cate_feature_names(self, input_feature_names=None):
...
+                    shap_out_copy = deepcopy(shap_out)
+                    shap_out_copy.base_values = shap_out_copy.base_values[..., i]
+                    shap_out_copy.values = shap_out_copy.values[..., i]
+                    shap_outs[output_names[0]][treatment_names[i]] = shap_out_copy
+            else:
+                shap_outs[output_names[0]][treatment_names[0]] = shap_out
+            return shap_outs
+        else:
+            return super()._shap_values(super().model_final.models_cate, F, feature_names=feature_names,
+                                        treatment_names=treatment_names, output_names=output_names)
```
kbattocchi, line 602: A lot of this logic looks extremely similar to LinearCateEstimator's `_shap_values` and also LinearModelFinalCateEstimatorMixin's `shap_values`.  Is it possible to extract the common logic so that it's all in just one place and making changes will be easier in the future?

econml/tests/test_shap.py
```diff
@@ -0,0 +1,131 @@
...
+from econml.ortho_forest import *
+from econml.drlearner import *
+from econml.metalearners import *
+from sklearn.linear_model import LinearRegression, LogisticRegression
+from sklearn.ensemble import RandomForestRegressor, RandomForestClassifier
+from sklearn.preprocessing import PolynomialFeatures
+
+
+class TestShap(unittest.TestCase):
+    def test_continues_t(self):
```
kbattocchi, line 17: Spelling:
```suggestion
    def test_continuous_t(self):
```

econml/utilities.py
```diff
@@ -1218,13 +1219,11 @@ def transpose_dictionary(d):
...
     Transpose a dictionary of dictionaries, bringing the keys from the second level
     to the top and vice versa
-
     Parameters
     ----------
     d: dict
         The dictionary to transpose; the values of this dictionary should all themselves
         be dictionaries
-
     Returns
```
kbattocchi, line 1227: I think this looks better with the line breaks, no?

econml/utilities.py
```diff
@@ -1274,6 +1273,46 @@ def predict(self, X):
         return self._clf.predict_proba(X)[:, 1:]
 
 
+def _shap_explain_cme(cme_model, X, d_y, d_t, feature_names=None, treatment_names=None, output_names=None):
+    """
+    Transpose a dictionary of dictionaries, bringing the keys from the second level
+    to the top and vice versa
+    Parameters
```
kbattocchi, line 1280:
```suggestion
    to the top and vice versa

    Parameters
```

econml/utilities.py
```diff
@@ -1274,6 +1273,46 @@ def predict(self, X):
...
+        number of outcome.
+    d_t: scalar
+        number of treatment (exclude control in discrete treatment scenario).
+    feature_names: optional None or list of strings of length X.shape[1] (Default=None)
+        The names of input features.
+    treatment_names: optional None or list (Default=None)
+        The name of treatment. In discrete treatment scenario, the name should not include control name.
+    output_names:  optional None or list (Default=None)
+        The name of the outcome.
+    Returns
```
kbattocchi, line 1296:
```suggestion
        The name of the outcome.

    Returns
```

setup.cfg
```diff
@@ -49,6 +49,7 @@ install_requires =
     graphviz
     matplotlib
     pandas < 1.1
+    shap @ git+https://github.com/slundberg/shap.git#egg=shap
```
kbattocchi, line 52: [minor] I believe in this format the egg is not required:
```suggestion
    shap @ git+https://github.com/slundberg/shap.git
```

kbattocchi (changes requested):
Mostly looks good, but I added some feedback.

econml/cate_estimator.py
```diff
@@ -436,6 +440,93 @@ def const_marginal_effect_inference(self, X=None):
...
+            except Exception as e:
+                print("Final model can't be parsed, explain const_marginal_effect() instead!")
+                return _shap_explain_cme(self.const_marginal_effect, X, d_y, d_t, feature_names, treatment_names,
+                                         output_names)
+            shap_out = explainer(X)
+            if d_y > 1:
+                for j in range(d_y):
+                    shap_out_copy = deepcopy(shap_out)
+                    shap_out_copy.base_values = shap_out_copy.base_values[..., j]
+                    shap_out_copy.values = shap_out_copy.values[..., j]
```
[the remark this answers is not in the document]
heimengqi: Good Point! I just checked slicing, it could only slice a specific row or column, but in this case here, we only slice the `.base_values` and `.values`, `.data` is same for each dimension of Y.

econml/cate_estimator.py
heimengqi: I think all the DML subclasses inherit `_BaseDML`, they should all have featurizer attribute? Also this class `LinearModelFinalCateEstimatorMixin` assumes final stage is a linear model, should always have featurizer? 

Yes but I follow the logic from the prediction function of `_FinalWrapper`, which call broadcast function first and then transform X (under a `_combine()` function), I could change the order here to be consistent with `cont_marginal_effect_inference`.

econml/cate_estimator.py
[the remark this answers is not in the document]
heimengqi: Great! Done!

econml/dml.py
heimengqi: Done.

econml/drlearner.py
heimengqi: Yeah I thought about that.  the only part repeats multiple times is 
```
        d_t = self._d_t[0] if self._d_t else 1

        if treatment_names is None:
            treatment_names = [f"T{i}" for i in range(d_t)]
        if output_names is None:
            output_names = [f"Y{i}" for i in range(d_y)]
```
Any idea on how to avoid duplicates for this?

For the loop to extract shap values of each T and each Y, they look similar but they are different given the structure of final model. `drlearner` for mulitple T is one single multi-output model, `dml` for multiple T should filter cate each given T, and the for loop under `_shap_values` for multiple T is one model per T, all those slice the `shap_values` in different way.

kbattocchi: If the details vary because the indexing is subtly different then maybe there's not much we can do to simplify the logic, in which case don't worry about it.

econml/cate_estimator.py
kbattocchi: The DML classes will all have a featurizer attribute, and at the moment none of our other classes inherits from this mixin, so _for now_ this logic is totally safe.  However, there's nothing to prevent someone from adding a new estimator in the future that inherits from this mixin but doesn't have a featurizer, so it would probably be better to check for the existence of the attribute as we do in GenericModelFinalInference.

heimengqi: Sure, add attribute check for `.featurizer` and `.fit_cate_intercept`

kbattocchi (approved):

econml/cate_estimator.py
```diff
@@ -8,9 +8,13 @@
 from functools import wraps
 from copy import deepcopy
 from warnings import warn
+from collections import defaultdict
+import shap
```
vsyrgkanis, line 12: We should create a single python file shap.py where all the interaction with shap happens. shap should only be imported in that file and nowhere else. All functionality that directly calls shap, should be happening in that file and the functions in any other files should be calling the intermediate functionalities we want.

This way we create a single interaction point and if shap changes anything in the future all we would need to do is change our shap.py file to handle the changes and offer the same api to the remainder of the code.

Also this would allow our users to not have any breaking changes unless they wont to use shap.

econml/cate_estimator.py
```diff
@@ -8,9 +8,13 @@
...
+from slicer import Alias
```
vsyrgkanis, line 13: Similarly Alias should only be imported in the shap.py file.

econml/cate_estimator.py
```diff
@@ -8,9 +8,13 @@
...
 from .inference import BootstrapInference
 from .utilities import (tensordot, ndim, reshape, shape, parse_final_model_params,
-                        inverse_onehot, Summary, get_input_columns)
+                        inverse_onehot, Summary, get_input_columns, broadcast_unit_treatments,
+                        cross_product, _shap_explain_cme)
```
vsyrgkanis, line 17: _shap_explain_cme should move into shap.py

econml/cate_estimator.py
```diff
@@ -456,6 +460,93 @@ def const_marginal_effect_inference(self, X=None):
...
+            The name of the outcome.
+
+        Returns
+        -------
+        shap_outs: nested dictionary of Explanation object
+            A nested dictionary by using each Y and each T as key and the shap_values explanation object as value.
+
+
+        """
+        d_t = self._d_t[0] if self._d_t else 1
```
vsyrgkanis, line 484: This naming of the dict seems to appear in multiple places. It should be written only once somewhere. Maybe in the shap.py file as a utility function.

econml/cate_estimator.py
```diff
@@ -456,6 +460,93 @@ def const_marginal_effect_inference(self, X=None):
...
+
+        Returns
+        -------
+        shap_outs: nested dictionary of Explanation object
+            A nested dictionary by using each Y and each T as key and the shap_values explanation object as value.
+        """
+
+        d_t = self._d_t[0] if self._d_t else 1
+        d_y = self._d_y[0] if self._d_y else 1
+        if treatment_names is None:
```
vsyrgkanis, line 518: Same here. This snippet appears in multiple places in identical manner.

econml/cate_estimator.py
```diff
@@ -456,6 +460,93 @@ def const_marginal_effect_inference(self, X=None):
...
+            The names of input features.
+        treatment_names: optional None or list (Default=None)
+            The name of treatment. In discrete treatment scenario, the name should not include control name.
+        output_names:  optional None or list (Default=None)
+            The name of the outcome.
+
+        Returns
+        -------
+        shap_outs: nested dictionary of Explanation object
+            A nested dictionary by using each Y and each T as key and the shap_values explanation object as value.
```
vsyrgkanis, line 513: The docstrings might need to mention what the default Y and T keys are , i.e. "Y0, Y1" etc and "T0, T1" etc, when treatment_names =None and output_names=None

econml/cate_estimator.py
```diff
@@ -456,6 +460,93 @@ def const_marginal_effect_inference(self, X=None):
...
+        Parameters
+        ----------
+        models: a single estimator or a list of estimators with one estimator per treatment
+            models for the model's final stage model.
+        X: (m, d_x) matrix
+            Features for each sample. Should be in the same shape of fitted X in final stage.
+        feature_names: optional None or list of strings of length X.shape[1] (Default=None)
+            The names of input features.
+        treatment_names: optional None or list (Default=None)
+            The name of treatment. In discrete treatment scenario, the name should not include control name.
```
vsyrgkanis, line 506: Maybe rephrase "control name". Say: should not include the name of the baseline treatment (i.e. the control treatment, which by default is the alphabetically smaller)

vsyrgkanis: Is this btw the same in other places where we use treatment_names? If yes, then I'm not sure if those docstrings are pointing this out.

econml/cate_estimator.py
```diff
@@ -456,6 +460,93 @@ def const_marginal_effect_inference(self, X=None):
...
+        d_t = self._d_t[0] if self._d_t else 1
+        d_y = self._d_y[0] if self._d_y else 1
+        if treatment_names is None:
+            treatment_names = [f"T{i}" for i in range(d_t)]
+        if output_names is None:
+            output_names = [f"Y{i}" for i in range(d_y)]
+
+        return _shap_explain_cme(self.const_marginal_effect, X, d_y, d_t, feature_names, treatment_names, output_names)
+
+    def _shap_values(self, models, X, *, feature_names=None, treatment_names=None, output_names=None):
```
vsyrgkanis, line 493: This function seems a bit out of place here, but seems ok. Maybe a different name would be more appropriate (e.g. _shap_explain_model_cate)?

Also maybe this is not the place for this and maybe similar to _shap_explain_cme, we should have in the shap.py file another utility function called _shap_explain_model_cate that takes as input: (estimator, models, X, feature_names, treatment_names, output_names)?

econml/cate_estimator.py
```diff
@@ -456,6 +460,93 @@ def const_marginal_effect_inference(self, X=None):
...
+        """
+
+        d_t = self._d_t[0] if self._d_t else 1
+        d_y = self._d_y[0] if self._d_y else 1
+        if treatment_names is None:
+            treatment_names = [f"T{i}" for i in range(d_t)]
+        if output_names is None:
+            output_names = [f"Y{i}" for i in range(d_y)]
+
+        if not isinstance(models, list):
```
vsyrgkanis, line 523: This logic seems very particular to the case of discrete treatments, where we have a single model per treatment. This logic should not appear in cate_estimator. Most prob once this function moves simply as a utility function, then it will be fine.

econml/cate_estimator.py
```diff
@@ -456,6 +460,93 @@ def const_marginal_effect_inference(self, X=None):
...
+        background = shap.maskers.Independent(X, max_samples=X.shape[0])
+
+        shap_outs = defaultdict(dict)
+        for i in range(d_t):
+            try:
+                explainer = shap.Explainer(models[i], background,
+                                           feature_names=feature_names)
+            except Exception as e:
+                print("Final model can't be parsed, explain const_marginal_effect() instead!")
+                return _shap_explain_cme(self.const_marginal_effect, X, d_y, d_t, feature_names, treatment_names,
```
vsyrgkanis, line 536: What happens if only one of the models fails to be built by the explainer? Why not use _shap_explain_cme only for that coordinate??

We could internally to _shap_explain_cme, build a wrapper for const_marginal_effect that returns only the i-th coordinate of const_marginal_effect and pass that function to shap. Then shap will return shap_Values only for that coordinate and we could be using the model explainers for the rest. 

Though I guess it's not clear what the compiutational benefit will be with this, since once you call shap on a generic function it doesn't change much the computation if that function is single output or multi-output. So maybe the current version is good too.

econml/cate_estimator.py
```diff
@@ -456,6 +460,93 @@ def const_marginal_effect_inference(self, X=None):
...
+                                         output_names)
+            shap_out = explainer(X)
```
vsyrgkanis, line 538: This loop also seems to be a snippet of code that't repeated elsewhere and you might want to write a utility function in shap.py that transposes the shap and is used internally within _shap_explain_model_cate and in _shap_explain_cme

econml/cate_estimator.py
```diff
@@ -685,6 +776,51 @@ def summary(self, alpha=0.1, value=0, decimals=3, feature_names=None, treatment_
         if len(smry.tables) > 0:
             return smry
 
+    def shap_values(self, X, *, feature_names=None, treatment_names=None, output_names=None):
+        d_t = self._d_t[0] if self._d_t else 1
```
vsyrgkanis, line 780: maybe here build a _shap_explain_linear_model_cate in shap.py that contains this logic here and also re-uses the utilities for creating the treatment names in the beginning and the utility that transposes the shap values at the end.

econml/cate_estimator.py
```diff
@@ -685,6 +776,51 @@ def summary(self, alpha=0.1, value=0, decimals=3, feature_names=None, treatment_
...
+        if feature_names is not None:
+            feature_names = np.tile(feature_names, d_t)
+        # index of X columns to filter for each T
+        ind_x = np.arange(d_t * d_x).reshape(d_t, d_x)
+        if self.fit_cate_intercept:  # skip intercept
+            ind_x = ind_x[:, 1:]
+
+        shap_outs = defaultdict(dict)
+        for i in range(d_t):
+            X_sub = X_new[T[:, i] == 1]
```
vsyrgkanis, line 804: I don't understand what this line does???

vsyrgkanis: Can you add a comment around it.

econml/cate_estimator.py
```diff
@@ -456,6 +460,93 @@ def const_marginal_effect_inference(self, X=None):
...
+        if treatment_names is None:
+            treatment_names = [f"T{i}" for i in range(d_t)]
+        if output_names is None:
+            output_names = [f"Y{i}" for i in range(d_y)]
+
+        if not isinstance(models, list):
+            models = [models]
+        assert len(models) == d_t, "Number of final stage models don't equals to number of treatments!"
+        # define masker by using entire dataset, otherwise Explainer will only sample 100 obs by default.
+        background = shap.maskers.Independent(X, max_samples=X.shape[0])
```
vsyrgkanis, line 527: Why not use the default max_samples as in shap? Wouldn't this make it much slower at no big quality improvement? Why does shap use max_samples=100. 

Also maybe this could be an input to the shap_values function then if it's something that the user should be able to control.

econml/dml.py
```diff
@@ -938,6 +941,17 @@ def __init__(self,
...
                          random_state=random_state)
 
+    def shap_values(self, X, *, feature_names=None, treatment_names=None, output_names=None):
+        if self.featurizer is not None:
+            F = self.featurizer.transform(X)
+        else:
+            F = X
+        feature_names = self.cate_feature_names(feature_names)
+
+        return super()._shap_values(self.model_cate, F, feature_names=feature_names,
```
vsyrgkanis, line 951: Why isn't featurization handled within _shap_values? I suspect once _shap_values moves into a utility function in shap.py, we should push the featurization inside _shap_values (e.g. pass featurizer as an extra argument or None)

econml/drlearner.py
```diff
@@ -564,6 +568,46 @@ def cate_feature_names(self, feature_names=None):
...
             raise AttributeError("Featurizer does not have a method: get_feature_names!")
 
+    def shap_values(self, X, *, feature_names=None, treatment_names=None, output_names=None):
+        if self.featurizer is not None:
+            F = self.featurizer.transform(X)
+        else:
+            F = X
+        feature_names = self.cate_feature_names(feature_names)
+
+        if self._multitask_model_final:
```
vsyrgkanis, line 578: Again maybe create a _shap_explain_multitask_model_cate in shap.py that performs this logic if self._multitask_model_final.

econml/metalearners.py
```diff
@@ -458,3 +458,8 @@ def _fit_weighted_pipeline(self, model_instance, X, y, sample_weight):
         else:
             last_step_name = model_instance.steps[-1][0]
             model_instance.fit(X, y, **{"{0}__sample_weight".format(last_step_name): sample_weight})
+
+    def shap_values(self, X, *, feature_names=None, treatment_names=None, output_names=None):
```
vsyrgkanis, line 462: Why is this the only metalearner that has a model_final based shap_values?

econml/utilities.py
```diff
@@ -1303,6 +1304,48 @@ def predict(self, X):
...
+    shap_outs: nested dictionary of Explanation object
+        A nested dictionary by using each Y and each T as a key and the shap_values explanation object as the value.
+
+    """
+    # define masker by using entire dataset, otherwise Explainer will only sample 100 obs by default.
+    background = shap.maskers.Independent(X, max_samples=X.shape[0])
+    shap_outs = defaultdict(dict)
+    for i in range(d_y):
+        for j in range(d_t):
+            def cmd_func(X):
```
vsyrgkanis, line 1340: It would be great if we could call shap fewer times!! 
For instance I expect that d_t would be > 1 more often than d_y.

Can shap explain a multi-output function?

If yes then we could be passing the whole vector of cme's for all t's at once and then just call shap for each d_y separately. 

This way a lot of shap compute will be re-used and it will be much faster.

vsyrgkanis (changes requested):
Another general request: I think we should have a Explain Cate with SHAP.ipynb notebook that portrays all the functionality with SHAP. It seems important. But we could also do this post-release and prioritize all the other changes requested.

econml/cate_estimator.py
```diff
@@ -8,12 +8,17 @@
 from functools import wraps
 from copy import deepcopy
 from warnings import warn
+from collections import defaultdict
```
vsyrgkanis, line 11: Is defaultdict needed?

econml/shap.py
```diff
@@ -0,0 +1,263 @@
...
+                                           data=shap_out.data, main_effects=main_effects,
+                                           feature_names=shap_out.feature_names)
+                shap_outs[output_names[j]][treatment_names[i]] = shap_out_new
+        else:
+            shap_outs[output_names[0]][treatment_names[i]] = shap_out
+
+    return shap_outs
+
+
+def _shap_explain_dml_model_cate(model_final, X, T, d_t, d_y, fit_cate_intercept,
```
vsyrgkanis, line 115: This seems more about a linear model final model cate as opposed to specific to dml. So different naming might be better.

econml/shap.py
```diff
@@ -0,0 +1,263 @@
...
+                                 feature_names=None, treatment_names=None, output_names=None):
+    """ 
+    The method to explain `model_cate` of parametric final stage DML using shap Explainer()
```
vsyrgkanis, line 118: Maybe the docstring could explain what types of cate models this explains, ie.. linear models that were fitted on the cross_product of featurizer(X) and T.

This way if we ever need this functionality somwhere else we can still re-use it (e.g. in LinearDMLIV etc.)

econml/cate_estimator.py
heimengqi: Removed it from importing

econml/shap.py
heimengqi: yes, but like your comment below said, it's specific for fitting on cross product of featurizer(X) and T, for example it doesn't work for linear DRLearner...any name recommendation?

econml/shap.py
heimengqi: Done

econml/dml.py
```diff
@@ -40,17 +40,20 @@
 from sklearn.base import TransformerMixin, clone
 from sklearn.exceptions import NotFittedError
 from sklearn.linear_model import (ElasticNetCV, LassoCV, LogisticRegressionCV)
+from sklearn.ensemble import RandomForestRegressor
 from sklearn.model_selection import KFold, StratifiedKFold, check_cv
 from sklearn.pipeline import Pipeline
 from sklearn.preprocessing import (FunctionTransformer, LabelEncoder,
                                    OneHotEncoder)
 from sklearn.utils import check_random_state
+import copy
```
vsyrgkanis, line 49: Do we need to import copy?

econml/drlearner.py
```diff
@@ -28,22 +28,26 @@
 """
 
 from warnings import warn
+from copy import deepcopy
+from collections import defaultdict
```
vsyrgkanis, line 32: Do we need to import defaultdict?

econml/shap.py
```diff
@@ -0,0 +1,263 @@
...
+            for j in range(d_y):
+                base_values = shap_out.base_values[..., j]
+                main_effects = shap_out.main_effects[..., ind_x[i], j]
+                values = shap_out.values[..., ind_x[i], j]
+                shap_out_new = shap.Explanation(values, base_values=base_values, data=data, main_effects=main_effects,
+                                                feature_names=feature_names)
+                shap_outs[output_names[j]][treatment_names[i]] = shap_out_new
+        else:
+            values = shap_out.values[..., ind_x[i]]
+            main_effects = shap_out.main_effects[..., ind_x[i], :]
```
vsyrgkanis, line 168: We should be collapsing this dimension, i.e.  
```python
main_effects = shap_out.main_effects[..., ind_x[i], 0]
```

econml/shap.py
```diff
@@ -0,0 +1,263 @@
...
+    output_names: List
+    """
+
+    d_t = d_t[0] if d_t else 1
+    d_y = d_y[0] if d_y else 1
+    if treatment_names is None:
+        treatment_names = [f"T{i}" for i in range(d_t)]
+    if output_names is None:
+        output_names = [f"Y{i}" for i in range(d_y)]
+    return (d_t, d_y, treatment_names, output_names)
```
vsyrgkanis, line 263: Maybe add a newline at end of file

econml/shap.py
vsyrgkanis: _shap_explain_joint_linear_model_cate 
in that you esitmate a single linear model jointly for all treatments

azure-pipelines.yml
```diff
@@ -66,6 +66,9 @@ jobs:
         - script: 'pip install sklearn-contrib-lightning'
           displayName: 'Install lightning'
 
+        - script: 'pip install --force-reinstall --no-cache-dir shap'
```
vsyrgkanis, line 69: Installing shap from release, also installs the latest scikit-learn, which is 0.24. So then building the docs crashes because it cannot run dml.py, since it dml.py imports the subsampledhonestforest which uses the deprecated scikit-learn.ensemble.forest module! 

We should just remove this installation instruction.

econml/cate_estimator.py
heimengqi: I don't think this will slower the computation much, since it's used to compute `base_values` (mean of prediction), the reason I change it since I also want to compare the `base_value` with our mean of `const_marginal_effect` in test to make sure we extract each Y and T correctly

econml/metalearners.py
heimengqi: all the other metalearners they don't have a straightforward final stage to estimate `theta(x)` directly?

econml/utilities.py
heimengqi: Yes, it does support multi-output function, now I change to what you recommend! Thanks!

econml/cate_estimator.py
vsyrgkanis: Is this addressed?

econml/cate_estimator.py
vsyrgkanis: Is this addressed?

econml/cate_estimator.py
vsyrgkanis: This seems addresed.

econml/cate_estimator.py
vsyrgkanis: This seems addressed.

econml/cate_estimator.py
vsyrgkanis: Is this addressed?

econml/cate_estimator.py
vsyrgkanis: Is this addressed?

econml/dml.py
vsyrgkanis: Is this addressed? It's ok to keep as is.

econml/cate_estimator.py
```diff
@@ -8,12 +8,15 @@
 from functools import wraps
 from copy import deepcopy
 from warnings import warn
+import shap
```
vsyrgkanis, line 11: We should remove this import

vsyrgkanis (approved):
All comments have been addressed!

PR #337, opened 2020-12-09
Retain column names from input dataframes/series

moprescu:
Functions like `cate_feature_names`, `summary`, `population_summary`
display the input column names if inputs are pandas DataFrames or
Series.
In this commit:
* changed cate_estimator, ortho_learner and inference classes to set
  input names upon fit
* moved input verification to the base class for some estimators
* modified customer scenario notebooks to reflect changes
* added tests


econml/cate_estimator.py
```diff
@@ -9,7 +9,8 @@
 from copy import deepcopy
 from warnings import warn
 from .inference import BootstrapInference
-from .utilities import tensordot, ndim, reshape, shape, parse_final_model_params, inverse_onehot, Summary
+from .utilities import tensordot, ndim, reshape, shape, parse_final_model_params,\
+    inverse_onehot, Summary, get_input_columns
```
kbattocchi, line 13: Rather than escaping the line break, I believe that this style is preferred:

```suggestion
from .utilities import (tensordot, ndim, reshape, shape, parse_final_model_params,
                        inverse_onehot, Summary, get_input_columns)
```

econml/inference.py
```diff
@@ -623,7 +633,9 @@ def summary_frame(self, alpha=0.1, value=0, decimals=3, feat_name=None, output_n
             The output dataframe includes point estimate, standard error, z score, p value and confidence intervals
             of the estimated metric of each treatment on each outcome for each sample X[i]
         """
-
+        feat_name = self.feat_name if feat_name is None else feat_name
+        treatment_name = self.treatment_name if treatment_name is None else treatment_name
+        output_name = self.output_name if output_name is None else output_name
```
kbattocchi, line 638: I think a more pythonic way to write this is

```suggestion
        # Assign names from dataframes if not provided as arguments
        feat_name = feat_name or self.feat_name
        treatment_name = treatment_name or self.treatment_name
        output_name = output_name or self.output_name
```

(although using `or` with non-booleans still seems a bit weird to me)

econml/inference.py
```diff
@@ -689,6 +701,8 @@ def population_summary(self, alpha=0.1, value=0, decimals=3, tol=0.001, output_n
             The population summary results instance contains the different summary analysis of point estimate
             for sample X on each treatment and outcome.
         """
+        treatment_name = self.treatment_name if treatment_name is None else treatment_name
+        output_name = self.output_name if output_name is None else output_name
```
kbattocchi, line 705:
```suggestion
        treatment_name = treatment_name or self.treatment_name
        output_name = output_name or self.output_name
```

econml/inference.py
```diff
@@ -830,7 +845,8 @@ def _expand_outputs(self, n_rows):
         assert shape(self.pred)[0] == shape(self.pred_stderr)[0] == 1
         pred = np.repeat(self.pred, n_rows, axis=0)
         pred_stderr = np.repeat(self.pred_stderr, n_rows, axis=0)
-        return NormalInferenceResults(self.d_t, self.d_y, pred, pred_stderr, self.inf_type, self.fname_transformer)
+        return NormalInferenceResults(self.d_t, self.d_y, pred, pred_stderr, self.inf_type,
+                                      self.fname_transformer)
```
kbattocchi, line 849: Should also propagate the names:
```suggestion
                                      self.fname_transformer, self.feat_name, self.output_name, self.treatment_name)
```

econml/inference.py
```diff
@@ -830,7 +845,8 @@ def _expand_outputs(self, n_rows):
...
 
 
 class EmpiricalInferenceResults(InferenceResults):
```
kbattocchi, line 852: Please propagate the names in `_expand_outputs` here, too.

econml/utilities.py
```diff
@@ -544,6 +546,33 @@ def check_input_arrays(*args, validate_len=True):
...
+
+    """
+    # Type to column extraction function
+    type_to_func = {
+        pd.DataFrame: lambda x: x.columns.tolist(),
+        pd.Series: lambda x: [x.name]
+    }
+    if type(X) in type_to_func:
+        return type_to_func[type(X)](X)
+    return None
```
kbattocchi, line 573: Seems simpler without the dictionary:

```suggestion
    if type(X) == pd.DataFrame:
        return X.columns.tolist()
    elif type(X) == pd.Series:
        return [X.name]
    else:
        return None
```

econml/cate_estimator.py
```diff
@@ -42,6 +43,14 @@ def _get_inference(self, inference):
         # because inf now stores state from fitting est2
         return deepcopy(inference)
 
+    def _set_input_names(self, Y, T, X):
```
kbattocchi, line 46: Instead of explicitly adding calls to this to `fit` calls in subclasses, could it be added to the logic in `wrap_fit` to happen automatically?

kbattocchi (approved):
Added a few comments; approved subject to their resolution

econml/cate_estimator.py
moprescu: Added it to `_prefit`. Still has some issues (e.g. when the ORF overrides the wrapped fit), but I made a workaround for that.

econml/utilities.py
moprescu: I left is a dictionary in case we will add other dataframes in the future. E.g. tensorflow, spark, lightgbm have their own dataframe-like structure that carry metadata. Hopefully, we will support all of them at some point.

econml/inference.py
moprescu: Great suggestions. I am changing this everywhere. the changes are out of date, but will address in a separate commit.

econml/inference.py
moprescu: The changes are out of date, but will address in a separate commit.

econml/inference.py
moprescu: The changes are out of date, but will address in a separate commit.

econml/inference.py
moprescu: I actually realized this doesn't work with `numpy` arrays as they you can't evaluate the statement `if array`. Reverting to the previous solution.

PR #343, opened 2020-12-16
Add support for refitting and nuisance averaging

kbattocchi:


econml/_ortho_learner.py
```diff
@@ -714,6 +787,64 @@ def score(self, Y, T, X=None, W=None, Z=None, sample_weight=None):
...
+        return self._n_splits
+
+    @n_splits.setter
+    def n_splits(self, n_splits):
+        self._n_splits = n_splits
+        self._cache_invalid_message = "Setting the number of splits invalidates the cached nuisances"
+
+    @property
+    def discrete_treatment(self):
+        return self._discrete_treatment
```
heimengqi, line 814: Do we really need to expose the first stage related arguments and allow users to set something else? I think the value of `refit` is to save computation time when the nuisance models are the same and the data and results are cached. Even they reset a value here they can't only refit final stage but fit the entire pipeline again, it's the same as reinitialize a new estimator and call `fit` in term of computing time and lines of code. Also it's wired to allow user to change the `discrete_treatment` flag since T is fixed and it should be either continuous our categorical. If we only allow user to change final stage specific argument, it could also simplify tons of properties and setters across multiple classes?    

Same for `_RLearner` and classes under DML file.

econml/dml.py
```diff
@@ -605,8 +741,14 @@ def fit(self, Y, T, X=None, W=None, *, sample_weight=None, sample_var=None, grou
...
         return super().fit(Y, T, X=X, W=W,
                            sample_weight=sample_weight, sample_var=sample_var, groups=groups,
+                           cache_values=cache_values, monte_carlo_iterations=monte_carlo_iterations,
                            inference=inference)
 
+    @DML.model_final.setter
+    def model_final(self, model):
+        raise AttributeError("LinearDML final model can't be chnaged from "
+                             "StatsModelsLinearRegression(fit_intercept=False)")
+
```
heimengqi, line 751: We can't change final stage model but we could change some arguments under that model? e.g. cov_type, like what you did for `SubsampledHonestForest`

heimengqi (commented):
Mostly looks good to me. But I all agree on what you've mentioned in our call, it's annoying to loop all possible arguments and add a setter of it, and it's hard to maintain. We might need to brainstorming and see is there any alternatives.

econml/dml.py
heimengqi: Another typo here I just randomly saw it, "changed"

econml/_ortho_learner.py
kbattocchi: My idea was to make all things that you can set to some value in the initializer also available to update at any other time.  It's true that many of these are not necessary for the refit functionality alone, and also that a workaround would be to create a new instance.

However, from the user perspective I think sklearn's behavior (where you can set any attribute at any time) would be much easier to understand than a policy where you can only set the attributes that don't affect the nuisances, since it is not always completely obvious which attributes affect the nuisances (e.g. the featurizer affects the Y residuals if we use the `linear_first_stages` flag, but doesn't affect any nuisances if it's False).

Eventually, I think we should enable setting attributes even for estimators that don't have support refit (e.g. the metalearners), but for this PR I'm just starting with the OrthoLearner class hierarchy.

econml/dml.py
kbattocchi: Yes, I think that would make sense - but for now, cov_type is set based on the inference object's cov_type, it's never set directly by the user, so I just left that behavior as is for now.  Ultimately this ties back into our discussion of whether to change the DML hierarchy by unifying the LinearDML and SparseDML classes and cleaning up how we implement inference, but that's out of scope for now.

econml/_ortho_learner.py
```diff
@@ -714,6 +787,64 @@ def score(self, Y, T, X=None, W=None, Z=None, sample_weight=None):
...
+
+    @discrete_treatment.setter
+    def discrete_treatment(self, discrete_treatment):
```
vsyrgkanis, line 817: Does this need to trigger a changed in the discrete treatment transformer??

econml/dml.py
```diff
@@ -485,8 +548,76 @@ def fit(self, Y, T, X=None, W=None, *, sample_weight=None, sample_var=None, grou
...
+        # only the final model needs to change
+        self._set_final_model()
+
+    @property
+    def bias_part_of_coef(self):
+        return self.fit_cate_intercept
+
+    @_OrthoLearner.discrete_treatment.setter
+    def discrete_treatement(self, discrete_treatement):
+        # super().discrete_treatment = discrete_treatment
```
vsyrgkanis, line 593: In general: we shoulld change the name of the _RLearner.model_final to _RLearner.rlearner_model_final, and _OrthoLearner.model_final to _OrthoLearner.ortholearner_model_final and then we should just be setting these variables. That would clean up the code even if the refit stuff were not there.

econml/dml.py
```diff
@@ -764,7 +914,69 @@ def fit(self, Y, T, X=None, W=None, *, sample_weight=None, sample_var=None, grou
                                "We recommend using the LinearDML estimator for this low-dimensional setting.")
         return super().fit(Y, T, X=X, W=W,
                            sample_weight=sample_weight, sample_var=None, groups=groups,
-                           inference=inference)
+                           cache_values=cache_values, monte_carlo_iterations=None, inference=inference)
+
+    @DML.model_final.setter
```
vsyrgkanis, line 919: I feel that for all these setters, there should be a logic within a __setattr__, as opposed to a separate method for each one, just so that we reduce lines of code. But this is secondary and I'm ok with doing it post release.

econml/tests/test_dml.py
```diff
@@ -1065,3 +1064,59 @@ def test_deprecation(self):
 
         d = pickle.dumps(LinearDMLCateEstimator())
         e = pickle.loads(d)
+
+    def test_refit(self):
```
vsyrgkanis, line 1068: I think we need several more tests here that the new params affect the final stage. Things like changing fit_intercept and then testing that all our effect methods take the change after refitting. 

Similarly maybe some other change in sparse_linear (e.g. alpha) or n_estimators in ForestDML.

Also a test that the "score_" attribute changes accordingly to the new model. Maybe do a model_selection experiment, where you run refit with two-three hyperparams, with 1 of them being the obvious best and the rest leading to random noise and read the score_ method each time after fit, and then make sure that the correct final model is chosen.

Similarly, that we can also do "score(Y, T, X, W)" out of sample and that the score returned is based on the new model.

vsyrgkanis: Also maybe a test: 
If I am in SparseLinearDML and I do:
SparseLinearDML().model_final = LinearRegression()

would that make the final model be LinearRegression()?

vsyrgkanis: Finall, I think we definitely need a "check_is_fitted" to be called in all our predict methods and raise an error if not fitted. This is even more important now, as when I change a param we need to be invalidating the fit and so if someone calls effect, they shouldn't be getting the old fitted model final results, but an error that says that estimator is not fitted yet. 

Currently our error messages are random internal errors of the estimators that some attribute is missing and not an interpretable error of the form "estimator is not fitted yet!".

vsyrgkanis: I think we should use an approach similar to sklearn, where fit sets internal attributes of the fit that are stateful attributes and doesn't alter the inptu params. (e.g. at fit time we create a model_final_ which is the fitted model final, we also set models_y_ which are the fitted models_y etc, these are all attributes created by fit and the inptu params remain unaltered as static values). 

Then sklearn has a generic check_is_fitted: https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.check_is_fitted.html
which just checks that the object has attributes ending in underscore. 

For now we could go with simpler solutions by just setting a flag: "fitted" etc.

vsyrgkanis: Post-release we also definitely need to showcase this functionality in the dml notebook. For instance having a cell that does sth like the following  as it is a superb functionality! 

```python
from econml.sklearn_extensions.linear_model import StatsModelsLinearRegression, StatsModelsRLM, DebiasedLasso
from sklearn.linear_model import ElasticNetCV, LassoCV

est = DML(model_y=RandomForestRegressor(min_samples_leaf=20, max_depth=3),
          model_t=RandomForestRegressor(min_samples_leaf=20, max_depth=3),
          model_final=StatsModelsLinearRegression(fit_intercept=False),
          random_state=123)
est.fit(Y_train, T_train, X=X_train, W=W_train, cache_values=True)

print(list(zip(est.coef_, *est.coef__interval())))
print(est.score_)

est.model_final = StatsModelsRLM(fit_intercept=True)
est.refit()
print(list(zip(est.coef_, *est.coef__interval())))
print(est.score_)

est.model_final = DebiasedLasso()
est.refit()
print(list(zip(est.coef_, *est.coef__interval())))
print(est.score_)

est.model_final = Lasso(alpha=0.1)
est.refit()
print(est.coef_)
print(est.score_)

est.model_final = ElasticNetCV(cv=3)
est.refit()
print(est.coef_)
print(est.score_)
```

I wish we could also use seamlessly the NonParamDML variants and the CausalForestDML variant, but those have a different `rlearner_model_final`, so it seems harder. but that would be very useful.

vsyrgkanis: Maybe if we just had an input flag that says: `linear_model_final`, then if that is True, we just don't use the weight trick and all functionality as in DML is good to go. and if linear_model_final=False, then we use the weight trick and we raise an error in the case when the user passes multiple treatments, that this is not allowed (I blieve that this will actually be done automatically but the model final wrapper). 

So I think it's just a matter of adding the flag: linear_model_final, and then we could do:
```python
est.linear_model_final=False
est.model_final = RandomForestRegressor()
est.refit()
```

This would allow us to do everything through refitting (except CausalForestDML, whcih uses a separate model final wrapper).

vsyrgkanis: For now we can do some of this by using the NonParamDML, but this still requires that the model accept weights. While for linear models we don't need weights. That seems good enough for now, but maybe the linear_model_final flag is still a good idea.

```python
from econml.dml import NonParamDML
from econml.sklearn_extensions.linear_model import StatsModelsLinearRegression, StatsModelsRLM, DebiasedLasso
from econml.sklearn_extensions.ensemble import SubsampledHonestForest
from sklearn.linear_model import LassoCV
from sklearn.ensemble import RandomForestRegressor, GradientBoostingRegressor

est = NonParamDML(model_y=RandomForestRegressor(min_samples_leaf=20, max_depth=3),
                  model_t=RandomForestRegressor(min_samples_leaf=20, max_depth=3),
                  model_final=SubsampledHonestForest(),
                  random_state=123)
est.fit(Y_train, T_train, X=X_train, W=W_train, cache_values=True)

print(list(zip(est.effect(X_val[:1]), *est.effect_interval(X_val[:1]))))
print(est.score_)
est.model_final = DebiasedLasso()
est.refit()
print(list(zip(est.effect(X_val[:1]), *est.effect_interval(X_val[:1]))))
print(est.score_)

est.model_final = Lasso(alpha=0.01)
est.refit()
print(est.effect(X_val[:1]))
print(est.score_)

est.featurizer = PolynomialFeatures(degree=3)
est.refit()
print(est.effect(X_val[:1]))
print(est.score_)

est.featurizer = None
est.model_final = RandomForestRegressor(min_samples_leaf=20, max_depth=3)
est.refit()
print(est.effect(X_val[:1]))
print(est.score_)

est.model_final = GradientBoostingRegressor(min_samples_leaf=20, max_depth=3)
est.refit()
print(est.effect(X_val[:1]))
print(est.score_)
```

econml/_ortho_learner.py
```diff
@@ -540,15 +558,70 @@ def fit(self, Y, T, X=None, W=None, Z=None, *, sample_weight=None, sample_var=No
...
+
+        Returns
+        -------
+        self : _OrthoLearner
+            This instance
+        """
+        assert self._cached_values, "Refit can only be called if values were cached during the original fit"
+        assert self._cache_invalid_message is None, "Can't refit: " + self._cache_invalid_message
+        cached = self._cached_values
+        self._prefit(Y=cached.Y,
```
vsyrgkanis, line 609: Why are we calling prefit?

econml/_ortho_learner.py
```diff
@@ -540,15 +558,70 @@ def fit(self, Y, T, X=None, W=None, Z=None, *, sample_weight=None, sample_var=No
...
+                                           sample_var=sample_var) if cache_values else None
+        self._fit_final(Y=Y, T=T, X=X, W=W, Z=Z,
+                        nuisances=nuisances,
+                        sample_weight=sample_weight,
+                        sample_var=sample_var)
+
+        self._cache_invalid_message = None
+        return self
+
+    # TODO: this doesn't currently allow inference, because that is handled by the wrap_fit attribute
```
vsyrgkanis, line 592: So is inference the same object as the one called by fit?

I guess this is also related to the fact that the inference object does not have the new model final. I suspect this might need to be solved simultaneously (i.e. propagate model final to the inference object and allow refit to take as input "inference=...")

econml/_ortho_learner.py
```diff
@@ -540,15 +558,70 @@ def fit(self, Y, T, X=None, W=None, Z=None, *, sample_weight=None, sample_var=No
...
-                        Z=self._subinds_check_none(Z, fitted_inds),
-                        nuisances=tuple([self._subinds_check_none(nuis, fitted_inds) for nuis in nuisances]),
-                        sample_weight=self._subinds_check_none(sample_weight, fitted_inds),
-                        sample_var=self._subinds_check_none(sample_var, fitted_inds))
+
+        all_nuisances = []
+        fitted_inds = None
+
+        for _ in range(monte_carlo_iterations or 1):
+            nuisances, new_inds = self._fit_nuisances(Y, T, X, W, Z, sample_weight=sample_weight, groups=groups)
```
vsyrgkanis, line 566: If we fit multiple times, what version is models_nuisance storing? Is it storing for each mc iteration and for each fold in the kfold, the fitted model? Or is it storing only the final version of mc iterations? The latter would be inconsistent.

Similarly, when we call "score" we use the fitted models_nuisance to predict. But which fitted models_nusiance? just the last mc iter? or all the mc_iters?

vsyrgkanis (changes requested):

econml/_ortho_learner.py
kbattocchi: As discussed offline, I don't think so because the transformation happens at fit time.

econml/dml.py
kbattocchi: Agreed.

vsyrgkanis:
@kbattocchi to address #350, we might want to implement public access methods to the cached nuissances, e.g. est.get_nuisances at the ortho_learner level and est.get_residuals at the rlearner level. 

Currently this can be done by:
```python
Yres = est._cached_values.nuisances[0]
Tres = est._cached_values.nuisances[0]
```
which is cumbersome.

econml/_ortho_learner.py
vsyrgkanis: I found a related bug, not sure if related to the transformer. So the following works:
```python
est = LinearDML(model_y=RandomForestRegressor(),
                model_t=RandomForestClassifier(min_samples_leaf=10),
                discrete_treatment=False,
                linear_first_stages=False,
                n_splits=6)
est.fit(Y, T, X=X, W=W)
te_pred = est.effect(X_test)
te_pred[0]
```

But the following raises an error, while from the current semantics it's supposed to work fine:
```python
est = LinearDML(model_y=RandomForestRegressor(),
                model_t=RandomForestClassifier(min_samples_leaf=10),
                discrete_treatment=True,
                linear_first_stages=False,
                n_splits=6)
est.fit(Y, T, X=X, W=W)
est.discrete_treatment = False
est.fit(Y, T, X=X, W=W)
te_pred[0]
```

Which means that there is some state of the object that is maintained but should have been reset when resetting the discrete_treatment.

vsyrgkanis (changes requested):

econml/_ortho_learner.py
kbattocchi: Hm.  I wouldn't expect that to work since it uses a classifier for model_t but a continuous treatment.  I think changing discrete_treatment would require that either the model_t is 'auto' (although I need to fix this logic) or else that the model also changes.

econml/tests/test_dml.py
kbattocchi: Regarding checking fittedness, I agree that we should make sure that fit has been called before calling effect (and we don't always do this currently).  However, note that sklearn will not reset an estimator to "unfitted" if attributes are set in between calls to fit; for example:
1. fit an estimator
2. set an attribute
3. call predict
will work just fine, and generate predictions using the previously fit model (using non-current settings).  I'd expect our behavior to match.

econml/_ortho_learner.py
vsyrgkanis: Thats why I added the first example that when called a priori it works. The classifier can handle the discrete treatment. So there seems a problem with changing diwcrete treatment

kbattocchi: Turns out this was due to a typo in property name for the DML setter.  It will work after my next push.

econml/_ortho_learner.py
kbattocchi: You're right, it's not necessary.

econml/_ortho_learner.py
kbattocchi: Right now, it is fitting only a single model multiple times and retaining information from the latest run (because users might expect a single model per fold, not one per fold*mc_iter combination), but I can change that.

Perhaps best would be to have a model wrapper that wraps all of the individual ones but provides a single prediction, similar to bootstrap?

kbattocchi: Actually I take it back: `_crossfit` is cloning the model, so we don't call fit repeatedly on the same instance.

vsyrgkanis:
This code fails:
```python
from econml.sklearn_extensions.linear_model import StatsModelsLinearRegression, StatsModelsRLM, DebiasedLasso
from sklearn.linear_model import ElasticNetCV, LassoCV

est = DML(model_y=RandomForestRegressor(min_samples_leaf=20, max_depth=3),
          model_t=RandomForestRegressor(min_samples_leaf=20, max_depth=3),
          model_final=StatsModelsLinearRegression(fit_intercept=False),
          random_state=123)
est.fit(Y_train, T_train, X=X_train, W=W_train, cache_values=True)
print(list(zip(est.coef_, *est.coef__interval())))
est.model_final = StatsModelsRLM(fit_intercept=True)
est.refit()
print(list(zip(est.coef_, *est.coef__interval())))
```

```
[(6.45655731600806, 6.050960723017014, 6.8621539089991055)]
4.274931642986886
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-6-0a8e6a227739> in <module>
     13 est.model_final = StatsModelsRLM(fit_intercept=True)
     14 est.refit()
---> 15 print(list(zip(est.coef_, *est.coef__interval())))
     16 print(est.score_)
     17 

c:\users\vasy\documents\alicedev\econml\econml\cate_estimator.py in call(self, *args, **kwargs)
    205                 return getattr(self._inference, name)(*args, **kwargs)
    206             else:
--> 207                 raise AttributeError("Can't call '%s' because 'inference' is None" % name)
    208         return call
    209 

AttributeError: Can't call 'coef__interval' because 'inference' is None
```

econml/dml.py
```diff
@@ -847,17 +1069,59 @@ class KernelDML(DML):
...
+                             "instead set the bw and dim attributes to modify the kernel")
+
+    @DML.model_final.setter
+    def model_final(self, model):
+        raise AttributeError("KernelDML final model can't be changed from ElasticNetCV")
+
+    @_OrthoLearner.random_state.setter
+    def random_state(self, random_state):
+        # super().random_state = random_state
+        super(KernelDML, KernelDML).random_state.__set__(self, random_state)
```
vsyrgkanis, line 1121: Why are we still using this notation!

econml/dml.py
```diff
@@ -847,17 +1069,59 @@ class KernelDML(DML):
...
+
+    @property
+    def dim(self):
+        return self._dim
+
+    @dim.setter
+    def dim(self, dim):
+        self._dim = dim
+        # super().featurizer = _RandomFeatures(self._dim, self._bw, self.random_state)
+        super(KernelDML, KernelDML).featurizer.__set__(self, _RandomFeatures(self._dim, self._bw, self.random_state))
```
vsyrgkanis, line 1107: Why are we still using this notation!

econml/dml.py
```diff
@@ -847,17 +1069,59 @@ class KernelDML(DML):
...
+
+    @property
+    def bw(self):
+        return self._bw
+
+    @bw.setter
+    def bw(self, bw):
+        self._bw = bw
+        # super().featurizer = _RandomFeatures(self._dim, self._bw, self.random_state)
+        super(KernelDML, KernelDML).featurizer.__set__(self, _RandomFeatures(self._dim, self._bw, self.random_state))
```
vsyrgkanis, line 1097: Why are we still using this notation!

econml/dml.py
```diff
@@ -771,7 +935,65 @@ def fit(self, Y, T, X=None, W=None, *, sample_weight=None, sample_var=None, grou
...
+            alpha=self.alpha,
+            fit_intercept=False,
+            max_iter=self.max_iter,
+            tol=self.tol,
+            random_state=self.random_state)
+
+    @_OrthoLearner.random_state.setter
+    def random_state(self, random_state):
+        # super().random_state = random_state
+        super(SparseLinearDML, SparseLinearDML).random_state.__set__(self, random_state)
```
vsyrgkanis, line 990: Why are we still using this notation!

econml/dml.py
```diff
@@ -936,7 +1209,41 @@ def __init__(self,
...
+
+    @_BaseDML.featurizer.setter
+    def featurizer(self, featurizer):
+        self._featurizer = clone(featurizer, safe=False)
+        # featurizer isn't actually used by first stage models since linear_first_stages is False,
+        # so we only need to update model_final
+        # _BaseDML's final model setter reuses its old featurizer
+        # so we need to pass _BaseDML, not DML to the super() call
+        # super(_BasaeDML).model_final = _FinalWrapper(...)
+        super(_BaseDML, _BaseDML).model_final.__set__(self, _FinalWrapper(
```
vsyrgkanis, line 1223: This is a bug. There is not model_final

vsyrgkanis (changes requested):

econml/dml.py
```diff
@@ -252,7 +270,7 @@ def model_cate(self):
             An instance of the model_final object that was fitted after calling fit which corresponds
             to the constant marginal CATE model.
         """
-        return super().model_final._model
+        return self._model_final
```
vsyrgkanis, line 273: This should simply:
```
self.model_final
```
without the underscore

kbattocchi:
Shelving in favor of #360

PR #344, opened 2020-12-17
Native build proof of concept

kbattocchi:
This PR is not intendended to be merged, but to figure out how to get our build process working with the new native changes


kbattocchi:
@vsyrgkanis This seems to work on my machine in a clean venv... Now lets see if it'll work on each of our VM targets.

kbattocchi:
@vsyrgkanis The current set of changes is enough to get the tests running on every platform; there are still some more changes that need to be made but this should at least partially unblock you.

PR #361, opened 2021-01-03
RScorer class for causal model selection

vsyrgkanis:


kbattocchi (changes requested):

econml/score/ensemble_cate.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import numpy as np
+
+
+class EnsembleCateEstimator:
+
+    def __init__(self, *, cate_models, weights):
+        self.cate_models = cate_models
```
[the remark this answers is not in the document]
vsyrgkanis: We need to store the fitted models. Cloning has the certificate that it only copies the parameters. 

We could deep copy? What do you think?

econml/score/ensemble_cate.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import numpy as np
+
+
+class EnsembleCateEstimator:
```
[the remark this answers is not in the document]
vsyrgkanis: Ensemble estimator does not support inference.

econml/tests/test_rscorer.py
```diff
@@ -0,0 +1,70 @@
...
+
+        scorer = RScorer(model_y=reg(), model_t=clf(),
+                         discrete_treatment=True, n_splits=3, mc_iters=2, mc_agg='median')
+        scorer.fit(Y_val, T_val, X=X_val)
+        rscore = [scorer.score(mdl) for _, mdl in models]
+        expected_te_val = np.zeros(X_val.shape[0])
+        mdl, score = scorer.best_model([mdl for _, mdl in models])
+        rootpehe_best = np.sqrt(np.mean((expected_te_val.flatten() - mdl.effect(X_val).flatten())**2))
+        mdl, score = scorer.ensemble([mdl for _, mdl in models])
+        rootpehe_ensemble = np.sqrt(np.mean((expected_te_val.flatten() - mdl.effect(X_val).flatten())**2))
```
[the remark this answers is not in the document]
vsyrgkanis: not clear what to expect in these results unfort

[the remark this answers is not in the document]
vsyrgkanis: I can see if the score is positively correlated with the true rootpehe score !

econml/score/ensemble_cate.py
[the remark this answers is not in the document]
kbattocchi: Good point - just add a comment that we're explicitly not cloning because we need the fitted models.  Perhaps also add a note in the docstring that references to the live models are stored, so that e.g. refitting would affect the results of the methods on this class.

econml/score/ensemble_cate.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def effect(self, X=None, *, T0=0, T1=1):
+        return np.average([mdl.effect(X=X, T0=T0, T1=T1) for mdl in self.cate_models],
+                          weights=self.weights, axis=0)
+
+    def marginal_effect(self, T, X=None):
+        return np.average([mdl.marginal_effect(T, X=X) for mdl in self.cate_models],
+                          weights=self.weights, axis=0)
+
+    def const_marginal_effect(self, X=None):
```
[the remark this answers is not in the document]
vsyrgkanis: Done

econml/score/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
vsyrgkanis: Done

econml/score/rscorer.py
```diff
@@ -0,0 +1,213 @@
...
+
+        loss(cate) = E_n[(Yres - cate(X)'Tres)^2]
+
+    Also calculates a baseline loss based on a constant treatment effect model, i.e.::
+
+        base_loss = min_{theta} E_n[(Yres - theta'Tres)^2]
+
+    Returns an analogue of the R-square score for regression::
+
+        score = 1 - loss(cate) / base_loss
```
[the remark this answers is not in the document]
vsyrgkanis: These are actually coding blocks and not meant to be latexified. I find that prettier in the end than latex blocks in docstrings

econml/score/ensemble_cate.py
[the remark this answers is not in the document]
vsyrgkanis: Done

econml/score/rscorer.py
```diff
@@ -0,0 +1,213 @@
...
+
+    Parameters
+    ----------
+    model_y: estimator
+        The estimator for fitting the response to the features. Must implement
+        `fit` and `predict` methods.  Must be a linear model for correctness when linear_first_stages is ``True``.
+
+    model_t: estimator
+        The estimator for fitting the treatment to the features. Must implement
+        `fit` and `predict` methods.  Must be a linear model for correctness when linear_first_stages is ``True``.
```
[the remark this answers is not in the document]
vsyrgkanis: Done

econml/score/ensemble_cate.py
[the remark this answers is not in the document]
kbattocchi: ok

econml/score/rscorer.py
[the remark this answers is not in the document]
kbattocchi: Ok

econml/tests/test_rscorer.py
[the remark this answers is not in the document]
vsyrgkanis: Done

econml/score/ensemble_cate.py
```diff
@@ -0,0 +1,68 @@
...
+            raise ValueError('Parameter `cate_models` should be a list of `BaseCateEstimator` objects.')
+        self._cate_models = value
+
+    @property
+    def weights(self):
+        return self._weights
+
+    @weights.setter
+    def weights(self, value):
+        weights = check_array(value, accept_sparse=False, ensure_2d=False, allow_nd=False, dtype='numeric',
```
kbattocchi, line 64: Seems a little weird that this validation happens here, but not in the initializer.

kbattocchi (approved):

econml/score/rscorer.py
```diff
@@ -0,0 +1,219 @@
...
+        If None, the random number generator is the :class:`~numpy.random.mtrand.RandomState` instance used
+        by :mod:`np.random<numpy.random>`.
+    """
+
+    def __init__(self, *,
+                 model_y,
+                 model_t,
+                 discrete_treatment=False,
+                 categories='auto',
+                 n_splits=2,
```
heimengqi, line 88: Why not expose `featurizer` here? I guess the init argument here is used to fit a baseline DML and get the score_ as total variance, and to compare the performance of CATE over ATE, so the more flexible to get similar nuisances with cate model the better? I know for non-DML cate model it's not possible...

econml/score/rscorer.py
```diff
@@ -0,0 +1,219 @@
...
+                                    model_t=self.model_t,
+                                    n_splits=self.n_splits,
+                                    discrete_treatment=self.discrete_treatment,
+                                    categories=self.categories,
+                                    random_state=self.random_state,
+                                    mc_iters=self.mc_iters,
+                                    mc_agg=self.mc_agg)
+        self.lineardml_.fit(y, T, X=None, W=np.hstack([v for v in [X, W] if v is not None]),
+                            sample_weight=sample_weight, groups=groups, cache_values=True)
+        self.base_score_ = self.lineardml_.score_
```
heimengqi, line 138: seems the score function under `_ortho_learner` needs to be updated to incorporate Monte Carlo simulation?  since now we cached nuisances it's will be much easier

econml/score/rscorer.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        Y_res, T_res = self.lineardml_._cached_values.nuisances
+        X = self.lineardml_._cached_values.W[:, :self.dx_]
+        sample_weight = self.lineardml_._cached_values.sample_weight
+        if Y_res.ndim == 1:
+            Y_res = Y_res.reshape((-1, 1))
+        if T_res.ndim == 1:
+            T_res = T_res.reshape((-1, 1))
+        effects = cate_model.const_marginal_effect(X).reshape((-1, Y_res.shape[1], T_res.shape[1]))
+        Y_res_pred = np.einsum('ijk,ik->ij', effects, T_res).reshape(Y_res.shape)
```
heimengqi, line 161: By looking at your notebook example, so even for cate estimator doesn't follow the residual to residual final stage (e.g. metalearners) but has const_marginal_effect, we could still assume this theta(x)*res_t formula in order to compute r score?

notebooks/Causal Model Selection with the RScorer.ipynb
```diff
@@ -0,0 +1,645 @@
+{
```
heimengqi, line 1: I found couples of issues in the notebook, it's hard to point to the right place here, I will list one by one and point to the cell.

- cell 5: please comment out all the other exp_te functions and only leave one.
- cell 15: last four lines should be under the for loop, effect label should change to something like 'effect estimate'
- cell 17: effect label should change to something like 'effect estimate'
- cell 27-29: cell 27-28 returns nan, cell 29 last model should be best.effect(), and the xticks is also wrong

heimengqi (changes requested):
I left couple of comments, I think the major issue is the score function needs to be updated to incorporate monte carlo simulation?

econml/score/ensemble_cate.py
vsyrgkanis: Why? this is a check that should happen whenever we assign a value to the weights. The initializer simply invokes the setter.

econml/score/rscorer.py
vsyrgkanis: Here we are not using linear_first_stages=True. So the featurizer doesn't even enter the nusiance calculation when linear_first_sttes=False. The featurizer only affects the final model

econml/score/rscorer.py
vsyrgkanis: what do you mean?

heimengqi: I see here is fine since it's the training set score, and the score method is for a new set of data X, but still need to incorporate MC simulation?

econml/score/rscorer.py
vsyrgkanis: yeap. That's exactly right. If something claims to be a cate then it should achieve good residual on residual loss even if it's not trained based on that. Rscore is a univoersal scoring algorithm

econml/score/ensemble_cate.py
kbattocchi: Looks like the initializer is setting the field directly, not calling through this property.

kbattocchi: (and same for the models)

econml/score/rscorer.py
heimengqi: Okay I think I misunderstood the baseline model a bit when I left the comment. Now I think how scorer module works is by fitting a baseline model on the validation set of data, and use the MSE of training set of baseline model (which is the test set of cate model for comparison) as total variance, and use nuisances from baseline model plus the cate learnt from target model to get MSE of cate model. Do you know any reference I could take a look? especially on why we could use theta(x) learnt from model cate multiply residual t learnt from baseline model to get prediction residual y?

heimengqi: I see intuitively could I think about res_y from baseline model as truth response variable on test set, and res_t from baseline as truth independent variable, and we want to learn model cate maximum r score for a given test set X.

econml/score/rscorer.py
vsyrgkanis: I don;t think we need this here. But in general we could enable the mc simulation in score of _ortho_learner, but that would require to store all iterations. This seems to be related to the TODO in ortholearner where we say that we need to potentially store all nuisance estimates from all mc iters. Currently we only store the final one. A new PR which fixes that could also fix the scoring to use all the mc iters.

heimengqi: Yes exactly!

econml/score/rscorer.py
vsyrgkanis: https://arxiv.org/pdf/1804.05146.pdf

vsyrgkanis: We are implementin the tau_risk_R in Equation (9)

econml/score/rscorer.py
kbattocchi: Wait, the stored nuisances are the aggregated ones, not the final ones; it's just the model where we store only the final one (but maybe that's what you meant)

econml/score/rscorer.py
vsyrgkanis: I added a ref in the docstring

heimengqi: Thanks!

econml/score/ensemble_cate.py
vsyrgkanis: I think that's done implicit. Exactly because there is a setter it automatically invokes the setter.

vsyrgkanis:
```python
class Test:
    def __init__(self, value):
        self.value = 10
    
    @property
    def value(self):
        return self._value
    
    @value.setter
    def value(self, value):
        print("Test")
        self._value = value
```

Then
```
Test(10)
```
prints "Test"

notebooks/Causal Model Selection with the RScorer.ipynb
vsyrgkanis: Done

heimengqi (approved):
Looks good!

econml/score/ensemble_cate.py
kbattocchi: Try:

```
class Test:
    def __init__(self, value):
        self._value = value # leading underscore

    @property
    def value(self):
        return self._value

    @value.setter
    def value(self, value):
        print("Test")
        self._value = value

kbattocchi: (Maybe we're agreeing and I should have put my comment in the initializer instead - it needs to be setting `self.weight` and `self.cate_models`, not `self._weight` and `self._cate_models`)

vsyrgkanis: yeap! Now we're good... :) 

Fixed

PR #363, opened 2021-01-04
Enabling summary() even when inference not available

vsyrgkanis:
Enabled construction of inference results even when stderr is None. So that we can still print pretty summaries with only point estimates.


econml/inference.py
```diff
@@ -118,17 +118,21 @@ def const_marginal_effect_inference(self, X):
...
         X, T = broadcast_unit_treatments(X, self.d_t)
         pred = reshape_treatmentwise_effects(self._predict(cross_product(X, T)), self._d_t, self._d_y)
-        if not hasattr(self.model_final, 'prediction_stderr'):
-            raise AttributeError("Final model doesn't support prediction standard eror, "
-                                 "please call const_marginal_effect_interval to get confidence interval.")
-        pred_stderr = reshape_treatmentwise_effects(self._prediction_stderr(cross_product(X, T)), self._d_t, self._d_y)
+        pred_stderr = None
+        if hasattr(self.model_final, 'prediction_stderr'):
+            pred_stderr = reshape_treatmentwise_effects(self._prediction_stderr(cross_product(X, T)),
+                                                        self._d_t, self._d_y)
```
kbattocchi, line 124: Since _prediction_stderr already has the check, maybe

```suggestion
        pred_stderr = self._prediction_stderr(cross_product(X, T))
        if pred_stderr is not None:
            pred_stderr = reshape_treatmentwise_effects(pred_stderr,
                                                        self._d_t, self._d_y)
```

unless we don't want to see the warning in this case.

econml/inference.py
```diff
@@ -351,14 +355,17 @@ def const_marginal_effect_interval(self, X, *, alpha=0.1):
...
-        pred = np.array([mdl.predict(X) for mdl in self.fitted_models_final])
-        if not hasattr(self.fitted_models_final[0], 'prediction_stderr'):
-            raise AttributeError("Final model doesn't support prediction standard eror, "
-                                 "please call const_marginal_effect_interval to get confidence interval.")
-        pred_stderr = np.array([mdl.prediction_stderr(X) for mdl in self.fitted_models_final])
-        return NormalInferenceResults(d_t=self.d_t, d_y=self.d_y, pred=np.moveaxis(pred, 0, -1),
-                                      # send treatment to the end, pull bounds to the front
-                                      pred_stderr=np.moveaxis(pred_stderr, 0, -1), inf_type='effect',
+        pred = np.moveaxis(np.array([mdl.predict(X) for mdl in self.fitted_models_final]), 0, -1)
+        if hasattr(self.fitted_models_final[0], 'prediction_stderr'):
```
kbattocchi, line 359: I guess it's safe to assume that if the first model has this attribute, they all do?

kbattocchi (approved):

econml/inference.py
vsyrgkanis: I just added the warning message there too. I didn't want to perform the crossproduct only to then return None.

econml/inference.py
vsyrgkanis: yeah this is used in the case where a single same model is trained for separate targets

PR #365, opened 2021-01-05
Averate Treatment Effect methods to all estimators

vsyrgkanis:
Added ate and ate inference related methods to all estimators and changed cate interpreter to use ate inference to produce confidence intervals, which internally uses population summary as default.


econml/inference.py
```diff
@@ -1032,8 +1050,7 @@ def stderr_mean(self):
         """
         return np.sqrt(np.mean(self.pred_stderr**2, axis=0))
 
-    @property
-    def zstat(self):
+    def zstat(self, *, value=None):
         """
         Get the z statistic of the mean point estimate of each treatment on each outcome for sample X.
```
kbattocchi, line 1056: Add parameter documentation for `value`

econml/inference.py
```diff
@@ -1045,11 +1062,11 @@ def zstat(self):
...
-        zstat = (self.mean_point - self.value) / self.stderr_mean
+        value = self.value if value is None else value
+        zstat = (self.mean_point - value) / self.stderr_mean
         return zstat
 
-    @property
-    def pvalue(self):
+    def pvalue(self, *, value=None):
         """
         Get the p value of the z test of each treatment on each outcome for sample X.
```
kbattocchi, line 1072: Add parameter documentation for `value`

econml/inference.py
```diff
@@ -1061,11 +1078,10 @@ def pvalue(self):
...
         """
-        pvalue = norm.sf(np.abs(self.zstat), loc=0, scale=1) * 2
+        pvalue = norm.sf(np.abs(self.zstat(value=value)), loc=0, scale=1) * 2
         return pvalue
 
-    @property
-    def conf_int_mean(self):
+    def conf_int_mean(self, *, alpha=None):
         """
         Get the confidence interval of the mean point estimate of each treatment on each outcome for sample X.
```
kbattocchi, line 1087: Add parameter documentation for `alpha`

econml/inference.py
```diff
@@ -1101,8 +1120,7 @@ def std_point(self):
         """
         return np.std(self.pred, axis=0)
 
-    @property
-    def percentile_point(self):
+    def percentile_point(self, *, alpha=None):
         """
         Get the confidence interval of the point estimate of each treatment on each outcome for sample X.
```
kbattocchi, line 1126: Add parameter documentation for alpha

econml/inference.py
```diff
@@ -1114,10 +1132,10 @@ def percentile_point(self):
...
             (e.g. if both are vectors, then the output of this method will also be a vector)
         """
-        lower_percentile_point = np.percentile(self.pred, (self.alpha / 2) * 100, axis=0)
-        upper_percentile_point = np.percentile(self.pred, (1 - self.alpha / 2) * 100, axis=0)
-        return np.array([lower_percentile_point]) if np.isscalar(lower_percentile_point) else lower_percentile_point, \
-            np.array([upper_percentile_point]) if np.isscalar(upper_percentile_point) else upper_percentile_point
+        alpha = self.alpha if alpha is None else alpha
+        lower_percentile_point = np.percentile(self.pred, (alpha / 2) * 100, axis=0)
+        upper_percentile_point = np.percentile(self.pred, (1 - alpha / 2) * 100, axis=0)
+        return lower_percentile_point, upper_percentile_point
```
kbattocchi, line 1138: Is it correct that we no longer wrap in an array if the outputs are scalar?

econml/inference.py
```diff
@@ -1134,8 +1152,7 @@ def stderr_point(self):
         """
         return np.sqrt(self.stderr_mean**2 + self.std_point**2)
 
-    @property
-    def conf_int_point(self):
+    def conf_int_point(self, *, alpha=None, tol=None):
         """
         Get the confidence interval of the point estimate of each treatment on each outcome for sample X.
```
kbattocchi, line 1158: Add parameter documentation for alpha, tol

econml/inference.py
```diff
@@ -1147,12 +1164,17 @@ def conf_int_point(self):
...
-        return np.array([lower_ci_point]) if np.isscalar(lower_ci_point) else lower_ci_point,\
-            np.array([upper_ci_point]) if np.isscalar(upper_ci_point) else upper_ci_point
+        alpha = self.alpha if alpha is None else alpha
+        tol = self.tol if tol is None else tol
+        lower_ci_point = np.array([self._mixture_ppf(alpha / 2, self.pred, self.pred_stderr, tol)])
+        upper_ci_point = np.array([self._mixture_ppf(1 - alpha / 2, self.pred, self.pred_stderr, tol)])
+        return lower_ci_point, upper_ci_point
 
-    def print(self):
+    def summary(self, alpha=0.1, value=0, decimals=3, tol=0.001, output_names=None, treatment_names=None):
```
kbattocchi, line 1173: Why do we need both `summary` and `print`, which are the same except for their default values?

If this method is really needed, please add a docstring.

econml/inference.py
```diff
@@ -1147,12 +1164,17 @@ def conf_int_point(self):
...
+        return self.print(alpha=alpha, value=value, decimals=decimals,
+                          tol=tol, output_names=output_names, treatment_names=treatment_names)
+
+    def print(self, *, alpha=None, value=None, decimals=None, tol=None, output_names=None, treatment_names=None):
         """
         Output the summary inferences above.
```
kbattocchi, line 1180: Add parameter documentation for all parameters.

kbattocchi (changes requested):
Mostly looks good, needs a few updates to docs.

econml/inference.py
vsyrgkanis: Yeap. was that an api we wanted to abid by?

econml/inference.py
vsyrgkanis: good point, maybe we can merge. My thought: print is an internal api method and summary public facing. Maybe if I change print to `_print` it might be better.

econml/inference.py
vsyrgkanis: Done

econml/inference.py
vsyrgkanis: Done

econml/inference.py
vsyrgkanis: Done

econml/inference.py
vsyrgkanis: Done

econml/inference.py
vsyrgkanis: I'm keeping it without a numpy array. Hope that's ok. I liked it this way as it's easier to use in the cate interpreters too. Also I would epxect that someone that has a single treatment and wants to get an ATE is expecting a scalar number.

econml/inference.py
vsyrgkanis: Done

econml/inference.py
vsyrgkanis: I moved it to a private helper function. We do need two functions, vecause we want to use them with different defaults

econml/inference.py
vsyrgkanis: Done

kbattocchi (approved):

PR #369, opened 2021-01-14
Add doc for interpretability

heimengqi:


doc/spec/interpretability.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Our package offers multiple interpretability tools to better understand the final model CATE.
+
+
+Tree Interpreter
+----------------
+
+Tree Interpreter provides a presentation-ready summary of the key features that explain the biggest differences in responsiveness to an intervention.
+
+:class:`.SingleTreeCateInterpreter` trains a single shallow decision tree for the treatment effect :math:`\theta(X)` you learnt from any of
```
kbattocchi, line 12: In the US, "learned" is almost always used instead of "learnt"

```suggestion
:class:`.SingleTreeCateInterpreter` trains a single shallow decision tree for the treatment effect :math:`\theta(X)` you learned from any of
```

doc/spec/interpretability.rst
```diff
@@ -0,0 +1,95 @@
...
+our available CATE estimators on a small set of feature :math:`X` that you are interested to learn heterogeneity from. The model will split on the cutoff
+points that maximize the treatment effect difference in each leaf. Finally each leaf will be a subgroup of customers respond to a treatment differently
```
kbattocchi, line 14:
```suggestion
points that maximize the treatment effect difference in each leaf. Finally each leaf will be a subgroup of customers that respond to a treatment differently
```

doc/spec/interpretability.rst
```diff
@@ -0,0 +1,95 @@
...
+with other leaves.
```
kbattocchi, line 15:
```suggestion
from other leaves. 
```

doc/spec/interpretability.rst
```diff
@@ -0,0 +1,95 @@
...
+    intrp.interpret(est, X)
+    # Plot the tree
+    intrp.plot(feature_names=['A', 'B', 'C'], fontsize=12)
+
+Policy Interpreter
+------------------
+Policy Interpreter offers similar functionality but taking cost into consideration. 
+
+Instead of training a tree model regressor on :math:`\theta(X)`, :class:`.SingleTreePolicyInterpreter` trains a tree model classifier by using whether the effect is above the cost as
+label. Finally we could make simple rules about what kind of customers we should target on in order to maximum the outcome of interest.
```
kbattocchi, line 47:
```suggestion
label. This results in simple rules to determine what kind of customers we should target on in order to maximize the outcome of interest.
```

doc/spec/interpretability.rst
kbattocchi, line 14: Do we want to phrase this in terms of customers, or try to make it more generic?

kbattocchi (approved):
Looks good.  I've suggested a few minor grammar corrections.

doc/spec/interpretability.rst
heimengqi: Sure I re-phrase this to be more generic

doc/spec/interpretability.rst
```diff
@@ -0,0 +1,95 @@
...
+    est.fit(y, t, X=X, W=W)
+    intrp = SingleTreeCateInterpreter(include_model_uncertainty=True, max_depth=2, min_samples_leaf=10)
+    # We interpret the CATE model's behavior based on the features used for heterogeneity
+    intrp.interpret(est, X)
+    # Plot the tree
+    intrp.plot(feature_names=['A', 'B', 'C'], fontsize=12)
+
+Policy Interpreter
+------------------
+Policy Interpreter offers similar functionality but taking cost into consideration.
```
vsyrgkanis, line 44: Maybe change the wording here. The policy interpreter, instead of fitting a tree to learn groups that have a different treatment effect, it tries to split the samples into different treatment groups. So in the case of binary treatments it tries to create sub-groups such that all samples within the group have either all positive effect or all negative effect. Thus it tries to separate responders from non-responders, as opposed to trying to find groups that have different levels of response. 

This way you can construct an interpretable personalized policy where you treat the groups with a postive effect and don't treat the group with a negative effect. Our policy tree provides the recommended treatment at each leaf node.

vsyrgkanis (approved):

doc/spec/interpretability.rst
heimengqi: Done!

PR #370, opened 2021-01-14
Vasilis/docs

vsyrgkanis:
changing the doc structure


econml/sklearn_extensions/linear_model.py
```diff
@@ -294,6 +305,17 @@ class WeightedMultiTaskLasso(WeightedModelMixin, MultiTaskLasso):
...
+    .. testcode::
+        :hide:
+
+        # Our classes that derive from sklearn ones sometimes include
+        # inherited docstrings that have embedded doctests; we need the following imports
+        # so that they don't break.
+
+        import numpy as np
+        from sklearn.linear_model import lasso_path
+
```
kbattocchi, line 318: I think these should only need to be imported once.
```suggestion
```

econml/sklearn_extensions/linear_model.py
```diff
@@ -326,6 +348,16 @@ def fit(self, X, y, sample_weight=None):
...
```
kbattocchi, line 360: Same here.
```suggestion
```

econml/sklearn_extensions/linear_model.py
```diff
@@ -438,6 +471,16 @@ def fit(self, X, y, sample_weight=None):
...
```
kbattocchi, line 483: And here.
```suggestion
```

econml/sklearn_extensions/linear_model.py
```diff
@@ -575,6 +619,16 @@ class DebiasedLasso(WeightedLasso):
...
```
kbattocchi, line 631: And here
```suggestion
```

econml/iv/tsls/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+from ._tsls import HermiteFeatures, DPolynomialFeatures, NonParamTSLS
```
kbattocchi, line 4: Is TSLS the right name for this module?  What if we ever want to add other GMM estimators, for instance (e.g. since 2SLS is apparently not asymptotically efficient, though maybe we don't care)?

kbattocchi (changes requested):
Shouldn't there be many more notebook changes to move everything to the new names?  Also found a few minor issues.

econml/bootstrap.py
```diff
@@ -1,281 +1,17 @@
...
+
+@deprecated("The econml.bootstrap.BootstrapEstimator class has been moved "
+            "to econml.inference._bootstrap.BootstrapEstimator and is no longer part of the public API; "
+            "an upcoming release will remove support for the old name and will consider `BootstrapEstimator` "
+            "as part of the private API with no guarantee of API consistency across releases. "
+            "Instead of wrapping CATE esitmators with the BootstrapEstimator to preduce bootstrap confidence "
+            "intervals, consider passing `inference='bootstrap'` or "
+            "`inference=econml.inference.BootstrapInference(n_bootstrap_samples=..., bootstrap_type=...)`, "
+            "as a keyword argument at the `fit` method of the CATE estimator.")
+class BootstrapEstimator(bootstrap.BootstrapEstimator):
```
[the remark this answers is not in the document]
vsyrgkanis: We'll still maintain the estimator in the code. But I think it's important for the users of the econml library to know that this is not the way to use the estimator with our cate estimates and that they shouldn't be using this class directly (we already had some users face this). So I do think we should make it a private class. One can still access it as `econml.inference._bootstrap.BootstrapEstimator`

econml/deepiv.py
```diff
@@ -1,461 +1,17 @@
...
+@deprecated("The econml.deepiv.DeepIVEstimator class has moved to econml.iv.nnet.DeepIV; "
+            "an upcoming release will remove support for the old name")
+class DeepIVEstimator(nnet.DeepIV):
+    pass
+
+
+@deprecated("The econml.deepiv.DeepIVEstimator class has been renamed to econml.iv.nnet.DeepIV; "
+            "an upcoming release will remove support for the old name")
+class DeepIVEstimator(nnet.DeepIV):
+    pass
```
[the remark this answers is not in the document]
vsyrgkanis: Done

econml/drlearner.py
```diff
@@ -1,1537 +1,29 @@
...
+
+@deprecated("The econml.drlearner.DRLearner class has been moved to econml.dr.DRLearner; "
+            "an upcoming release will remove support for the old name")
+class DRLearner(dr.DRLearner):
+    pass
+
+
+@deprecated("The econml.drlearner.LinearDRLearner class has been moved to econml.dr.LinearDRLearner; "
+            "an upcoming release will remove support for the old name")
+class DRLearner(dr.DRLearner):
```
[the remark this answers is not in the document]
vsyrgkanis: Done

econml/drlearner.py
```diff
@@ -1,1537 +1,29 @@
...
+    pass
+
+
+@deprecated("The econml.drlearner.SparseLinearDRLearner class has been moved to econml.dr.SparseLinearDRLearner; "
+            "an upcoming release will remove support for the old name")
+class DRLearner(dr.DRLearner):
```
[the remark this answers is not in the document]
vsyrgkanis: Done

econml/drlearner.py
```diff
@@ -1,1537 +1,29 @@
...
+    pass
+
+
+@deprecated("The econml.drlearner.ForestDRLearner class has been moved to econml.dr.ForestDRLearner; "
+            "an upcoming release will remove support for the old name")
+class DRLearner(dr.DRLearner):
```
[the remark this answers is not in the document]
vsyrgkanis: Done

econml/sklearn_extensions/linear_model.py
```diff
@@ -199,6 +199,17 @@ class WeightedLasso(WeightedModelMixin, Lasso):
     n_iter_ : int | array-like, shape (n_targets,)
         number of iterations run by the coordinate descent solver to reach
         the specified tolerance.
+
+    .. testcode::
+        :hide:
```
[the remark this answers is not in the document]
vsyrgkanis: Done

econml/sklearn_extensions/linear_model.py
vsyrgkanis: Unforutnately  now that each class is in a separate autosummary page, we need this to be on each page. We can no longer just put it at the top of the file for instance.

econml/sklearn_extensions/linear_model.py
vsyrgkanis: same as above

econml/sklearn_extensions/linear_model.py
vsyrgkanis: same as above

econml/sklearn_extensions/linear_model.py
vsyrgkanis: same as above

econml/iv/tsls/__init__.py
vsyrgkanis: I think maybe the right name is npiv, as we are really doing non-parametric IV by expanding the treatments to some basis. In fact the method we actually implement is typically referred to as Sieve 2SLS, as we use sieves to estimate the non-parametric IV method. 

What do you think of: `econml.iv.npiv.SieveTSLS`

Or alternatively, we could do: `econml.npiv.SieveTSLS`
But then we would also need to move `DeepIV` there too, as: `econml.npiv.DeepIV`

But the latter one might be weird, as then why isn't `dmliv` and `driv`, just top level modules, e.g.: `econml.dmliv.DMLATEIV` and `econml.driv.IntentToTreatDRIV`

I kind of liked that anything IV related is under the `econml.iv` module. So maybe just the first proposal of moving tsls to `econml.iv.npiv.SieveTSLS` is ok.

Alternatively, we could do: `econml.iv.sieve.SieveTSLS`

vsyrgkanis:
> Shouldn't there be many more notebook changes to move everything to the new names? Also found a few minor issues.

Most of the actual import paths did not change because the name of the folder is rhe name of the previous file and all public classes of each folder are included in the init

econml/iv/tsls/__init__.py
vsyrgkanis: I went with `econml.iv.sieve.SieveTSLS` seems best suited for what the method is, from the academic lit perspective and also better inline with our naming conventions

econml/tests/test_bootstrap.py
```diff
@@ -292,7 +292,7 @@ def test_stratify_orthoiv(self):
         X = np.array([1, 1, 2, 2, 1, 2, 1, 2]).reshape(-1, 1)
         est = LinearIntentToTreatDRIV(model_Y_X=LinearRegression(), model_T_XZ=LogisticRegression(),
                                       flexible_model_effect=LinearRegression(), cv=2)
-        inference = BootstrapInference(n_bootstrap_samples=20)
+        inference = BootstrapInference(n_bootstrap_samples=20, n_jobs=-1, verbose=3)
```
kbattocchi, line 295: Should we be testing that the output is more verbose in this case?

kbattocchi (approved):

PR #373, opened 2021-01-15
Fix automl handling of positional initializer args

### LINKED ISSUE #371: Bug in automated_ml integration
There seems to be a bug here:
https://github.com/microsoft/EconML/blob/35c5418618d4ca9828f5465c090dd17e5e9a263c/econml/automated_ml.py#L292

Maybe also need to add a test where an `automl_config` is passed as positional argument so as to capture this path. Currently this seems to be the reason why this bug is not captured by tests.

kbattocchi:
Fixes #371.

Also fixes a minor drlearner issue.

### DIFF
--- a/econml/automated_ml.py
+++ b/econml/automated_ml.py
@@ -41,11 +41,10 @@ SAMPLE_WEIGHTS_MODELS_SET = set([
 
 def setAutomatedMLWorkspace(create_workspace=False,
                             create_resource_group=False, workspace_region=None, *,
-                            subscription_id=None, resource_group=None, workspace_name=None, auth=None):
+                            auth=None, subscription_id, resource_group, workspace_name):
     """Set configuration file for AutomatedML actions with the EconML library. If
     ``create_workspace`` is set true, a new workspace is created
-    for the user. If ``create_workspace`` is set true, a new workspace is
-    created for the user.
+    for the user.
 
     Parameters
     ----------
@@ -68,8 +67,7 @@ def setAutomatedMLWorkspace(create_workspace=False,
         authentication portal in the browser.
 
     subscription_id: String, required
-       Definition of a class that will serve as the parent class of the
-       AutomatedMLMixin. This class must inherit from _BaseDML.
+       Azure subscription ID for the subscription under which to run the models
 
     resource_group: String, required
        Name of resource group of workspace to be created or set.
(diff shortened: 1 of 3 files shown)


econml/tests/test_automated_ml.py
```diff
@@ -45,10 +45,10 @@
     AutomatedKernelDML = addAutomatedML(KernelDML)
     AutomatedNonParamDML = \
         addAutomatedML(NonParamDML)
-    AutomatedForestDML = addAutomatedML(ForestDML)
+    AutomatedCausalForestDML = addAutomatedML(CausalForestDML)
 
     AUTOML_SETTINGS_REG = {
-        'experiment_timeout_minutes': 1,
+        'experiment_timeout_minutes': 15,
```
kbattocchi, line 51: 15 minutes is the new lowest allowed value for the experiment timeout.

vsyrgkanis: Wow, that’s a lot

econml/automated_ml.py
```diff
@@ -285,12 +283,12 @@ def __init__(self, *args, **kwargs):
...
         new_args = ()
-        for var in args:
+        for idx, arg in enumerate(args):
             # If item is an automl config, get its corresponding
             # AutomatedML Model and add it to new_Args
-            if isinstance(var, EconAutoMLConfig):
-                var = self._get_automated_ml_model(kwarg, key)
-            new_args += (var,)
+            if isinstance(arg, EconAutoMLConfig):
+                arg = self._get_automated_ml_model(arg, f"arg{idx}")
```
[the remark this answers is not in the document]
vsyrgkanis: and what if "arg0" is also a keyword argument?

econml/tests/test_automated_ml.py
```diff
@@ -154,28 +152,21 @@ def test_param(self):
...
-        Y, T, X, _ = ihdp_surface_B()
-        est = AutomatedForestDML(model_y=automl_model_reg(),
-                                 model_t=GradientBoostingClassifier(),
-                                 discrete_treatment=True,
-                                 n_estimators=1000,
-                                 subsample_fr=.8,
-                                 min_samples_leaf=10,
-                                 min_impurity_decrease=0.001,
-                                 verbose=0, min_weight_fraction_leaf=.01)
+        est = AutomatedCausalForestDML(model_y=automl_model_reg(),
```
vsyrgkanis, line 155: where did the data generation go?

vsyrgkanis (approved):

econml/automated_ml.py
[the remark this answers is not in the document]
kbattocchi: I believe it is just used to make it easier for a person to identify the experiment when looking through experiments in AzureML studio, and that the service itself adds a unique suffix so that the whole thing will be unambiguous in any case.

econml/tests/test_automated_ml.py
kbattocchi: We were already performing the data generation in the big `try` block up top, and we made use of this fact in one test but not the others, which I found confusing.

PR #389, opened 2021-01-28
Relax tensorflow  version limit

kbattocchi:


akelleh:
@kbattocchi Looks like this might fix the DeepIV implementation! need a hand pushing it over the finish line?

akelleh:
I'll make some time this weekend, if you don't get to it before me!

akelleh:
> @akelleh Yes, that would be great! Thanks for the offer.

Having some trouble even getting tests running... Following the build instructions on the main page in a new virtualenv (`python setup.py develop` with an updated setuptools), `python setup.py pytest` just returns `error: invalid command 'pytest'`. Any trick to make this work?

kbattocchi:
Whoops, sorry - looks like we haven't updated the docs after restructuring our package a bit.  Try:

    pip install -e .[tf,plt]
    pip install pytest-runner
    python setup.py pytest

kbattocchi:
You should also set the environment variable PYTEST_ADDOPTS to `-m "not (notebook or automl or dml)"` so that you're not running a bunch of tests that are unimpacted (and in the case of the notebook and automl tests, have additional dependencies).

akelleh:
> You should also set the environment variable PYTEST_ADDOPTS to `-m "not (notebook or automl or dml)"` so that you're not running a bunch of tests that are unimpacted (and in the case of the notebook and automl tests, have additional dependencies).

Thanks! Will do. I'm taking your PR for a spin now -- so far so good!

PR #398, opened 2021-02-03
Fixed column input names bugs.

### LINKED ISSUE #375: Bug in default treatment name parsing from pandas Dataframe
When treatment is discrete and the user passes a pandas dataframe, then we store the wrong default `treatment_names`.  e.g.

```python
from econml.dml import LinearDML
import pandas as pd
est = LinearDML(discrete_treatment=True, linear_first_stages=False)
est.fit(y, pd.DataFrame(T, columns=['treatment']), X=X)
est.summary()
```

We should not be storing the raw column name, but maybe the column names of the output of `pd.get_dummies()`, or maybe just the raw classes of the treatment or even better the outcome of `get_feature_names` of the sklearn one-hot encoder, applied to the column name of the pandas data frame.

Given how to create the expanded feature names can be up to the transformer of the `TreatmentExpansionMixin`, most problably we need to create an abstract method in the `TreatmentExpansionMixin` that has to be implemented by all children, and which produces the transformed treatment names, given the base name. One problem I foresee is that the transformer is created at fit time, while the input names are constructed at `_prefit` time. So the `transformer` object might not even be available.

### LINKED ISSUE #376: Bug in default variables when refitting
This line is problematic:
https://github.com/microsoft/EconML/blob/69fadc3951b94bb2b001144b973e59e5d109cfaa/econml/_cate_estimator.py#L79

If we fit the same estimator twice, but with different variable names, then because the first call to set inputs, would set the flags, then the second fit call would not re-set the names, which is wrong. e.g.
```python
from econml.dml import LinearDML
import pandas as pd
T = np.random.binomial(2, .5, size=(1000, 2))
est = LinearDML(discrete_treatment=False, linear_first_stages=False)
est.fit(y, pd.DataFrame(T, columns=['T1', 'T2']), W=X)
est.summary()
est.fit(y, pd.DataFrame(T, columns=['A1', 'A2']), W=X)
est.summary()
```
Still produces summary results with T1, T2 etc and not A1, A2 etc.

We should just remove this check. The convention is that only the cate estimator sets the input names and no child. So we can just remove this flag alltogether. I don't see a reason why it should exist.

moprescu:
* Fixes #375, fixes #376 
* Added tests


econml/_cate_estimator.py
```diff
@@ -44,18 +44,24 @@ def _get_inference(self, inference):
...
         }
         if set_flag:
             # This flag is true when names are set in a child class instead
             # If names are set in a child class, add an attribute reflecting that
             self._input_names_set = True
 
+    def _set_encoded_treatment_names(self, one_hot_encoder):
+        """Works with sklearn OHEs"""
+        if hasattr(self, "_input_names"):
+            self._input_names["treatment_names"] = one_hot_encoder.get_feature_names(
```
kbattocchi, line 62: Do we need to ensure that this isn't called more than once?

econml/utilities.py
```diff
@@ -570,7 +573,8 @@ def get_input_columns(X):
     }
     if type(X) in type_to_func:
         return type_to_func[type(X)](X)
-    return None
+    len_X = 1 if np.ndim(X) == 1 else np.asarray(X).shape[1]
+    return ["{}{}".format(prefix, i) for i in range(len_X)]
```
kbattocchi, line 577: I think a format string literal is clearer:
```suggestion
    return [f"{prefix}{i}" for i in range(len_X)]
```

econml/_cate_estimator.py
kbattocchi: For instance, what happens if fit is called multiple times in the corresponding classes?

kbattocchi (changes requested):
Mostly looks good; just a few questions

econml/_cate_estimator.py
```diff
@@ -1190,7 +1242,6 @@ def summary(self, T, *, alpha=0.1, value=0, decimals=3,
         try:
             coef_table = self.coef__inference(T).summary_frame(
                 alpha=alpha, value=value, decimals=decimals, feature_names=feature_names,
-                treatment_names=treatment_names,
```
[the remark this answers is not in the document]
kbattocchi: And if we really don't need them for some reason, I think you can get rid of the line above that generates them.

econml/_cate_estimator.py
```diff
@@ -1202,7 +1253,6 @@ def summary(self, T, *, alpha=0.1, value=0, decimals=3,
         try:
             intercept_table = self.intercept__inference(T).summary_frame(
                 alpha=alpha, value=value, decimals=decimals, feature_names=None,
-                treatment_names=treatment_names,
```
kbattocchi, line 1199: Same question here.

econml/inference/_inference.py
```diff
@@ -192,8 +192,11 @@ def effect_inference(self, X, *, T0, T1):
         e_stderr = np.einsum(einsum_str, cme_stderr, np.abs(dT)) if cme_stderr is not None else None
         d_y = self._d_y[0] if self._d_y else 1
         # d_t=1 here since we measure the effect across all Ts
+        # A dummy treatment names is passed in
```
kbattocchi, line 195: Consider rewording, since it is not actually passed in

```suggestion
        # Omit treatment name argument to fall back to the default
```

econml/inference/_inference.py
```diff
@@ -240,8 +243,11 @@ def effect_inference(self, X, *, T0, T1):
         e_stderr = self._prediction_stderr(cross_product(X, T1 - T0))
         d_y = self._d_y[0] if self._d_y else 1
         # d_t=1 here since we measure the effect across all Ts
+        # A dummy treatment names is passed in
```
kbattocchi, line 246: Consider rewording, since it is not actually passed in

```suggestion
        # Omit treatment name argument to fall back to the default
```

econml/inference/_inference.py
```diff
@@ -425,9 +431,12 @@ def effect_inference(self, X, *, T0, T1):
         pred = pred[np.arange(T0.shape[0]), ..., ind]
         pred_stderr = pred_stderr[np.arange(T0.shape[0]), ..., ind] if pred_stderr is not None else None
         # d_t=1 here since we measure the effect across all Ts
+        # A dummy treatment names is passed in
```
kbattocchi, line 434: Consider rewording, since it is not actually passed in

```suggestion
        # Omit treatment name argument to fall back to the default
```

econml/inference/_inference.py
```diff
@@ -462,9 +471,12 @@ def fname_transformer(x):
         else:
             def fname_transformer(x):
                 return x
+        # d_t=1 here since we measure the effect across all Ts
+        # A dummy treatment names is passed in
```
kbattocchi, line 475: Consider rewording, since it is not actually passed in

```suggestion
        # Omit treatment name argument to fall back to the default
```

econml/inference/_inference.py
```diff
@@ -486,9 +498,13 @@ def intercept__inference(self, T):
             warn("Final model doesn't have a `intercept_stderr_` attribute. "
                  "Only point estimates will be available.")
             intercept_stderr = None
+        # d_t=1 here since we measure the effect across all Ts
+        # A dummy treatment names is passed in
```
kbattocchi, line 502: Consider rewording, since it is not actually passed in

```suggestion
        # Omit treatment name argument to fall back to the default
```

econml/orf/_ortho_forest.py
```diff
@@ -1300,8 +1300,12 @@ def effect_inference(self, X=None, *, T0=0, T1=1):
             a dataframe summary of these inference results.
         """
         eff, scales = self._effect_inference_helper(X, T0, T1)
+        # d_t=1 here since we measure the effect across all Ts
+        # A dummy treatment names is passed in
```
kbattocchi, line 1304: Consider rewording, since it is not actually passed in

```suggestion
        # Omit treatment name argument to fall back to the default
```

econml/utilities.py
```diff
@@ -570,7 +570,8 @@ def get_input_columns(X, prefix="X"):
         return None
     if np.ndim(X) == 0:
         raise ValueError(
-            "Expected array-like object for imput with prefix {prefix} but got '{X}' object instead.".format(prefix, X))
+            "Expected array-like object for input \
+            with prefix {prefix} but got '{X}' object instead.".format(prefix, X))
```
kbattocchi, line 574: Use a format literal string:

```suggestion
            f"Expected array-like object for imput with prefix {prefix} but got '{X}' object instead.")
```

kbattocchi (commented):
A few more comments on the latest batch of changes.

econml/_cate_estimator.py
```diff
@@ -211,6 +215,54 @@ def ate(self, X=None, *, T0, T1):
...
+
+        Returns
+        -------
+        out_feature_names: list of strings or None
+            Returns feature names.
+        """
+        if feature_names is not None:
+            return feature_names
+        if hasattr(self, "_input_names"):
+            return self._input_names["output_names"]
```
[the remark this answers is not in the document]
moprescu: Great catch, fixed!

econml/_cate_estimator.py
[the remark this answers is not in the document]
moprescu: It creates summary stats for one specific treatment `T` and the treatment name doesn't appear in the table, so it's not needed. It's only for `coef__inference` and `intercept__inference`, the other inferences require treatment names.

econml/inference/_bootstrap.py
```diff
@@ -221,6 +223,8 @@ def fname_transformer(x):
             if prefix == 'effect' or (isinstance(self._wrapped, LinearModelFinalCateEstimatorDiscreteMixin) and
                                       (inf_type == 'coefficient' or inf_type == 'intercept')):
                 d_t = 1
+                # Dummy treatment name
+                input_names["treatment_names"] = None
```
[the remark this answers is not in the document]
moprescu: Same as above, it's not actually needed. Also, if `None` is being passed in, `InferenceResults` will create dummy treatment names internally.

econml/_cate_estimator.py
moprescu: Same as above.

econml/_cate_estimator.py
moprescu: this is called only in the `fit` function of estimators. The input names get refreshed during the `_refit` step so they are coupled., i.e. they are only called together.

econml/_cate_estimator.py
```diff
@@ -205,6 +215,54 @@ def ate(self, X=None, *, T0, T1):
...
+        Public interface for getting output names.
+
+        Returns
+        -------
+        output_names: list of strings
+            Returns output names.
+        """
+        if hasattr(self, "_input_names"):
+            return self._input_names["output_names"]
+        return None
```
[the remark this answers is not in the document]
vsyrgkanis: Also they should be taking as input treatment_names and implement the override logic. This will save override logic from elsewhere. You are already doing this for cate_feature_names and I think we should do it here too.

econml/_cate_estimator.py
```diff
@@ -205,6 +215,54 @@ def ate(self, X=None, *, T0, T1):
...
+        Public interface for getting treatment names.
+
+        Returns
+        -------
+        treatment_names: list of strings
+            Returns treatment names.
+        """
+        if hasattr(self, "_input_names"):
+            return self._input_names["treatment_names"]
+        return None
```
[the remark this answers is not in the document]
vsyrgkanis: Also they should be taking as input treatment_names and implement the override logic. This will save override logic from elsewhere. You are already doing this for cate_feature_names and I think we should do it here too.

econml/_cate_estimator.py
```diff
@@ -904,8 +962,8 @@ def summary(self, alpha=0.1, value=0, decimals=3, feature_names=None, treatment_
             converted to various output formats.
         """
         # Get input names
-        treatment_names = self._input_names["treatment_names"] if treatment_names is None else treatment_names
-        output_names = self._input_names["output_names"] if output_names is None else output_names
+        treatment_names = self.cate_treatment_names() if treatment_names is None else treatment_names
```
vsyrgkanis, line 965: Once the `cate_treatment_names` take an optional input, with default None, this line should just be: `self.cate_treatment_names(treatment_names)` and `self.cate_output_names(output_names)`

econml/_cate_estimator.py
```diff
@@ -1168,8 +1226,8 @@ def summary(self, T, *, alpha=0.1, value=0, decimals=3,
         """
         # Get input names
         feature_names = self.cate_feature_names() if feature_names is None else feature_names
-        treatment_names = self._input_names["treatment_names"] if treatment_names is None else treatment_names
-        output_names = self._input_names["output_names"] if output_names is None else output_names
+        treatment_names = self.cate_treatment_names() if treatment_names is None else treatment_names
```
vsyrgkanis, line 1229: Once the cate_treatment_names take an optional input, with default None, this line should just be: self.cate_treatment_names(treatment_names) and self.cate_output_names(output_names)

econml/_ortho_learner.py
```diff
@@ -605,6 +605,7 @@ def fit(self, Y, T, X=None, W=None, Z=None, *, sample_weight=None, sample_var=No
                 self.transformer = FunctionTransformer(
                     func=_EncoderWrapper(self._one_hot_encoder).encode,
                     validate=False)
+                self._set_encoded_treatment_names(self._one_hot_encoder)
```
vsyrgkanis, line 608: I'm not a super fan of this: This is logic that is at the level of the `TreatmentExpansionMixin`, but now we are pushing it to a lower level, which seems wrong. 

Suppose for instance that someone comes and wants to contribute an estimator. They would need to know that if I expand treatments, then I need to call this function too, but this is not enforced by any library contract.

I think we should think harder if there is a way to push this to the TreatmentExpansionMixin level or to at least make this part of a contract that is enforcible, e.g. that any class that inherits fthe TreatmentExpansionMixin, must also be implementing an `ecnoded_treamtnent_names` and whenever this happens we update the treatment names at the `TreatmentExpansionMixin` level. This way if someone implements an estimator that inherits from this mixin, they'll get an abstract class error, becuase they haven't implemented this.

econml/inference/_inference.py
```diff
@@ -240,8 +243,11 @@ def effect_inference(self, X, *, T0, T1):
         e_stderr = self._prediction_stderr(cross_product(X, T1 - T0))
         d_y = self._d_y[0] if self._d_y else 1
         # d_t=1 here since we measure the effect across all Ts
+        # A dummy treatment names is passed in
         return NormalInferenceResults(d_t=1, d_y=d_y, pred=e_pred,
-                                      pred_stderr=e_stderr, inf_type='effect', **self._input_names)
+                                      pred_stderr=e_stderr, inf_type='effect',
+                                      feature_names=self._input_names["feature_names"],
```
vsyrgkanis, line 249: I don't like that this is accessing the `_input_names` dict directly. This hsould instead be accessing the public `cate_feature_names` and `cate_output_names`.

econml/inference/_inference.py
```diff
@@ -425,9 +431,12 @@ def effect_inference(self, X, *, T0, T1):
         pred = pred[np.arange(T0.shape[0]), ..., ind]
         pred_stderr = pred_stderr[np.arange(T0.shape[0]), ..., ind] if pred_stderr is not None else None
         # d_t=1 here since we measure the effect across all Ts
+        # Omit treatment name argument to fall back to the default
         return NormalInferenceResults(d_t=1, d_y=self.d_y, pred=pred,
                                       pred_stderr=pred_stderr,
-                                      inf_type='effect', **self._input_names)
+                                      inf_type='effect',
+                                      feature_names=self._input_names["feature_names"],
```
vsyrgkanis, line 438: I don't like that this is accessing the _input_names dict directly. This hsould instead be accessing the public cate_feature_names and cate_output_names.

econml/inference/_inference.py
```diff
@@ -425,9 +431,12 @@ def effect_inference(self, X, *, T0, T1):
         pred = pred[np.arange(T0.shape[0]), ..., ind]
         pred_stderr = pred_stderr[np.arange(T0.shape[0]), ..., ind] if pred_stderr is not None else None
         # d_t=1 here since we measure the effect across all Ts
+        # Omit treatment name argument to fall back to the default
```
vsyrgkanis, line 434: Why is this the case? Why not pass the treatment names too? Seems very weird behavior

econml/inference/_inference.py
```diff
@@ -462,9 +471,12 @@ def fname_transformer(x):
         else:
             def fname_transformer(x):
                 return x
+        # d_t=1 here since we measure the effect across all Ts
```
vsyrgkanis, line 474: I don't see why it should be a problem to pass a treatment name. This seems a logic that should be fixed inside the inference resutls if it's problematic to pass in a treamtnet name

econml/inference/_inference.py
```diff
@@ -462,9 +471,12 @@ def fname_transformer(x):
...
+        # Omit treatment name argument to fall back to the default
         return NormalInferenceResults(d_t=1, d_y=self.d_y, pred=coef, pred_stderr=coef_stderr,
                                       inf_type='coefficient', fname_transformer=fname_transformer,
-                                      **self._input_names)
+                                      feature_names=self._input_names["feature_names"],
```
vsyrgkanis, line 478: I don't like that this is accessing the _input_names dict directly. This hsould instead be accessing the public cate_feature_names and cate_output_names.

econml/inference/_inference.py
```diff
@@ -486,9 +498,13 @@ def intercept__inference(self, T):
             warn("Final model doesn't have a `intercept_stderr_` attribute. "
                  "Only point estimates will be available.")
             intercept_stderr = None
+        # d_t=1 here since we measure the effect across all Ts
```
vsyrgkanis, line 501: similarly here

econml/inference/_inference.py
```diff
@@ -486,9 +498,13 @@ def intercept__inference(self, T):
...
+        # Omit treatment name argument to fall back to the default
         return NormalInferenceResults(d_t=1, d_y=self.d_y, pred=self.fitted_models_final[ind].intercept_,
                                       pred_stderr=intercept_stderr,
-                                      inf_type='intercept', **self._input_names)
+                                      inf_type='intercept',
+                                      feature_names=self._input_names["feature_names"],
```
vsyrgkanis, line 506: I don't like that this is accessing the _input_names dict directly. This hsould instead be accessing the public cate_feature_names and cate_output_names.

econml/metalearners/_metalearners.py
```diff
@@ -77,6 +77,7 @@ def fit(self, Y, T, X, *, inference=None):
         # Check inputs
         Y, T, X, _ = check_inputs(Y, T, X, multi_output_T=False)
         T = self._one_hot_encoder.fit_transform(T.reshape(-1, 1))
+        self._set_encoded_treatment_names(self._one_hot_encoder)
```
vsyrgkanis, line 80: I think the fact that this single line appears everywhere is a bit problematic and the reason why I think this logic should be going into the TreatmentExpansionMixin. For instance I believe you've forgotten to add this line to DRIV and more generally, this could become even hairier to maintain with more and more estimators.

econml/metalearners/_metalearners.py
```diff
@@ -170,6 +171,7 @@ def fit(self, Y, T, X=None, *, inference=None):
             X = np.zeros((Y.shape[0], 1))
         Y, T, X, _ = check_inputs(Y, T, X, multi_output_T=False)
         T = self._one_hot_encoder.fit_transform(T.reshape(-1, 1))
+        self._set_encoded_treatment_names(self._one_hot_encoder, drop_first=True)
```
vsyrgkanis, line 174: I think the fact that this single line appears everywhere is a bit problematic and the reason why I think this logic should be going into the TreatmentExpansionMixin. For instance I believe you've forgotten to add this line to DRIV and more generally, this could become even hairier to maintain with more and more estimators.

econml/metalearners/_metalearners.py
```diff
@@ -277,6 +279,7 @@ def fit(self, Y, T, X, *, inference=None):
         if Y.ndim == 2 and Y.shape[1] == 1:
             Y = Y.flatten()
         T = self._one_hot_encoder.fit_transform(T.reshape(-1, 1))
+        self._set_encoded_treatment_names(self._one_hot_encoder)
```
vsyrgkanis, line 282: I think the fact that this single line appears everywhere is a bit problematic and the reason why I think this logic should be going into the TreatmentExpansionMixin. For instance I believe you've forgotten to add this line to DRIV and more generally, this could become even hairier to maintain with more and more estimators.

econml/metalearners/_metalearners.py
```diff
@@ -403,6 +406,7 @@ def fit(self, Y, T, X, *, inference=None):
         # Check inputs
         Y, T, X, _ = check_inputs(Y, T, X, multi_output_T=False)
         T = self._one_hot_encoder.fit_transform(T.reshape(-1, 1))
+        self._set_encoded_treatment_names(self._one_hot_encoder)
```
vsyrgkanis, line 409: I think the fact that this single line appears everywhere is a bit problematic and the reason why I think this logic should be going into the TreatmentExpansionMixin. For instance I believe you've forgotten to add this line to DRIV and more generally, this could become even hairier to maintain with more and more estimators.

econml/orf/_ortho_forest.py
```diff
@@ -649,6 +649,7 @@ def fit(self, Y, T, X, W=None, *, inference='auto'):
             self.transformer = FunctionTransformer(
                 func=_EncoderWrapper(self._one_hot_encoder).encode,
                 validate=False)
+            self._set_encoded_treatment_names(self._one_hot_encoder)
```
vsyrgkanis, line 652: I think the fact that this single line appears everywhere is a bit problematic and the reason why I think this logic should be going into the TreatmentExpansionMixin. For instance I believe you've forgotten to add this line to DRIV and more generally, this could become even hairier to maintain with more and more estimators.

econml/orf/_ortho_forest.py
```diff
@@ -995,6 +996,7 @@ def fit(self, Y, T, X, W=None, *, inference='auto'):
         self.transformer = FunctionTransformer(
             func=_EncoderWrapper(self._one_hot_encoder).encode,
             validate=False)
+        self._set_encoded_treatment_names(self._one_hot_encoder)
```
vsyrgkanis, line 999: I think the fact that this single line appears everywhere is a bit problematic and the reason why I think this logic should be going into the TreatmentExpansionMixin. For instance I believe you've forgotten to add this line to DRIV and more generally, this could become even hairier to maintain with more and more estimators.

econml/orf/_ortho_forest.py
```diff
@@ -1298,8 +1300,12 @@ def effect_inference(self, X=None, *, T0=0, T1=1):
             a dataframe summary of these inference results.
         """
         eff, scales = self._effect_inference_helper(X, T0, T1)
+        # d_t=1 here since we measure the effect across all Ts
```
vsyrgkanis, line 1303: This again is super weird.

vsyrgkanis (dismissed):

econml/_cate_estimator.py
```diff
@@ -205,6 +215,54 @@ def ate(self, X=None, *, T0, T1):
...
+        Returns
+        -------
+        out_feature_names: list of strings or None
+            Returns feature names.
+        """
+        if feature_names is not None:
+            return feature_names
+        if hasattr(self, "_input_names"):
+            return self._input_names["feature_names"]
+        return None
```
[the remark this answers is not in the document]
moprescu: This is already done in `_set_input_names`. The default is to set names like this when metadata is missing. E.g.:
```Python
 self._input_names = {
            "feature_names": get_input_columns(X, prefix="X"),
            "output_names": get_input_columns(Y, prefix="Y"),
            "treatment_names": get_input_columns(T, prefix="T")
        }
```

econml/_cate_estimator.py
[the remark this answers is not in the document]
moprescu: > These should be returning default names e.g. [y0, y1...] so that we uniformize the use of default names across our package.

This is already done, see above answer.

> Also they should be taking as input treatment_names and implement the override logic. This will save override logic from elsewhere. You are already doing this for cate_feature_names and I think we should do it here too.

Great suggestions, fixed!

econml/_cate_estimator.py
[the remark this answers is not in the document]
moprescu: > These should be returning default names e.g. [y0, y1...] so that we uniformize the use of default names across our package.

This is already done, see above answer.

> Also they should be taking as input treatment_names and implement the override logic. This will save override logic from elsewhere. You are already doing this for cate_feature_names and I think we should do it here too.

Great suggestions, fixed!

econml/_cate_estimator.py
moprescu: Fixed!

econml/_cate_estimator.py
moprescu: Done!

econml/metalearners/_metalearners.py
moprescu: I refactored this and removed much of the old logic due to updates in since sklearn 0.20. Take a look!

econml/metalearners/_metalearners.py
moprescu: I refactored this and removed much of the old logic due to updates in since sklearn 0.20.

econml/orf/_ortho_forest.py
moprescu: I refactored this and removed much of the old logic due to updates in since sklearn 0.20.

econml/orf/_ortho_forest.py
moprescu: I refactored this and removed much of the old logic due to updates in since sklearn 0.20.

econml/orf/_ortho_forest.py
moprescu: Fixed by setting `d_t=None` when summarizing over all treatments.

econml/_ortho_learner.py
moprescu: I refactored this and removed much of the old logic due to updates in since sklearn 0.20.

econml/inference/_inference.py
moprescu: Fixed!

econml/inference/_inference.py
moprescu: Fixed!

econml/inference/_inference.py
moprescu: Fixed by setting `d_t=None` when summarizing over all treatments.

econml/inference/_inference.py
moprescu: Fixed by setting `d_t=None` when summarizing over all treatments.

econml/inference/_inference.py
moprescu: Fixed!

econml/inference/_inference.py
moprescu: Fixed by setting `d_t=None` when summarizing over all treatments.

econml/inference/_inference.py
moprescu: Fixed!

econml/metalearners/_metalearners.py
moprescu: I refactored this and removed much of the old logic due to updates in since sklearn 0.20.

econml/orf/_ortho_forest.py
```diff
@@ -248,6 +248,8 @@ def __init__(self,
         self.verbose = verbose
         self.batch_size = batch_size
         self.categories = categories
+        if self.categories != 'auto':
+            self.categories = [self.categories]  # OneHotEncoder expects a 2D array with features per column
```
kbattocchi, line 252: This and the corresponding changes to `DMLOrthoForest.fit` and `DROrthoForest.fit` look wrong.  In order to better align with SKLearn and user expectations it should be possible to modify categories between calls to fit to use the same estimator with datasets of different shapes.

kbattocchi, line 252: But on a related note it also looks like there's an existing bug in those `fit` methods because they are reassiging `self.categories` to the list-wrapped version if it's not `'auto'`, which means that multiple calls to fit will wrap the list multiple times and therefore will break.  Could you add a test that calls `fit` more than once on those OrthoForest flavors with non-default categories, and make sure that it passes?

kbattocchi (changes requested):
This looks great.  However, there are a few issues with how categories are treated that need to be addressed.

econml/_cate_estimator.py
```diff
@@ -44,18 +44,28 @@ def _get_inference(self, inference):
...
+            "treatment_names": get_input_columns(T, prefix="T")
         }
         if set_flag:
             # This flag is true when names are set in a child class instead
             # If names are set in a child class, add an attribute reflecting that
             self._input_names_set = True
 
+    def _set_encoded_treatment_names(self, one_hot_encoder, drop_first=False):
+        """Works with sklearn OHEs"""
+        # Drop first
```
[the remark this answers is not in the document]
kbattocchi: Oops, looks like this was already changed in a subsequent commit.

econml/_cate_estimator.py
```diff
@@ -59,11 +59,12 @@ def _set_input_names(self, Y, T, X, set_flag=False):
 
     def _set_encoded_treatment_names(self, one_hot_encoder, drop_first=False):
```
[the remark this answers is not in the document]
moprescu: I just noticed I forgot to remove this piece. This functionality has been moved to `TreatmentExpansionMixin` under `_set_transformed_treatment_names`. `_set_encoded_treatment_names` is not being used anywhere and doesn't belong in the `BaseCateEstimator` anymore. Will push another commit with the removal.

econml/metalearners/_metalearners.py
```diff
@@ -185,8 +184,11 @@ def fit(self, Y, T, X=None, *, inference=None):
             X = np.zeros((Y.shape[0], 1))
         Y, T, X, _ = check_inputs(Y, T, X, multi_output_T=False)
         T = self.transformer.fit_transform(T.reshape(-1, 1))
-        self._d_t = (T.shape[1] - 1,)
-        feat_arr = np.concatenate((X, T), axis=1)
+        self._d_t = (T.shape[1], )
+        # Note: unlike other Metalearners, we don't drop the first column because
```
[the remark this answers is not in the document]
moprescu: The change from previous iterations: we are dropping first in the OHE transformer just like in the other cate estimators, and we reconstruct the dropped column when fitting the algorithm. That's the only place where the controls column is needed. I will change the comment a bit to reflect that we add back the controls column.

econml/metalearners/_metalearners.py
```diff
@@ -133,7 +147,10 @@ def __init__(self, *,
                  overall_model,
                  categories='auto'):
         self.overall_model = clone(overall_model, safe=False)
+        if categories != 'auto':
+            categories = [categories]  # OneHotEncoder expects a 2D array with features per column
         self.categories = categories
+        self.transformer = OneHotEncoder(categories=self.categories, sparse=False, drop='first')
```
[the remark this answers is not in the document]
moprescu: Good point. The fundamental problem here is that `categories` is set during `__init__` or at some other point by the user rather than during `fit` despite the fact that `categories` is a property of `T`. I fixed this by implementing a special setter for `categories` in `TreatmentExpansionMixin`, see latest commit.

econml/metalearners/_metalearners.py
```diff
@@ -72,23 +75,34 @@ def fit(self, Y, T, X, *, inference=None):
...
+
+        Returns
+        -------
+        out_treatment_names: list of strings
+            Returns expanded treatment names.
+        """
+        if treatment_names is not None:
+            return self.transformer.get_feature_names(treatment_names).tolist()
+        # Treatment names is None, default to BaseCateEstimator
+        return super().cate_treatment_names()
```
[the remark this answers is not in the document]
moprescu: Good catch, moved it to `TreatmentExpansionMixin`.

econml/metalearners/_metalearners.py
```diff
@@ -39,7 +39,10 @@ def __init__(self, *,
                  models,
                  categories='auto'):
         self.models = clone(models, safe=False)
+        if categories != 'auto':
+            categories = [categories]  # OneHotEncoder expects a 2D array with features per column
         self.categories = categories
+        self.transformer = OneHotEncoder(categories=self.categories, sparse=False, drop='first')
```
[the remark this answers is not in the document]
moprescu: Good point. I fixed this by implementing a special setter for `categories` in `TreatmentExpansionMixin`, see latest commit.

econml/metalearners/_metalearners.py
```diff
@@ -247,7 +275,10 @@ def __init__(self, *,
         self.models = clone(models, safe=False)
         self.cate_models = clone(cate_models, safe=False)
         self.propensity_model = clone(propensity_model, safe=False)
+        if categories != 'auto':
+            categories = [categories]  # OneHotEncoder expects a 2D array with features per column
         self.categories = categories
+        self.transformer = OneHotEncoder(categories=self.categories, sparse=False, drop='first')
```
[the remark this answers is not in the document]
moprescu: Good point. I fixed this by implementing a special setter for `categories` in `TreatmentExpansionMixin`, see latest commit.

econml/metalearners/_metalearners.py
```diff
@@ -379,7 +421,10 @@ def __init__(self, *,
         self.models = clone(models, safe=False)
         self.final_models = clone(final_models, safe=False)
         self.propensity_model = clone(propensity_model, safe=False)
+        if categories != 'auto':
+            categories = [categories]  # OneHotEncoder expects a 2D array with features per column
         self.categories = categories
+        self.transformer = OneHotEncoder(categories=self.categories, sparse=False, drop='first')
```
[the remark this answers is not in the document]
moprescu: Good point. I fixed this by implementing a special setter for `categories` in `TreatmentExpansionMixin`, see latest commit.

econml/orf/_ortho_forest.py
moprescu: Some of this is because the rebase onto master kept some old code around. I fixed the issues and I implemented a special setter for `categories` in `TreatmentExpansionMixin`, see latest commit.

moprescu: The special setter for `categories` in `TreatmentExpansionMixin` fixed this issue.

econml/inference/_inference.py
```diff
@@ -148,7 +147,10 @@ def const_marginal_effect_inference(self, X):
             warn("Final model doesn't have a `prediction_stderr` method, "
                  "only point estimates will be returned.")
         return NormalInferenceResults(d_t=self.d_t, d_y=self.d_y, pred=pred,
-                                      pred_stderr=pred_stderr, inf_type='effect', **self._input_names)
+                                      pred_stderr=pred_stderr, inf_type='effect',
+                                      feature_names=self._est.cate_feature_names(),
+                                      output_names=self._est.cate_output_names(),
+                                      treatment_names=self._est.cate_treatment_names())
```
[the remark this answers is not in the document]
moprescu: I think it's better to pass these explicitly since in some cases the `treatment_names` could be None.

econml/inference/_inference.py
```diff
@@ -702,32 +707,34 @@ def summary_frame(self, alpha=0.1, value=0, decimals=3,
...
+            to_include['zstat'] = self._array_to_frame(self._d_t, self.d_y, self.zstat(value),
                                                        output_names=output_names, treatment_names=treatment_names)
-            to_include['pvalue'] = self._array_to_frame(self.d_t, self.d_y, self.pvalue(value),
+            to_include['pvalue'] = self._array_to_frame(self._d_t, self.d_y, self.pvalue(value),
                                                         output_names=output_names, treatment_names=treatment_names)
-            to_include['ci_lower'] = self._array_to_frame(self.d_t, self.d_y, ci_mean[0],
+            to_include['ci_lower'] = self._array_to_frame(self._d_t, self.d_y, ci_mean[0],
                                                           output_names=output_names, treatment_names=treatment_names)
-            to_include['ci_upper'] = self._array_to_frame(self.d_t, self.d_y, ci_mean[1],
+            to_include['ci_upper'] = self._array_to_frame(self._d_t, self.d_y, ci_mean[1],
```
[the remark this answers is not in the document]
moprescu: I noticed that too, but i though maybe there was a reason for this, maybe something to do with future usage. I will defer to Maggie's PR.

econml/inference/_inference.py
```diff
@@ -1299,36 +1309,36 @@ def _print(self, *, alpha=None, value=None, decimals=None, tol=None, output_name
...
+        res1 = self._res_to_2darray(self._d_t, self.d_y, self.mean_point, decimals)
         if self.pred_stderr is not None:
-            res1 = np.hstack((res1, self._res_to_2darray(self.d_t, self.d_y, self.stderr_mean, decimals)))
-            res1 = np.hstack((res1, self._res_to_2darray(self.d_t, self.d_y, self.zstat(value=value), decimals)))
-            res1 = np.hstack((res1, self._res_to_2darray(self.d_t, self.d_y, self.pvalue(value=value), decimals)))
-            res1 = np.hstack((res1, self._res_to_2darray(self.d_t, self.d_y,
+            res1 = np.hstack((res1, self._res_to_2darray(self._d_t, self.d_y, self.stderr_mean, decimals)))
+            res1 = np.hstack((res1, self._res_to_2darray(self._d_t, self.d_y, self.zstat(value=value), decimals)))
+            res1 = np.hstack((res1, self._res_to_2darray(self._d_t, self.d_y, self.pvalue(value=value), decimals)))
+            res1 = np.hstack((res1, self._res_to_2darray(self._d_t, self.d_y,
```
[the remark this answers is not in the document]
moprescu: I noticed that too, but i though maybe there was a reason for this, maybe something to do with future usage. I will defer to Maggie's PR.

econml/_cate_estimator.py
```diff
@@ -742,11 +820,47 @@ def _expand_treatments(self, X=None, *Ts):
...
 
         return (X,) + tuple(outTs)
 
+    def _set_transformed_treatment_names(self):
+        """Works with sklearn OHEs"""
+        if hasattr(self, "_input_names"):
+            self._input_names["treatment_names"] = self.transformer.get_feature_names(
+                self._input_names["treatment_names"]).tolist()
+
+    def __setattr__(self, name, value):
```
kbattocchi, line 834: This seems like the wrong approach to me... Among other things it seems like it would break serialization/deserialization, but also it only handles the repetition in setting categories but not in defining the transformer from that.  I think it would be more natural to override fit itself, never modify the user-set categories, and just define a local variable that handles the list wrapping, and then set transformer based on that, and have all subclasses call super.fit (except that _OrthoLearner probably needs to override it to deal with the fact that continuous treatments don’t need transformers).

moprescu: So I don't agree with this. Here is why:

1. I tested it and it does not break serialization. Classes with overridden setters can be pickled.
2. The transformer is an OHE that is defined at `fit` time with whatever `self.categories` is defined at that point. The alternative is to define a wrapped `categories` in every `fit`, thus duplicating code everywhere and having the same issue as @vasilismsr mentioned: there is boilerplate code that is repeated everywhere despite the fact that it belongs to `TreatmentExpansionMixin` and that makes the code hard to follow and it makes it hard for others to contribute.
3. I am against overriding `fit`. `fit` is already very complicated due to the inheritance and the wrapping. The ORF already has that issue and needs to override `fit` everywhere and losing some of the logic implemented in the `BaseCateEstimator`. I am open to other alternatives, but I think this is a recipe for more issues down the line. 

Since the `categories` attribute is defined some time before `fit`, we could move the transformer definition in the `prefit`, however that will have some implication there that there is a `categories` attribute in the child class that is not mentioned in `TreatmentExpansionMixin` and referencing attributes that might be in a child makes me uneasy. Furthermore, the transformer definition is tied to whether there is a discrete treatment which is again an attribute of the child class and not of the `TreatmentExpansionMixin`. This is an issue for both the `_OrthoLearner` and the `DMLOrthoForest` classes.

kbattocchi (approved):
LGTM

PR #400, opened 2021-02-04
Enable calling dowhy from econml

heimengqi:
Each of our estimator will have a `.dowhy` property, which will allow user to get causal graph and run refutation test.

I still need to update the code to generate columns names, after Miruna's column name PR is in, currently it simplifies the logic to get the columns names, but we might lose the exact name if user inputs pandas dataframe.


econml/dowhy.py
```diff
@@ -0,0 +1,222 @@
...
+        parameters = [p for p in init_signature.parameters.values()
+                      if p.name != 'self' and p.kind != p.VAR_KEYWORD]
+        for p in parameters:
+            if p.kind == p.VAR_POSITIONAL:
+                raise RuntimeError("cate estimators should always "
+                                   "specify their parameters in the signature"
+                                   " of their __init__ (no varargs)."
+                                   " %s with constructor %s doesn't "
+                                   " follow this convention."
+                                   % (self._cate_estimator, init_signature))
```
kbattocchi, line 49: Use a format literal string for readability:

```suggestion
                raise RuntimeError("cate estimators should always "
                                   "specify their parameters in the signature"
                                   " of their __init__ (no varargs)."
                                   f" {self._cate_estimator} with constructor {init_signature} doesn't "
                                   " follow this convention.")
```

econml/dowhy.py
```diff
@@ -0,0 +1,222 @@
...
+        self._cate_estimator = cate_estimator
+
+    def _get_params(self):
+        init = self._cate_estimator.__init__
+        # introspect the constructor arguments to find the model parameters
+        # to represent
+        init_signature = inspect.signature(init)
+        # Consider the constructor parameters excluding 'self'
+        parameters = [p for p in init_signature.parameters.values()
+                      if p.name != 'self' and p.kind != p.VAR_KEYWORD]
```
kbattocchi, line 41: [very minor] Technically the self argument can have an arbitrary name (but in practice it should always be called 'self', so this probably isn't important), so it might be better to assert that there is at least one positional argument and ignore the first (which will be the self argument, whatever its name).

econml/dowhy.py
```diff
@@ -0,0 +1,222 @@
...
+        for p in parameters:
+            if p.kind == p.VAR_POSITIONAL:
```
kbattocchi, line 43: If we don't allow a VAR_POSITIONAL argument, should we also disallow VAR_KEYWORD? Right now it looks like if there is a VAR_KEYWORD param we're just completely ignoring it?

econml/dowhy.py
```diff
@@ -0,0 +1,222 @@
...
+            outcome=outcome_names,
+            graph=graph,
+            common_causes=feature_names + confounder_names if X.shape[1] > 0 or W.shape[1] > 0 else None,
+            instruments=instrument_names if Z.shape[1] > 0 else None,
+            effect_modifiers=feature_names if X.shape[1] > 0 else None,
+            estimand_type=estimand_type,
+            proceed_when_unidetifiable=proceed_when_unidentifiable,
+            missing_nodes_as_confounders=missing_nodes_as_confounders
+        )
+        self.identified_estimand_ = self.dowhy_.identify_effect(proceed_when_unidentifiable=True)
```
kbattocchi, line 146: Why is `proceed_when_unidentifiable` True here instead of using the parameter with the same name?

econml/dowhy.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        return self.dowhy_.refute_estimate(
+            self.identified_estimand_, self.estimate_, method_name=method_name, **kwargs
+        )
+
+    # We don't allow user to call refit_final from this class, since internally dowhy effect estimate will only update
+    # cate estimator but not the effect.
+    def refit_final(self, inference=None):
+        raise AttributeError(
+            "Method refit_final is not allowed here! Please call it from Cate Estimator directly! ")
```
kbattocchi, line 196: I'm not sure that we want to recommend calling it on the estimator directly, since (unless I'm misunderstanding) this wrapper will still not function as expected after refitting.

econml/dowhy.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __getattr__(self, attr):
+        # don't proxy special methods
+        if attr.startswith('__'):
+            raise AttributeError(attr)
+        elif attr in ['_cate_estimator', 'dowhy_',
+                      'identified_estimand_', 'estimate_']:
+            self.__dict__[attr] = value
```
kbattocchi, line 204: This looks like a copy/paste of the setter code (`value` isn't even defined in this method).  If your code works, I assume that means this whole `elif` block can be removed because it's never called.

econml/dowhy.py
```diff
@@ -0,0 +1,222 @@
...
+        elif hasattr(self.estimate_._estimator_object, attr):
+
+            return getattr(self.estimate_._estimator_object, attr)
+        else:
+            return getattr(self.dowhy_, attr)
+
+    def __setattr__(self, attr, value):
+        if attr in ['_cate_estimator', 'dowhy_',
+                    'identified_estimand_', 'estimate_']:
+            self.__dict__[attr] = value
```
kbattocchi, line 216: Assuming it works, I think maybe 

```suggestion
            super().__setattr__(attr, value)
```
would be slightly more understandable than setting the value directly.

econml/dowhy.py
```diff
@@ -0,0 +1,222 @@
...
+        elif attr.startswith('dowhy__'):
+            setattr(self.dowhy_, attr[7:], value)
```
kbattocchi, line 218: I think this is slightly easier to understand:

```suggestion
            setattr(self.dowhy_, attr[len('dowhy__'):], value)
```

econml/dowhy.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __getattr__(self, attr):
+        # don't proxy special methods
+        if attr.startswith('__'):
+            raise AttributeError(attr)
+        elif attr in ['_cate_estimator', 'dowhy_',
+                      'identified_estimand_', 'estimate_']:
+            self.__dict__[attr] = value
+        elif attr.startswith('dowhy__'):
+            return getattr(self.dowhy_, attr[7:])
```
kbattocchi, line 206: I think this is slightly easier to understand:
```suggestion
            return getattr(self.dowhy_, attr[len('dowhy__'):])
```

econml/dowhy.py
```diff
@@ -0,0 +1,222 @@
...
+            "Method refit_final is not allowed here! Please call it from Cate Estimator directly! ")
+
+    def __getattr__(self, attr):
+        # don't proxy special methods
+        if attr.startswith('__'):
+            raise AttributeError(attr)
+        elif attr in ['_cate_estimator', 'dowhy_',
+                      'identified_estimand_', 'estimate_']:
+            self.__dict__[attr] = value
+        elif attr.startswith('dowhy__'):
```
kbattocchi, line 205: Do we document anywhere that it's possible to access the dowhy attributes in this way if desired?  And since we already expose the inner object via the public `dowhy_` attribute, does it make sense to support passing attributes through like this?

econml/dowhy.py
```diff
@@ -0,0 +1,222 @@
...
+            return getattr(self.dowhy_, attr[7:])
+        elif hasattr(self.estimate_._estimator_object, attr):
```
kbattocchi, line 207: If both the estimator and the dowhy_ have the same property, it might be nice to warn the user that the call is ambiguous and that we're defaulting to the estimator's attribute rather than dowhy's.

econml/metalearners/_metalearners.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, *,
                  models,
                  categories='auto'):
         self.models = clone(models, safe=False)
+        self.categories = categories
```
kbattocchi, line 42: Note that if this attribute (which is now public) is set between calls to fit, the new value won't be used since the OneHotEncoder won't be updated.

kbattocchi: So maybe add a TODO to this file (if there isn't already one) that we need to move a bunch of metalearner logic to `fit` to be more like sklearn/ortholearner.

econml/tests/test_dowhy.py
```diff
@@ -0,0 +1,66 @@
...
+                  "dmlateiv": DMLATEIV(model_Y_W=reg(),
+                                       model_T_W=clf(),
+                                       model_Z_W=reg(),
+                                       discrete_treatment=True,
+                                       discrete_instrument=False)}
+        for name, model in models.items():
+            with self.subTest(name=name):
+                est = model
+                if name == "xlearner":
+                    est_dowhy = est.dowhy.fit(Y, T, X=np.hstack((X, W)), W=None)
```
kbattocchi, line 48: Why do we have to stack X and W in this case?

econml/utilities.py
```diff
@@ -1266,6 +1266,18 @@ def transpose_dictionary(d):
     return output
 
 
+def reshape_arrays_2dim(length, *args):
+    new_args = []
+    for arg in args:
+        if arg is None:
+            new_args.append(np.array([]).reshape(length, 0))
```
kbattocchi, line 1273: Is this logic correct?  For example, if X is None for DML, then that's not equivalent to passing a 0-column X, it's equivalent to passing a column of ones for X instead.  I would expect that the right thing to do is to just pass None through to the output and let the estimator handle it accordingly.

kbattocchi (changes requested):
Mostly looks good, but added a few questions/comments.

econml/dowhy.py
heimengqi: if False, it will pop up a question ask the user whether you want to continue, since here we will do identify, estimate anyway, so we should always proceed. However, when define CausalModel, we still allow user to set this parameters, then they could call est_dowhy.identify_effect() themselves.

econml/dowhy.py
heimengqi: Yes, the logic the user should follow is 
```
est=LinearDML()
est_dowhy=est.dowhy.fit()
```
I mean if you want to call refit, please call
```
est.fit()
est.refit_final()
```

Any suggestion on writing a better error message maybe?

econml/tests/test_dowhy.py
heimengqi: it's a personal choice, since metalearner we don't have W, I could either input X,W, or stack of them as X.

econml/utilities.py
heimengqi: This function is only used to concatenate a pandas dataframe as the input to dowhy CausalModel(). I don't want to check one by one whether they are None, when concatenate data and column names.

econml/dowhy.py
heimengqi: Good catch! If remove those lines I will get an error "maximum recursion depth exceeded", and I know why now. When we search `hasattr(self.estimate_._estimator_object, attr)`, `self.estimate_` causes `__getattr__` to be called again, so it's a infinite loop.

econml/dowhy.py
kbattocchi: Well, the question to me is whether there's really a refit story that makes sense, or whether we should just say it is unsupported and not suggest any workaround other than doing a full fit instead.

I guess my question is, let's say I do

```python3
est = LinearDML()
dw = est.dowhy_
dw.fit(...)
est.refit_final()
```

It seems like the current error message suggests this is the right way to do things, but it's still not okay to try to call `dw.refute`, etc., right, because its internal estimator is out of date? Is there some way to sync the refitted estimator back up to the dowhy wrapper?

If not, then I think the error message should just be

```suggestion
            "Method refit_final is not allowed through a dowhy object; please perform a full fit instead.")
```

econml/dowhy.py
kbattocchi: Hm. That's still somewhat surprising to me because `self.estimate_` should be a normal attribute, right, in which case I'd have thought that `__getattr__` would not even be called...  Or is that something that only happens before self.estimate_ is ever set for the first time?

econml/utilities.py
kbattocchi: Hmm... So does X=None work for DML through dowhy?

kbattocchi (approved):
LGTM

econml/_shap.py
```diff
@@ -144,12 +150,12 @@ def _shap_explain_model_cate(cme_model, models, X, d_t, d_y, feature_names=None,
         except Exception as e:
             print("Final model can't be parsed, explain const_marginal_effect() instead!", repr(e))
             return _shap_explain_cme(cme_model, X, d_t_, d_y_,
-                                     feature_names=feature_names_,
+                                     feature_names=None,
```
vsyrgkanis, line 153: why aren't we passing the un-featureized feature names here??

econml/_shap.py
```diff
@@ -302,13 +315,13 @@ def _shap_explain_multitask_model_cate(cme_model, multitask_model_cate, X, d_t,
...
```
vsyrgkanis, line 318: why aren't we passing the un-featureized feature names here??

econml/metalearners/_metalearners.py
```diff
@@ -39,6 +39,8 @@ def __init__(self, *,
...
+        # TODO: We should move this logic to fit, currenly update category will not update one hot encoder.
```
vsyrgkanis, line 43: We have to move this logic to fit in this PR

econml/metalearners/_metalearners.py
```diff
@@ -128,6 +130,7 @@ def __init__(self, *,
                  overall_model,
                  categories='auto'):
         self.overall_model = clone(overall_model, safe=False)
+        self.categories = categories
         if categories != 'auto':
```
vsyrgkanis, line 134: We have to move this logic to fit in this PR

econml/metalearners/_metalearners.py
```diff
@@ -238,6 +241,7 @@ def __init__(self, *,
         self.models = clone(models, safe=False)
         self.cate_models = clone(cate_models, safe=False)
         self.propensity_model = clone(propensity_model, safe=False)
+        self.categories = categories
         if categories != 'auto':
```
vsyrgkanis, line 245: We have to move this logic to fit in this PR

econml/metalearners/_metalearners.py
```diff
@@ -366,6 +370,7 @@ def __init__(self, *,
         self.models = clone(models, safe=False)
         self.final_models = clone(final_models, safe=False)
         self.propensity_model = clone(propensity_model, safe=False)
+        self.categories = categories
         if categories != 'auto':
```
vsyrgkanis, line 374: We have to move this logic to fit in this PR

econml/orf/_ortho_forest.py
```diff
@@ -928,7 +930,7 @@ def __init__(self, *,
         # Define parameter estimators
         parameter_estimator = DROrthoForest.parameter_estimator_func
```
vsyrgkanis, line 931: We have to move this logic to fit in this PR

econml/orf/_ortho_forest.py
```diff
@@ -928,7 +930,7 @@ def __init__(self, *,
...
         second_stage_parameter_estimator = DROrthoForest.second_stage_parameter_estimator_gen(
```
vsyrgkanis, line 932: We have to move this logic to fit in this PR

econml/orf/_ortho_forest.py
```diff
@@ -928,7 +930,7 @@ def __init__(self, *,
...
-            lambda_reg)
+            self.lambda_reg)
         # Define moment and mean gradient estimator
         moment_and_mean_gradient_estimator = DROrthoForest.moment_and_mean_gradient_estimator_func
```
vsyrgkanis, line 935: We have to move this logic to fit in this PR

econml/orf/_ortho_forest.py
```diff
@@ -928,7 +930,7 @@ def __init__(self, *,
...
         if categories != 'auto':
```
vsyrgkanis, line 936: We have to move this logic to fit in this PR

econml/dowhy.py
```diff
@@ -0,0 +1,226 @@
...
+        # create dataframe
+        n_obs = Y.shape[0]
+        Y, T, X, W, Z = reshape_arrays_2dim(n_obs, Y, T, X, W, Z)
+
+        # currently dowhy only support single outcome and single treatment
+        assert Y.shape[1] == 1, "Can only accept single dimensional outcome."
+        assert T.shape[1] == 1, "Can only accept single dimensional treatment."
+
+        # column names
+        if outcome_names is None:
```
vsyrgkanis, line 119: WE should be using @moprescu  `cate_outcome_names`

vsyrgkanis: Similarly for all the other names below, so that default names are uniformized

econml/dowhy.py
```diff
@@ -0,0 +1,226 @@
...
+        """Refute an estimated causal effect.
+
+        If method_name is provided, uses the provided method. In the future, we may support automatic
+        selection of suitable refutation tests.
+        Following refutation methods are supported:
+            * Adding a randomly-generated confounder: "random_common_cause"
+            * Adding a confounder that is associated with both treatment and outcome: "add_unobserved_common_cause"
+            * Replacing the treatment with a placebo (random) variable): "placebo_treatment_refuter"
+            * Removing a random subset of the data: "data_subset_refuter"
+
```
vsyrgkanis, line 170: Can we point to the docs of the `refute_estimate` method in DoWhy for more info

econml/dowhy.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+
+    # We don't allow user to call refit_final from this class, since internally dowhy effect estimate will only update
+    # cate estimator but not the effect.
+    def refit_final(self, inference=None):
+        raise AttributeError(
+            "Method refit_final is not allowed through a dowhy object; please perform a full fit instead.")
+
+    def __getattr__(self, attr):
+        # don't proxy special methods
```
vsyrgkanis, line 196: I suspect at some point we should be checking if the class is fitted and raisiing a warning. 

I beleive a common pitfall would be that people will fit the cate estimatir class before calling the `dowhy()` method and then callign dowhy(), believing that now the class has been fitted. Instead we need to say that the need to first call dowhy and then fit the esimator.

vsyrgkanis (changes requested):

econml/_shap.py
heimengqi: Yes that's the plan. My concern is we have to keep track too many names here. First of all, the logic is we have to check whether the user inputs a feature name, if not see if we could get feature name from `._input_names`, then we have to save those names for both un-featurized names and featurized names and pick the right one when shap explain final model or `const_marginal_effect`. I will follow this logic unless we find a way to simplify it.

econml/metalearners/_metalearners.py
heimengqi: Will do

econml/dowhy.py
heimengqi: Yes, this part is pending to change once Miruna's PR is in.

econml/dowhy.py
heimengqi: Done

econml/orf/_ortho_forest.py
heimengqi: I have moved all the categories related logic to fit, but for example this one, it's also the input of super class, can't be moved to fit unless we change of the input of super class as well.

econml/_shap.py
heimengqi: I think for each shap helper function (except for `_shap_explain_cme`), I will input raw X, fitted featurizer, input name dict, cate_feature_names function, and keep both the raw feature names and featurized feature names from`_define_names`  helper function. It's also needed to be done once Miruna's PR is merged.

Currently there is a bug if users don't specify feature_names, we use the feature_names from _input_names, but it's un-featurized...

econml/dowhy.py
heimengqi: I guess I should use the helper function `get_input_columns` directly since here I need the input name of raw data, for example cate_feature_name will give me name after featurizer, and cate_treatment_name will give me the name after OHE if by any chance user calls fit first.

vsyrgkanis (approved):

econml/dowhy.py
```diff
@@ -0,0 +1,228 @@
...
+import inspect
+import pandas as pd
+import numpy as np
+import warnings
+from dowhy import CausalModel
+from econml.utilities import check_input_arrays, reshape_arrays_2dim
+
+
+class DoWhyWrapper:
+    """A wrapper class to allow user call other methods from dowhy package through EconML.
```
kbattocchi, line 22:
```suggestion
    """
    A wrapper class to allow user call other methods from dowhy package through EconML.
```

econml/dowhy.py
```diff
@@ -0,0 +1,228 @@
...
+                                   f"{self._cate_estimator} with constructor {init_signature} doesn't "
+                                   "follow this convention.")
+        # Extract and sort argument names excluding 'self'
+        return sorted([p.name for p in parameters])
+
+    def fit(self, Y, T, X=None, W=None, Z=None, *, outcome_names=None, treatment_names=None, feature_names=None,
+            confounder_names=None, instrument_names=None, graph=None, estimand_type="nonparametric-ate",
+            proceed_when_unidentifiable=True, missing_nodes_as_confounders=False,
+            control_value=0, treatment_value=1, target_units="ate", **kwargs):
+        """Estimate the counterfactual model from data through dowhy package.
```
kbattocchi, line 54:
```suggestion
        """
        Estimate the counterfactual model from data through dowhy package.
```

econml/dowhy.py
```diff
@@ -0,0 +1,228 @@
...
+                                                     target_units=target_units,
+                                                     method_params={
+                                                         "init_params": init_params,
+                                                         "fit_params": kwargs,
+                                                     },
+                                                     )
+        return self
+
+    def refute_estimate(self, *, method_name, **kwargs):
+        """Refute an estimated causal effect.
```
kbattocchi, line 161:
```suggestion
        """
        Refute an estimated causal effect.
```

econml/dowhy.py
heimengqi: done

PR #406, opened 2021-02-11
Test install

kbattocchi:
This PR is just to test the interactions of pip and numpy 1.20, and whether those break either us or our dependencies.  It is not intended ever to be merged.


auzaneykbal:
I got error when installing, Python 3.8
pip install econml==0.9.0b1

ERROR: Command errored out with exit status 1:
error: command 'gcc' failed with exit status 1

kbattocchi:
@auzaneykbal Could you open a new issue with some more information, please? What operating system are you using? Also, could you post the entire error message in case there are other details that might help us understand?

PR #413, opened 2021-02-19
Create new extras

kbattocchi:
This PR creates new `tf` and `interpret` extras, which install `tensorflow` and `keras` for `tf` and `matplotlib` and `graphviz` for `interpret`, as well as an `all` extra which installs all of those.  This allows us to remove some heavy dependencies from the main package which are only used in a very small number of places (`DeepIV` is the only place `keras` and `tensorflow` are used, and the CATE tree interpreters are the only place that `matplotlib` and `graphviz` are used).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -78,7 +78,7 @@ jobs:
 
         - script: 'python setup.py build_sphinx -b doctest'
           displayName: 'Run doctests'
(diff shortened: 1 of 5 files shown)


gaugup (approved):

heimengqi (approved):
LGTM.

PR #415, opened 2021-02-19
Fixed dependency on sklearn 0.24.0.

moprescu:


econml/sklearn_extensions/model_selection.py
```diff
@@ -376,12 +377,12 @@ def _cross_val_predict(estimator, X, y=None, *, groups=None, cv=None,
     # independent, and that it is pickle-able.
     parallel = Parallel(n_jobs=n_jobs, verbose=verbose,
                         pre_dispatch=pre_dispatch)
-    # TODO. The API of the private scikit-learn `_fit_and_predict` has changed
-    # between 0.23.2 and 0.24. For this to work with <0.24, we need to add a
-    # case analysis based on sklearn version.
     predictions = parallel(delayed(_fit_and_predict)(
         clone(estimator, safe=safe), X, y, train, test, verbose, fit_params, method)
         for train, test in splits)
+    if sklearn.__version__ < "0.24.0":
```
kbattocchi, line 383: [minor] It might be better to use:

```suggestion
    from pkg_resources import parse_version
    if parse_version(sklearn.__version__) < parse_version("0.24.0"):
```

because comparing versions as strings doesn't always work (e.g. "0.10.0" < "0.2.0").  However, I _think_ in this case it is unlikely to matter unless someone is using something like sklearn 0.9 or sklearn 0.100, which are not practically relevant.

kbattocchi (approved):
This is great, but I have a minor suggestion on how to compare versions.

PR #418, opened 2021-02-22
Adjust ate inference to get the exact stderr when final stage is linear.

heimengqi:
Add some context for this implementation:
In order to get the analytical inference of E(theta(x)), we need to know stderr of E(theta(x)), however previously we could only get the conservative upper bound of it. Given the fact that only in linear cate scenario, E(theta(x))=theta(E(x)), and we could get STD(theta(E(x))) by calling `prediction_stderr(X.mean(axis=0))`, which will return the exact standard error of the mean. 

For `const_marginal_effect` we learn theta(x) alone, then the stderr could be adjusted when final stage is linear in both continues and categorical scenario. For `effect`, in continues scenario we learn cross product of X and T (as XT), we get the stderr by calling `prediction_stderr(XT.mean(axis=0))`, in categorical scenario we learn the effect of a specific treatment comparing to baseline, for each sample, the effect could be different given what the treatment is. We couldn't learn theta(XT.mean(axis=0)) unless T1 is a constant or a list of constant. So I only adjust `effect_inference` only when T1 is a constant or list of constant.


econml/inference/_inference.py
```diff
@@ -559,14 +592,14 @@ def __init__(self, d_t, d_y, pred, inf_type, fname_transformer=None,
...
-        self.pred = pred
+        self.pred = copy.deepcopy(pred)
         self.inf_type = inf_type
         self.fname_transformer = fname_transformer
         self.feature_names = feature_names
         self.output_names = output_names
         self.treatment_names = treatment_names
 
-    @property
+    @ property
```
kbattocchi, line 602: I think there was a syntax error in your file at some point that led to an incorrect parse that caused the autoformatter to insert spaces in these decorators - please revert them.
```suggestion
    @property
```

econml/inference/_inference.py
```diff
@@ -581,8 +614,8 @@ def point_estimate(self):
         """
         return self.pred
 
-    @property
-    @abc.abstractmethod
+    @ property
+    @ abc.abstractmethod
```
kbattocchi, line 618:
```suggestion
    @property
    @abc.abstractmethod
```

econml/inference/_inference.py
```diff
@@ -597,7 +630,7 @@ def stderr(self):
         """
         pass
 
-    @property
+    @ property
```
kbattocchi, line 633:
```suggestion
    @property
```

econml/inference/_inference.py
```diff
@@ -614,7 +647,7 @@ def var(self):
             return self.stderr**2
         return None
 
-    @abc.abstractmethod
+    @ abc.abstractmethod
```
kbattocchi, line 650:
```suggestion
    @abc.abstractmethod
```

econml/inference/_inference.py
```diff
@@ -635,7 +668,7 @@ def conf_int(self, alpha=0.1):
         """
         pass
 
-    @abc.abstractmethod
+    @ abc.abstractmethod
```
kbattocchi, line 671:
```suggestion
    @abc.abstractmethod
```

econml/inference/_inference.py
```diff
@@ -800,7 +833,7 @@ def _reshape_array(self, arr):
         arr = arr.flatten()
         return arr
 
-    @abc.abstractmethod
+    @ abc.abstractmethod
```
kbattocchi, line 836:
```suggestion
    @abc.abstractmethod
```

econml/inference/_inference.py
```diff
@@ -839,19 +872,24 @@ class NormalInferenceResults(InferenceResults):
...
-    def __init__(self, d_t, d_y, pred, pred_stderr, inf_type, fname_transformer=None,
+    def __init__(self, d_t, d_y, pred, pred_stderr, mean_pred_stderr, inf_type, fname_transformer=None,
                  feature_names=None, output_names=None, treatment_names=None):
-        self.pred_stderr = pred_stderr
+        self.pred_stderr = copy.deepcopy(pred_stderr)
+        self.mean_pred_stderr = mean_pred_stderr
         super().__init__(d_t, d_y, pred, inf_type, fname_transformer, feature_names, output_names, treatment_names)
 
-    @property
+    @ property
```
kbattocchi, line 892:
```suggestion
    @property
```

econml/inference/_inference.py
```diff
@@ -948,10 +995,10 @@ class EmpiricalInferenceResults(InferenceResults):
 
     def __init__(self, d_t, d_y, pred, pred_dist, inf_type, fname_transformer=None,
                  feature_names=None, output_names=None, treatment_names=None):
-        self.pred_dist = pred_dist
+        self.pred_dist = copy.deepcopy(pred_dist)
         super().__init__(d_t, d_y, pred, inf_type, fname_transformer, feature_names, output_names, treatment_names)
 
-    @property
+    @ property
```
kbattocchi, line 1001:
```suggestion
    @property
```

econml/inference/_inference.py
```diff
@@ -1076,7 +1128,7 @@ def _repr_html_(self):
         '''Display as HTML in IPython notebook.'''
         return self._print().as_html()
 
-    @property
+    @ property
```
kbattocchi, line 1131:
```suggestion
    @property
```

econml/inference/_inference.py
```diff
@@ -1091,7 +1143,7 @@ def mean_point(self):
         """
         return np.mean(self.pred, axis=0)
 
-    @property
+    @ property
```
kbattocchi, line 1146:
```suggestion
    @property
```

econml/inference/_inference.py
```diff
@@ -1180,7 +1234,7 @@ def conf_int_mean(self, *, alpha=.1):
                 np.array([_safe_norm_ppf(1 - alpha / 2, loc=p, scale=err)
                           for p, err in zip(mean_point, stderr_mean)])
 
-    @property
+    @ property
```
kbattocchi, line 1237:
```suggestion
    @property
```

econml/inference/_inference.py
```diff
@@ -1246,7 +1300,7 @@ def conf_int_point(self, *, alpha=.1, tol=.001):
         upper_ci_point = np.array([self._mixture_ppf(1 - alpha / 2, self.pred, self.pred_stderr, tol)])
         return lower_ci_point, upper_ci_point
 
-    @property
+    @ property
```
kbattocchi, line 1303:
```suggestion
    @property
```

kbattocchi (changes requested):
A few comments, mostly minor.  But shouldn't there also be something that tests that we get back a non-None mean_stderr for the linear case?

econml/inference/_inference.py
```diff
@@ -240,15 +242,38 @@ def effect_inference(self, X, *, T0, T1):
...
         elif self.featurizer is not None:
             X = self.featurizer.transform(X)
-        e_pred = self._predict(cross_product(X, T1 - T0))
-        e_stderr = self._prediction_stderr(cross_product(X, T1 - T0))
+        XT = cross_product(X, T1 - T0)
+        e_pred = self._predict(XT)
+        e_stderr = self._prediction_stderr(XT)
         d_y = self._d_y[0] if self._d_y else 1
+
+        mean_XT = XT.mean(axis=0).reshape(1, -1)
```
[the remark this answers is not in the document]
heimengqi: Thanks, done!

econml/inference/_inference.py
heimengqi: Sure, will do all!

heimengqi:
> A few comments, mostly minor. But shouldn't there also be something that tests that we get back a non-None mean_stderr for the linear case?

Sure. I will add a test around that. I also added some context about this PR to explain why I do it in this way. Please also help confirm the whether logic seems right to you.

kbattocchi (approved):
LGTM

PR #422, opened 2021-02-25
Fix a few 0.9.0 bugs

### LINKED ISSUE #421: Feature name checks should be more robust
Not all sklearn transformers handle `get_feature_names` consistently.  For example, for `sklearn.compose.ColumnTransformer`, `get_feature_names` will return generated feature names corresponding to the transforms used with it, but it does not accept the original feature names as an input (see [their docs](https://scikit-learn.org/stable/modules/generated/sklearn.compose.ColumnTransformer.html#sklearn.compose.ColumnTransformer.get_feature_names)).  Unfortunately, the fact that this method exists with this signature breaks our assumptions, as seen here:

```python3
from econml.dml import LinearDML
from sklearn.compose import ColumnTransformer
import numpy as np

Y, T, X = [np.random.normal(size=(100,d)) for d in [1,1,2]]
est = LinearDML(featurizer=ColumnTransformer([('passthrough','passthrough',[0])]))
est.fit(Y,T,X=X)
inf = est.effect_inference(X) # throws an exception!
```

So we should do some more careful checking everywhere we use `get_feature_names` (which is unfortunately in a lot of places...) and fall back to some other behavior if `get_feature_names` can't be used to transform the previous feature names.

It looks like there is ongoing work in the sklearn repo towards more consistency (see e.g. https://github.com/scikit-learn/scikit-learn/pull/18444), but we should work around this in the meantime.

moprescu:
* Remove sklearn version constraint and allow for sklearn > 0.22.0 
* Make `cate_feature_names` more robust to different kinds of featurizers (fixes #421)

### DIFF
--- a/econml/utilities.py
+++ b/econml/utilities.py
@@ -16,6 +16,7 @@ from sklearn.base import TransformerMixin, BaseEstimator
 from sklearn.linear_model import LassoCV, MultiTaskLassoCV, Lasso, MultiTaskLasso
 from functools import reduce, wraps
 from sklearn.utils import check_array, check_X_y
+from sklearn.utils.validation import assert_all_finite
 import warnings
 from warnings import warn
 from sklearn.model_selection import KFold, StratifiedKFold, GroupKFold
@@ -513,7 +514,7 @@ def check_inputs(Y, T, X, W=None, multi_output_T=True, multi_output_Y=True):
     return Y, T, X, W
 
 
-def check_input_arrays(*args, validate_len=True):
+def check_input_arrays(*args, validate_len=True, force_all_finite=True):
     """Cast input sequences into numpy arrays.
 
     Only inputs that are sequence-like will be converted, all other inputs will be left as is.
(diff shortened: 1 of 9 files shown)


econml/utilities.py
```diff
@@ -582,6 +583,25 @@ def get_input_columns(X, prefix="X"):
...
+        arg_no = len(inspect.getfullargspec(featurizer.get_feature_names).args)
+        if arg_no == 1:
+            return featurizer.get_feature_names()
+        elif arg_no == 2:
+            return featurizer.get_feature_names(feature_names)
+    # Featurizer doesn't have 'get_feature_names' or has atypical 'get_feature_names'
+    try:
+        # Get feature names using featurizer
+        dummy_X = np.empty((1, len(feature_names)))
+        return get_input_columns(featurizer.transform(dummy_X))
```
heimengqi, line 598: Minor suggestion, instead of using the default prefix X, might be better to use a prefix like "TransX" to make it clear it's the X after transformation. It's possible that after transformation the dimension of X might change, use the same prefix might be a bit confused.

heimengqi (approved):
Looks good!

econml/utilities.py
```diff
@@ -582,6 +583,25 @@ def get_input_columns(X, prefix="X"):
...
+        # Get number of arguments, some sklearn featurizer don't accept feature_names
+        arg_no = len(inspect.getfullargspec(featurizer.get_feature_names).args)
+        if arg_no == 1:
+            return featurizer.get_feature_names()
+        elif arg_no == 2:
+            return featurizer.get_feature_names(feature_names)
+    # Featurizer doesn't have 'get_feature_names' or has atypical 'get_feature_names'
+    try:
+        # Get feature names using featurizer
+        dummy_X = np.empty((1, len(feature_names)))
```
kbattocchi, line 597: I think np.empty uses uninitialized memory so it might be better to use something like np.zeros in case the uninitialized memory happens to evaluate to NaN or infinity in some place and the transformer can't handle that.

Also it seems a bit weird that the type of the array that we're passing might not match what the transformer was trained on at all (e.g. if it was trained on strings and we pass numbers), but perhaps there's nothing better that we can do given that we're in an unusual situation anyway.

kbattocchi (approved):
Looks good; I added one suggestion but nothing major.

heimengqi:
https://github.com/microsoft/EconML/pull/419#discussion_r583969973 
This comment should be addressed in your PR.

econml/utilities.py
moprescu: I was hoping `np.empty` had random values so we don't run into issues at constant values. For example, `log(x)` has issues as 0, `1/x-1` has issues at 1. I guess the best is to pass in an array of `np.ones`, I will change it to that. For the future, maybe our estimators can record the featurized dimensions after fitting just like to record `d_x` now, but that required more complex changes. I will leave that for another PR.

econml/utilities.py
moprescu: Good point! I think i will use `feat(X)` as a prefix for people to understand that it's the featurizer applied to `X` and avoid naming confusion.

Issue #432, opened 2021-03-17
ForestDRLearner DoWhy Fit not working


jbel1026:
Hello,
I am trying to implement forestdrlearner.dowhy.fit() and it is throwing a recursion error, this method has been working on all other econml estimators. Is this a known issue? Can this be looked into? 
The propensity and regression models are both XGBoost 
I am using econml V. 0.9.2

![image](https://user-images.githubusercontent.com/80842968/111528510-9ce2ea80-8737-11eb-9d6f-1081555be205.png)

jbel1026:
Additonally, I get a different error when trying to refute the estimate using the DoWhyWrapper
thanks!
![image](https://user-images.githubusercontent.com/80842968/111673771-b183b900-87f1-11eb-8157-8a1e93bcfed0.png)

heimengqi:
Hi @jbel1026 , thanks for reaching out! I am working on fixing the first issue you are raising, that's related to our deprecated args in `ForestDRLearner`, I will have PR solving this issue today. 

For the second one, are you using `DRLearner`?

jbel1026:
Hello @heimengqi , thanks so much for the quick reply with this, both examples are using the `ForestDRLearner`

heimengqi:
I see, since you didn't fit successfully with `ForestDRLearner`, you can't call `refute_estimate` then, you should do it in a sequential way.

jbel1026:
@heimengqi I figured as much, thanks!

jbel1026:
@heimengqi Thanks for all your help, should I be able to run this now by just reinstalling V0.9.2 and running the commands?

heimengqi:
We just released V0.10.0 yesterday, which included this fix. Please upgrade to the new version, it will solve this issue. Thanks!

PR #434, opened 2021-03-18
Fix deprecated args and dowhy wrapper

heimengqi:
- Fix the deprecated args in ForestDRLearner, this will fix the issue #432 
- Ignore the "deprecated" args when get the parameters for cate estimator under dowhy wrapper
- Fix the feature name bug, should save the feature names from input data frame before transferring it to numpy arrays.


econml/dr/_drlearner.py
```diff
@@ -1523,7 +1523,7 @@ def n_crossfit_splits(self, value):
 
     @property
     def criterion(self):
-        return self.criterion
+        return "mse"
```
kbattocchi, line 1526: Seems a bit weird that we allow `est.criterion = 'deprecated'` but not `est.criterion = 'mse'`, and yet we return `"mse"` here...

If this is identical to sklearn's approach then I guess we can stick with it, but seems unintuitive to me.

econml/dr/_drlearner.py
```diff
@@ -1533,7 +1533,7 @@ def criterion(self, value):
 
     @property
     def max_leaf_nodes(self):
-        return self.max_leaf_nodes
+        return None
```
kbattocchi, line 1536: Parallel comment here.

econml/dowhy.py
```diff
@@ -106,30 +110,41 @@ def fit(self, Y, T, X=None, W=None, Z=None, *, outcome_names=None, treatment_nam
         -------
         self
         """
-
-        Y, T, X, W, Z = check_input_arrays(Y, T, X, W, Z)
```
kbattocchi, line 110: Can you add a test that reflects this fix?

kbattocchi (changes requested):
Please add tests for all bugs fixed.

econml/dr/_drlearner.py
heimengqi: Not exactly the same approach, they have a decorator for deprecated args and the deprecated args are not defined in init. Or can I just remove these deprecated args? since it should be removed for next release anyway?

econml/dowhy.py
heimengqi: Done

econml/dr/_drlearner.py
heimengqi: @kbattocchi @vsyrgkanis Any final conclusion on this? Then I could make it ready to merge.

kbattocchi: How hard would it be to mimic sklearn's approach?  I think it would be best to either do that or remove the deprecated args as you suggested.

kbattocchi (approved):
LGTM

econml/dr/_drlearner.py
heimengqi: I removed all the deprecated args, also for other classes, especially clean up the n_splits/n_crossfit_splits. It's not so clear to me how to mirror sklearn, they seem just add a warning, might worth create an issue to explore how to handle that systematically.

PR #441, opened 2021-03-25
Vasilis/casestudy

### LINKED ISSUE #437: Error in customer scenario notebook
In cell [16] of the multi-investment attribution notebook:
https://github.com/microsoft/EconML/blob/7005b242531669854f9e0b6b944b590b9b4fdf66/notebooks/CustomerScenarios/Case%20Study%20-%20Multi-investment%20Attribution%20at%20A%20Software%20Company.ipynb

I believe it should be `X.shape[0]` in both places.

vsyrgkanis:
Fixed small mistakes in multi-investment attribution case study (fixes #437).
Changed how aggregation happens in policy ensembles to follow a "voting" paradigm, where the treatment with the most votes is chosen.
Added a `predict_proba` method to policy learners.


econml/tests/test_policy_forest.py
```diff
@@ -290,6 +290,8 @@ def test_non_standard_input(self,):
         forest = PolicyForest(n_estimators=20, n_jobs=1, random_state=123).fit(X, y)
         pred = forest.predict(X)
         pred_val = forest.predict_value(X)
+        pred_prob = forest.predict_proba(X)
+        np.testing.assert_allclose(pred_prob.shape, (X.shape[0], 2))
```
kbattocchi, line 294: Shapes should be equal, not merely close :)
```suggestion
        assert pred_prob.shape == (X.shape[0], 2)
```

econml/policy/_forest/_tree.py
```diff
@@ -261,6 +261,30 @@ def predict(self, X, check_input=True):
...
+        Returns
+        -------
+        treatment_proba : array-like of shape (n_samples, n_treatments)
+            The probability of each treatment recommendation
+        """
+        check_is_fitted(self)
+        X = self._validate_X_predict(X, check_input)
+        pred = self.tree_.predict(X)
+        proba = np.zeros(pred.shape)
+        proba[np.arange(X.shape[0]), np.argmax(pred, axis=1)] = 1
```
kbattocchi, line 285: Here and everywhere else argmax is used, do we care at all about what happens in the case of ties?

econml/policy/_forest/_forest.py
```diff
@@ -467,6 +504,9 @@ def predict(self, X):
         Returns
         -------
         treatment : array-like of shape (n_samples)
-            The recommded treatment, i.e. the treatment index with the largest reward for each sample
+            The recommded treatment, i.e. the treatment index with the largest expected reward for each sample.
```
kbattocchi, line 507: Not sure about phrasing...  Maybe something more like 
```suggestion
            The recommded treatment, i.e. the treatment index most often predicted to have the highest reward for each sample.
```

econml/policy/_forest/_forest.py
```diff
@@ -455,6 +455,43 @@ def predict_value(self, X):
...
+
+        # Assign chunk of trees to jobs
+        n_jobs, _, _ = _partition_estimators(self.n_estimators, self.n_jobs)
+
+        # avoid storing the output of every estimator by summing them here
+        y_hat = np.zeros((X.shape[0], self.n_outputs_), dtype=np.float64)
+
+        # Parallel loop
+        lock = threading.Lock()
+        Parallel(n_jobs=n_jobs, verbose=self.verbose, backend='threading', require="sharedmem")(
```
kbattocchi, line 487: `require="sharedmem"` implies `backend='threading'`; see https://joblib.readthedocs.io/en/latest/parallel.html which says:

> It is not recommended to hard-code the backend name in a call to Parallel in a library. Instead it is recommended to set soft hints (prefer) or hard constraints (require) so as to make it possible for library users to change the backend from the outside using the parallel_backend context manager.
 ```suggestion
        Parallel(n_jobs=n_jobs, verbose=self.verbose, require="sharedmem")(

```

File: econml/policy/_drlearner.py [hunk 5]
```diff
@@ -98,6 +98,24 @@ def predict_value(self, X):
         """
         return self.drlearner_.const_marginal_effect(X)
 
+    def predict_proba(self, X):
+        """ Predict the probability of recommending each treatment
+
+        Parameters
+        ----------
+        X : {array-like} of shape (n_samples, n_features)
```
kbattocchi, line 106: Why the curly brackets?
```suggestion
        X : array-like of shape (n_samples, n_features)

```

kbattocchi (approved):
I've made a few minor suggestions, but this looks good to me.

econml/policy/_forest/_tree.py
vsyrgkanis: nope. we need to return on. So we could just defer to what argmax does.

PR #446, opened 2021-03-31
Implement DynamicDML

moprescu:
* DynamicDML with all the OrthoLearner functionality
* API tests similar to the DML ones


econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,660 @@
...
+    @property
+    def featurizer_(self):
+        # NOTE This is used by the inference methods and has to be the overall featurizer. intended
+        # for internal use by the library
+        return self.ortho_learner_model_final_._model_final_trained[0]._featurizer
+
+    @property
+    def model_final_(self):
+        # NOTE This is used by the inference methods and is more for internal use to the library
+        #      We need to use the rlearner's copy to retain the information from fitting
```
kbattocchi, line 650:
```suggestion
        #      We need to use the ortho_learner's copy to retain the information from fitting
```

econml/tests/dgp.py
```diff
@@ -0,0 +1,156 @@
...
+    def __init__(self, n_periods, n_treatments, n_x):
+        self.n_periods = n_periods
+        self.n_treatments = n_treatments
+        self.n_x = n_x
+        return
+
+    def create_instance(self, *args, **kwargs):
+        pass
+
+    def _gen_data_with_policy(self, n_units, policy_gen, random_seed=123):
```
kbattocchi, line 19: Looks like maybe this and `create_instance` should be abstract?

kbattocchi (commented):
Here are some comments - please address what you're able to.

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,737 @@
...
+        A tuple (y_res, T_res, X, W), of the residuals from the first stage estimation
+        along with the associated X and W. Samples are not guaranteed to be in the same
+        order as the input order.
+        """
+        if not hasattr(self, '_cached_values'):
+            raise AttributeError("Estimator is not fitted yet!")
+        if self._cached_values is None:
+            raise AttributeError("`fit` was called with `cache_values=False`. "
+                                 "Set to `True` to enable residual storage.")
+        Y_res, T_res = self._cached_values.nuisances
```
heimengqi, line 736: residuals for Y and T are not correct here and it's complicated...we fit nuisance iteratively on different subgroup and it's not clear to me how to store all the nuisances but still inherit through `_OrthoLearner` we might need a dictionary format to store the residuals but that's not the case for the other classes.

heimengqi (changes requested):

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,660 @@
...
+                T_pred = self._model_t_trained[kappa][tau].predict(
+                    self._filter_or_None(X, inds_predict + kappa),
+                    self._filter_or_None(W, inds_predict + kappa))
+                T_res[np.arange(Y.shape[0]) % self.n_periods == kappa, ..., tau] = T_slice\
+                    - shape_formatter(T_slice, T_pred).reshape(T_slice.shape)
+        return Y_res, T_res
+
+    def score(self, Y, T, X=None, W=None, sample_weight=None, groups=None):
+        # TODO: implement scores
+        # TODO: fix correctness?
```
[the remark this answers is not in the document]
moprescu: Implemented now!

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,660 @@
...
+    def bias_part_of_coef(self):
+        return self.ortho_learner_model_final_._model_final._fit_cate_intercept
+
+    @property
+    def fit_cate_intercept_(self):
+        return self.ortho_learner_model_final_._model_final._fit_cate_intercept
+
+    @property
+    def original_featurizer(self):
+        # NOTE: important to use the rlearner_model_final_ attribute instead of the
```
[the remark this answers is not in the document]
moprescu: Fixed!

econml/tests/dgp.py
moprescu: Fixed!

econml/dml/dynamic_dml.py
moprescu: Fixed!

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,660 @@
...
+        else:
+            T_score = None
+        return Y_score, T_score
+
+    def _get_shape_formatter(self, X, W):
+        if (X is None) and (W is None):
+            return lambda x, x_pred: np.tile(x_pred.reshape(1, -1), (x.shape[0], 1))
+        return lambda x, x_pred: x_pred
+
+    def _filter_or_None(self, X, filter_idx):
```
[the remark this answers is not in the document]
moprescu: Fixed!

econml/tests/dgp.py
moprescu: Fixed!

moprescu: Fixed!

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,660 @@
...
+                    XT_res[period][self.n_periods - 1 - tau].reshape(-1, d_xt, 1) @ cur_resT.reshape(-1, 1, d_xt),
+                    axis=0)
+                M[kappa * d_xt:(kappa + 1) * d_xt,
+                  tau * d_xt:(tau + 1) * d_xt] = m_kappa_tau
+        return np.linalg.inv(M) @ Sigma @ np.linalg.inv(M).T
+
+
+class _DynamicFinalWrapper(_FinalWrapper):
+
+    def predict_with_res(self, X, T_res):
```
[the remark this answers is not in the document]
moprescu: It's because the predicted Y_res depends on the featurization X &#x2297; T_res which only happens in the wrapper.

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,660 @@
...
+                 random_state=None):
+        self.fit_cate_intercept = fit_cate_intercept
+        self.linear_first_stages = linear_first_stages
+        self.featurizer = clone(featurizer, safe=False)
+        self.model_y = clone(model_y, safe=False)
+        self.model_t = clone(model_t, safe=False)
+        super().__init__(discrete_treatment=discrete_treatment,
+                         discrete_instrument=False,
+                         categories=categories,
+                         cv=GroupKFold(cv) if isinstance(cv, int) else cv,
```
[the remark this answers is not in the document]
moprescu: The user can pass either an int or a cv object. If it's an int, we use GroupKFold (because it's a group split). I don't delegate this to `_OrthoLearner` because of some conflict with stratification when `T` is categorical.

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,660 @@
...
+            and in index form for tau. For example, the residuals for (kappa, tau) can be retrieved via
+            T_res[np.arange(n) % n_periods == kappa, ..., tau]. For tau < kappa, the entries of this
+            matrix are np.nan.
+            This shape is required for _OrthoLearner's crossfitting.
+        """
+        assert Y.shape[0] % self.n_periods == 0, \
+            "Length of training data should be an integer multiple of time periods."
+        inds_predict = np.arange(Y.shape[0])[np.arange(Y.shape[0]) % self.n_periods == 0]
+        Y_res = np.full(Y.shape, np.nan)
+        T_res = np.full(T.shape + (self.n_periods, ), np.nan)
```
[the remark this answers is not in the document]
moprescu: Done!

econml/_ortho_learner.py
```diff
@@ -770,18 +771,19 @@ def _fit_nuisances(self, Y, T, X=None, W=None, Z=None, sample_weight=None, group
...
         self.score_ = None
         if hasattr(self._ortho_learner_model_final, 'score'):
             self.score_ = self._ortho_learner_model_final.score(Y, T, **filter_none_kwargs(X=X, W=W, Z=Z,
                                                                                            nuisances=nuisances,
-                                                                                           sample_weight=sample_weight)
-                                                                )
+                                                                                           sample_weight=sample_weight,
+                                                                                           groups=groups))
 
     def const_marginal_effect(self, X=None):
```
vsyrgkanis, line 788: when I call cons_marginal_effecT_inference I get this bad printing:
image

moprescu: It's a weird effect from Pandas. You can get the same effect with:
```Python
pd.DataFrame({"a":["$(T0)_0$", "$(T1)_0$"]})
```
However, this:
```Python
pd.DataFrame({"a":["(T0)$_0$", "(T1)$_0$"]})
```
looks ok. I will change it to this.

vsyrgkanis:
Could you also add a snippet of code in the README, as it seems a substantive addition to advertise!

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,660 @@
...
+                self._model_final_trained[kappa].predict(X).reshape(
+                x_dy_shape + (d_t, )
+            )
+        return preds
+
+    def score(self, Y, T, X=None, W=None, Z=None, nuisances=None, sample_weight=None, sample_var=None):
+        # TODO: implement score
+        return None
+
+    def period_filter_gen(self, p, n):
```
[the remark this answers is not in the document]
moprescu: Done!

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,660 @@
...
+        Y_res = np.full(Y.shape, np.nan)
+        T_res = np.full(T.shape + (self.n_periods, ), np.nan)
+        shape_formatter = self._get_shape_formatter(X, W)
+        for kappa in np.arange(self.n_periods):
+            Y_slice = Y[inds_predict + self.n_periods - 1]
+            Y_pred = self._model_y_trained[kappa].predict(
+                self._filter_or_None(X, inds_predict + kappa),
+                self._filter_or_None(W, inds_predict + kappa))
+            Y_res[np.arange(Y.shape[0]) % self.n_periods == kappa] = Y_slice\
+                - shape_formatter(Y_slice, Y_pred).reshape(Y_slice.shape)
```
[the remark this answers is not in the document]
moprescu: Fixed!

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,660 @@
...
+                  tau * d_xt:(tau + 1) * d_xt] = m_kappa_tau
+        return np.linalg.inv(M) @ Sigma @ np.linalg.inv(M).T
+
+
+class _DynamicFinalWrapper(_FinalWrapper):
+
+    def predict_with_res(self, X, T_res):
+        fts = self._combine(X, T_res, fitting=False)
+        prediction = self._model.predict(fts)
+        if self._intercept is not None:
```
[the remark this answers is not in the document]
moprescu: This is a replica of `predict` of the same model, when residuals are not passed in. So the intercept logic is replicated here.

notebooks/Dynamic Double Machine Learning Examples.ipynb
```diff
@@ -0,0 +1,692 @@
...
+     "output_type": "stream",
+     "text": [
+      "Effect of target policy over baseline policy: 4.33\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Effect of target policy over baseline policy\n",
+    "# Must specify a treatment for each period\n",
+    "baseline_policy = np.zeros((1, n_periods))\n",
```
vsyrgkanis, line 236: Since the notebook can specify the treamtent dimension in the beginning it would be great if the whole notebook works when I set n_treatments > 1

Currently the code breaks

vsyrgkanis (approved):

vsyrgkanis (commented):
The text that lies under the table in the summary results is good for a static setting, but might not be the best for the summary results displayed in the dynamic setting. So we might want to not show that text here in the summary

vsyrgkanis:
In the plot of effects in the notebook, is theta_0 the effect from treatments in period 0 or the effect of treatment at lag 0 from the outcome. Similarly for the rest of the params.

vsyrgkanis:
In the notebook when you say: "number of panels" do you mean "number of units"

vsyrgkanis:
The results in the notebook, even with 5000 units are very poor! We need to investigate this before we push:
image

vsyrgkanis:
There is definitely a bug somewhere. 

First: unlike how you describe it (T0)_0 is the effect from period 0, it's actually the effect from the lag 0 period, i.e. from the last period'; we might want to change this if you want, but currently that's what you compute).

Second: The results from dynamic dml for the last period parameters (which I believe are (T0)_0), should exactly match the results if you had run simple LinearDML solely on data from the last period.

But what I get from DynamicMDL is:
image

and what I get from LinearDML with data solely from the last period is:
image

The truth is 0.7, so LinearDML gets the correct thing, but DynamicDML gets the wrong thing.

vsyrgkanis:
Results are almost identical when n_periods =1
DynamicDML
image
LinearDML
image


So there is something when we have multiple periods that gets messed up

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,660 @@
...
+    and at predict time returns :math:`\\theta(X)`. The score method returns the MSE of this final
+    residual on residual regression.
+    Assumes model final is parametric with no intercept.
+    """
+    # TODO: update docs
+
+    def __init__(self, model_final, n_periods):
+        self._model_final = model_final
+        self.n_periods = n_periods
+        self._model_final_trained = {k: clone(self._model_final, safe=False) for k in np.arange(n_periods)}
```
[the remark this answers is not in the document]
moprescu: Some of the details of the models are retrieved before fit time in downstream logic.

vsyrgkanis:
Oh!! It was just the parameters of lassocv you set in the notebook that mess thigns up. So just leae the default tol and the default alpha grid. That improves performance significantly 
image

and also matches the LinearDML.

Also I would say we might want to reverse the parameters to mean what you are descrbing and not the lag version that was in the prototype. 

So the (T0)_0 should mean the effect of period 0 on final outcome and (T0)_m the effect of final period. 

Currrently it's the reverse and that messes also the "effect(X, T0, T1)"
here one needs to give T0 and T1 in the reverse order too.

moprescu:
82d9badc4ca.png)



> Oh!! It was just the parameters of lassocv you set in the notebook that mess thigns up. So just leae the default tol and the default alpha grid. That improves performance significantly
> image
> 
> and also matches the LinearDML.
> 
> Also I would say we might want to reverse the parameters to mean what you are descrbing and not the lag version that was in the prototype.
> 
> So the (T0)_0 should mean the effect of period 0 on final outcome and (T0)_m the effect of final period.
> 
> Currrently it's the reverse and that messes also the "effect(X, T0, T1)"
> here one needs to give T0 and T1 in the reverse order too.

Fixed!

econml/tests/dgp.py
```diff
@@ -0,0 +1,166 @@
...
+            self.epsilon.reshape(1, -1))
+        for t in np.arange(self.n_periods - 2, -1, -1):
+            # Invert indices to match latest API
+            self.true_hetero_effect[t, :] = cross_product(
+                add_constant(self.x_hetero_effect.reshape(1, -1), has_constant='add'),
+                self.zeta.reshape(1, -1) @ np.linalg.matrix_power(
+                    self.Beta, (self.n_periods - 1 - t) - 1) @ self.Alpha)
+        return self
+
+    def hetero_effect_fn(self, t, x):
```
vsyrgkanis, line 115: I think the dgp still allows heterogneeity with respect to the time -varying X, which will be incompatible wiht what the method asssumes, which is only wrt X[0].

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,789 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
vsyrgkanis, line 1: Given that we plan on adding other approaches too for dynamic panel learning (e.g. a doubly robust one for discrete treatments and other time-varying methods), it might be better to place `dynamic_dml` in a new module: `dynamic`, e.g.
`from econml.dynamic import DynamicDML`
or
`from econml.dynamic.dml import DynamicDML`

econml/dml/dynamic_dml.py
```diff
@@ -0,0 +1,789 @@
...
+
+        ...etc.
+
+        Parameters
+        ----------
+        Y: (n, d_y) matrix or vector of length n
+            Outcomes for each sample (required: n = n_groups * n_periods)
+        T: (n, d_t) matrix or vector of length n
+            Treatments for each sample (required: n = n_groups * n_periods)
+        X: optional(n, d_x) matrix or None (Default=None)
```
vsyrgkanis, line 568: It might be good to mention here and in the beginning of the docstring that heterogeneity is measured only with respect to the period[0] value of X and the rest of the X's are just used as time-varying controls. Since this is a complex part of the API.

vsyrgkanis (approved):

econml/tests/dgp.py
vsyrgkanis: my bad, you're doing this correctly

econml/dml/dynamic_dml.py
vsyrgkanis: Done

econml/dml/dynamic_dml.py
vsyrgkanis: Done

notebooks/Dynamic Double Machine Learning Examples.ipynb
vsyrgkanis: Done

PR #447, opened 2021-04-01
Add causal analysis solution

kbattocchi:


econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1023 @@
...
+                                      discrete_treatment=discrete_treatment,
+                                      random_state=123)
+                est.tune(y, T, X=X_xf, W=W)
+            est.fit(y, T, X=X_xf, W=W, cache_values=True)
+
+            # effect doesn't depend on W, so only pass in first row
+            global_inference = est.const_marginal_effect_inference(X=X_xf[0:1])
+            # For the local model, change the featurizer to include X
+            est.featurizer = PolynomialFeatures(degree=1, include_bias=True)
+            est.refit_final()
```
kbattocchi, line 451: @vasilismsr If I understand correctly, you were suggesting that it might be better to do the following instead?

```suggestion
            est.featurizer = PolynomialFeatures(degree=1, include_bias=True)
            est.fit(y, T, X=X_xf, W=W, cache_values=True)

            global_inference = est.ate_inference(X=X_xf)
```

vsyrgkanis: only if heterogeneity = 'linear'
if heterogeneity = 'forest', we should do: est.ate__inference() 
which is the doubly robust variance that runs an average over the training data and has the doubly robust correction. 
This way we always have non-conservative intervals for the global effects.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+        (for classification) are used. If 'automl', then a kfold cross-validation and model selection is performed
+        among several models and the best is chosen.
+        TODO. Add other options, such as {'azure_automl', 'forests', 'boosting'} that will use particular sub-cases
+        of models or also integrate with azure autoML. (post-MVP)
+    heterogeneity_model: one of {'linear', 'forest'}, optional (default='linear')
+        What type of model to use for treatment effect heterogeneity. 'linear' means that a heterogeneity model
+        of the form theta(X)=<a, X> will be used, while 'forest' means that a forest model will be trained instead.
+        TODO. Add other options, such as {'automl'} for performing
+        model selection for the causal effect, or {'sparse_linear'} for using a debiased lasso. (post-MVP)
+    automl: bool, default True
```
vsyrgkanis, line 263: This option should be removed from docstring

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+                est = CausalForestDML(model_y=self._model_y,
+                                      model_t=model_t,
+                                      featurizer=featurizer,
+                                      discrete_treatment=discrete_treatment,
+                                      random_state=123)
+                est.tune(y, T, X=X_xf, W=W)
+            est.fit(y, T, X=X_xf, W=W, cache_values=True)
+
+            # effect doesn't depend on W, so only pass in first row
+            global_inference = est.const_marginal_effect_inference(X=X_xf[0:1])
```
vsyrgkanis, line 448: This needs to be altered to the new logic about global inference, right?

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+                                fit_cate_intercept=False,
+                                linear_first_stages=False,
+                                random_state=123)
+            else:
+                est = CausalForestDML(model_y=self._model_y,
+                                      model_t=model_t,
+                                      featurizer=featurizer,
+                                      discrete_treatment=discrete_treatment,
+                                      random_state=123)
+                est.tune(y, T, X=X_xf, W=W)
```
vsyrgkanis, line 444: Tune doesn't seem to have been fixed and so this will through an error when the inputs are pandas dataframes

vsyrgkanis (changes requested):

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+        model selection for the causal effect, or {'sparse_linear'} for using a debiased lasso. (post-MVP)
+    automl: bool, default True
+        Whether to automatically perform model selection over a variety of models
+    n_jobs: int, default -1
+        Degree of parallelism to use when training models via joblib.Parallel
+    """
+
+    _results = namedtuple("_results", field_names=[
+        "feature_index", "feature_name", "feature_baseline", "feature_levels", "hinds",
+        "X_transformer", "W_transformer", "estimator", "global_inference", "d_t"])
```
vsyrgkanis, line 271: We might want to remove d_t if `_results.d_t` not used anywhere

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+                                             hinds=hinds,
+                                             X_transformer=X_transformer,
+                                             W_transformer=W_transformer,
+                                             estimator=est,
+                                             d_t=d_t,
+                                             global_inference=global_inference)
+
+            return insights, result
+
+        if self.feature_names is None:
```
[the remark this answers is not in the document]
vsyrgkanis: If we are in a warm start why are we not using the same feature names as in the first time. 

We could be getting some weirdness here if user passes once a numpy array and then a pandas data frame with warm start

vsyrgkanis (changes requested):

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Done.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Yes, it will be in the next update.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+                                                      heterogeneity_inds[ind] != self._cache[ind][1].hinds)]
+        else:
+            new_inds = train_inds
+
+            self._cache = {}  # store mapping from feature to insights, results
+
+            # train the Y model
+
+            # perform model selection for the Y model using all X, not on a per-column basis
+            self._x_transform = ColumnTransformer([('encode',
```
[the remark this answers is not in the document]
kbattocchi: Done

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+            # perform model selection
+            model_t = (_first_stage_clf(WX, T, automl=self.nuisance_models == 'automl')
+                       if discrete_treatment else _first_stage_reg(WX, T, automl=self.nuisance_models == 'automl'))
+
+            # For the global model, use a constant featurizer to fit the ATE
+            # Ideally, this could be PolynomialFeatures(degree=0, include_bias=True), but degree=0 is unsupported
+            # So instead we'll use our own class
+            featurizer = _ConstantFeatures()
+
+            # TODO: support other types of heterogeneity via an initializer arg
```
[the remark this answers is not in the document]
kbattocchi: Done

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+            heterogeneity_inds = [heterogeneity_inds for _ in train_inds]
+
+        # replace None elements of heterogeneity_inds and ensure indices are numeric
+        heterogeneity_inds = {ind: list(range(X.shape[1])) if hinds is None else _get_column_indices(X, hinds)
+                              for ind, hinds in zip(train_inds, heterogeneity_inds)}
+
+        if warm_start and self.nuisance_models != self.nuisance_models_:
+            warnings.warn("warm_start will be ignored since the nuisance models have changed"
+                          f" from {self.nuisance_models_} to {self.nuisance_models} since the previous call to fit")
+            new_inds = train_inds
```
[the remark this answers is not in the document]
kbattocchi: Fixed

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+                                featurizer=featurizer,
+                                discrete_treatment=discrete_treatment,
+                                fit_cate_intercept=False,
+                                linear_first_stages=False,
+                                random_state=123)
+            else:
+                est = CausalForestDML(model_y=self._model_y,
+                                      model_t=model_t,
+                                      featurizer=featurizer,
+                                      discrete_treatment=discrete_treatment,
```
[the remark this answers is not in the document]
kbattocchi: Done

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+        if self.feature_names is None:
+            if hasattr(X, "iloc"):
+                feature_names = X.columns
+            else:
+                feature_names = [f"x{i}" for i in range(X.shape[1])]
+
+        self.feature_names_ = feature_names
+
+        # extract subset matching new columns
+        feature_names = _safe_indexing(feature_names, new_inds)
```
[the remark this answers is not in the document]
kbattocchi: Done

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+            :Index: ['feature', 'feature_value']
+            :Rows: For each feature that is numerical, we have an entry with index ['{feature_name}', 'num'], where
+                    'num' is literally the string 'num' and feature_name is the input feature name.
+                    For each feature that is categorical, we have an entry with index ['{feature_name}',
+                    '{cat}v{base}'] where cat is the category value and base is the category used as baseline.
+                    If all features are numerical then the feature_value index is dropped in the dataframe, but not
+                    in the serialized dict.
+        """
+        # a global inference indicates the effect of that one feature on the outcome
+        return self._pandas_summary(lambda res: res.global_inference, props=self._point_props(alpha), n=1)
```
vsyrgkanis, line 682: This requires some change if global_inference results is either a pop summary or a point estimate summary dependent on the heterogeneity model

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+        Gets the local feature importance as dictionary
+
+        Dictionary entries for predictions, etc. will be nested lists of shape (n_rows, d_y, sum(d_t))
+
+        Only for serialization purposes to upload to AzureML
+        """
+        return self._dict_summary(self._local_effect_inference(Xtest), props=self._point_props(alpha),
+                                  kind='local')
+
+    def _check_feature_index(self, X, feature_index):
```
vsyrgkanis, line 797: WE might want to rename this because this seems like a safe accessing of the _result for some feature index.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+
+        (numeric_index,) = _get_column_indices(X, [feature_index])
+        results = [res for res in self._results
+                   if res.feature_index == numeric_index]
+
+        assert len(results) != 0, f"The feature index supplied was not fitted"
+        (result,) = results
+        return result
+
+    def whatif(self, X, Xnew, feature_index, y):
```
vsyrgkanis, line 813: Add an alpha parameter to whatif for confidence level

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+        result = self._check_feature_index(X, feature_index)
+        inf = result.estimator.effect_inference(
+            X=result.X_transformer.transform(X), T0=T0, T1=T1)
+
+        # we want to offset the inference object by the baseline estimate of y
+        inf.translate(y)
+
+        return inf
+
+    def _whatif_dict(self, X, Xnew, feature_index, y):
```
vsyrgkanis, line 852: Add an alpha to the dict here too

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+        y: array-like
+            Observed labels or outcome of a predictive model for baseline y values
+
+        Returns
+        -------
+        dict : dict
+            The counterfactual predictions, as a dictionary
+        """
+
+        inf = self.whatif(X, Xnew, feature_index, y)
```
vsyrgkanis, line 876: use alphs

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+        props = self._point_props(0.05)
```
vsyrgkanis, line 877: use alpha

vsyrgkanis (changes requested):

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+        return make_pipeline(StandardScaler(), Lasso(alpha=model.steps[1][1].alpha_))
+
+
+def _first_stage_clf(X, y, *, make_regressor=False, automl=True):
+    if automl:
+        model = GridSearchCVList([make_pipeline(StandardScaler(), LogisticRegression()),
+                                  RandomForestClassifier(
+                                      n_estimators=100, random_state=123),
+                                  GradientBoostingClassifier(random_state=123)],
+                                 param_grid_list=[{'C': [0.01, .1, 1, 10, 100]},
```
heimengqi, line 131: since it's wrapper by a pipeline, parameter C should be logisticregression__C

heimengqi (changes requested):

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Done

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Good catch, thanks!  Fixed in my next push.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: This logic has been fixed as suggested.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1038 @@
...
+            res.update([(key, value.tolist()) for key, value in props])
+            return {**self._shared, **res}
+
+        return self._summarize(summary=make_dict,
+                               get_inference=get_inference,
+                               props=props,
+                               expand_arr=expand_arr,
+                               drop_sample=drop_sample)
+
+    def global_causal_effect(self, alpha=0.05):
```
kbattocchi, line 647: @vasilismsr I think this default for alpha was taken from the prototype, but we use different values for alpha in different places in the library; most of our inference methods use alpha=0.1; however grf and the _StatsModelsWrapper use alpha=0.05.  Should those places (and this class) be made to default to alpha=0.1 to be consistent with the rest of the package?

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: I've added alpha here.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: It turns out not to make sense to add alpha to the regular `whatif` method because the output is just an InferenceResults object that can use any alpha when getting the pvalues, etc. and does not have an alpha baked in.

However, in this method we need to commit to one because the dictionary itself contains the confidence intervals based on a specific confidence level.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: done

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: I mentioned it in a different comment, but alpha doesn't belong here because the result is an InferenceResults object that doesn't use a pre-specified confidence level - its methods can take any alpha and return the corresponding pvalue, zstat, or confidence interval.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1044 @@
...
+        else:
+            feature_names = self.feature_names
+
+        self.feature_names_ = feature_names
+
+        # extract subset of names matching new columns
+        new_feat_names = _safe_indexing(feature_names, new_inds)
+
+        cache_updates = dict(zip(new_inds,
+                                 joblib.Parallel(n_jobs=self.n_jobs,
```
vsyrgkanis, line 486: Most prob we should be using a minimum of n_jobs and the number of new_inds.

vsyrgkanis (changes requested):

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: hm I run it with 1 job and it still was saying: starting joblib with 8 concurrent workers.

vsyrgkanis:
I tried the following snippet:
```python
A = np.hstack([T.reshape(-1, 1), X, W])
Xcols = [f'X{i}' for i in range(n_x)] 
Wcols = [f'W{i}' for i in range(n_w)]
Adf = pd.DataFrame(A, columns=['T'] + Xcols + Wcols)
ca = CausalAnalysis(feature_inds=['T'], heterogeneity_inds=[Xcols], categorical=[])
ca.fit(Adf, Y)
ca.heterogeneity_model = 'forest'
ca.fit(Adf, Y, warm_start=True)
```python

And it did not refit the model, even though I changed the type of heterogeneity model to be used.

**Comment by vsyrgkanis** [comment]
If we warm start without having fitted before, then the error message is cryptic and not: "can't warm start without having fitted". In fact it's not clear to me why warm_start in the first fit should be raising an error. If we have not fitted before, then warm_start=True should behave as if it was false with no error:

```python
ca = CausalAnalysis(feature_inds=['T'], heterogeneity_inds=[Xcols], categorical=[])
ca.fit(Adf, Y, warm_start=True)
```
AttributeError                            Traceback (most recent call last)
<ipython-input-53-393ffab97503> in <module>
----> 1 ca.fit(Adf, Y, warm_start=True)

c:\users\vasy\documents\econml\econml\solutions\causal_analysis\_causal_analysis.py in fit(self, X, y, warm_start)
    299                     "This object has not been fit, so warm_start does not apply")
    300 
--> 301             if self._d_x != X.shape[1]:
    302                 raise ValueError(
    303                     f"Can't warm start: previous X had {self._d_x} columns, new X has {X.shape[1]} columns")

AttributeError: 'CausalAnalysis' object has no attribute '_d_x'

vsyrgkanis:
Can we maybe fix this warning from LightGBM when we set it to automl: 
[LightGBM] [Warning] Accuracy may be bad since you didn't explicitly set num_leaves OR 2^max_depth > num_leaves. (num_leaves=31).

Most prob we need to set some better defaults or a different grid

vsyrgkanis:
If we set an incompatible input for heterogeneity inds, the error message is not user friendly:

```python
ca = CausalAnalysis(feature_inds=['T', 'X0'],
                    heterogeneity_inds=[Xcols],
                    categorical=[])
ca.fit(Adf, Y)
```

---------------------------------------------------------------------------
_RemoteTraceback                          Traceback (most recent call last)
_RemoteTraceback: 
"""
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\externals\loky\process_executor.py", line 431, in _process_worker
    r = call_item()
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\externals\loky\process_executor.py", line 285, in __call__
    return self.fn(*self.args, **self.kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\_parallel_backends.py", line 595, in __call__
    return self.func(*args, **kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\parallel.py", line 262, in __call__
    return [func(*args, **kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\parallel.py", line 262, in <listcomp>
    return [func(*args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\solutions\causal_analysis\_causal_analysis.py", line 371, in process_feature
    hinds = heterogeneity_inds[feat_ind]
KeyError: 1
"""

The above exception was the direct cause of the following exception:

vsyrgkanis:
This throws a weird error cannot underwstand what it is:

```python
A = np.hstack([T.reshape(-1, 1), X, W])
Xcols = [f'X{i}' for i in range(n_x)] 
Wcols = [f'W{i}' for i in range(n_w)]
Adf = pd.DataFrame(A, columns=['T'] + Xcols + Wcols)
ca = CausalAnalysis(feature_inds=['T', 'X0', 'W0'],
                    heterogeneity_inds=[Xcols, [], []],
                    categorical=[])
ca.fit(Adf, Y)
```
```
---------------------------------------------------------------------------
_RemoteTraceback                          Traceback (most recent call last)
_RemoteTraceback: 
"""
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\externals\loky\process_executor.py", line 431, in _process_worker
    r = call_item()
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\externals\loky\process_executor.py", line 285, in __call__
    return self.fn(*self.args, **self.kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\_parallel_backends.py", line 595, in __call__
    return self.func(*args, **kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\parallel.py", line 262, in __call__
    return [func(*args, **kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\parallel.py", line 262, in <listcomp>
    return [func(*args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\solutions\causal_analysis\_causal_analysis.py", line 422, in process_feature
    est.fit(y, T, X=X_xf, W=W, cache_values=True)
  File "c:\users\vasy\documents\econml\econml\utilities.py", line 1262, in m
    return to_wrap(*args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\dml\dml.py", line 659, in fit
    return super().fit(Y, T, X=X, W=W,
  File "c:\users\vasy\documents\econml\econml\utilities.py", line 1262, in m
    return to_wrap(*args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\dml\dml.py", line 505, in fit
    return super().fit(Y, T, X=X, W=W, sample_weight=sample_weight, freq_weight=freq_weight,
  File "c:\users\vasy\documents\econml\econml\utilities.py", line 1262, in m
    return to_wrap(*args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\dml\_rlearner.py", line 369, in fit
    return super().fit(Y, T, X=X, W=W,
  File "c:\users\vasy\documents\econml\econml\utilities.py", line 1262, in m
    return to_wrap(*args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\_cate_estimator.py", line 130, in call
    m(self, Y, T, *args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\_ortho_learner.py", line 601, in fit
    Y, T, X, W, Z, sample_weight, freq_weight, sample_var, groups = check_input_arrays(
  File "c:\users\vasy\documents\econml\econml\utilities.py", line 544, in check_input_arrays
    new_arg = check_array(arg, dtype=None, ensure_2d=False, accept_sparse=True,
  File "C:\ProgramData\Anaconda3\lib\site-packages\sklearn\utils\validation.py", line 72, in inner_f
    return f(**kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\sklearn\utils\validation.py", line 533, in check_array
    dtype_orig = np.result_type(*dtypes_orig)
  File "<__array_function__ internals>", line 5, in result_type
ValueError: at least one array or dtype is required
"""

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
<ipython-input-47-3bd9030e5f5c> in <module>
      2                     heterogeneity_inds=[Xcols, [], []],
      3                     categorical=[])
----> 4 ca.fit(Adf, Y)

c:\users\vasy\documents\econml\econml\solutions\causal_analysis\_causal_analysis.py in fit(self, X, y, warm_start)
    484 
    485         cache_updates = dict(zip(new_inds,
--> 486                                  joblib.Parallel(n_jobs=self.n_jobs,
    487                                                  verbose=1)(joblib.delayed(process_feature)(feat_name, feat_ind)
    488                                                             for feat_name, feat_ind in zip(new_feat_names, new_inds))))

C:\ProgramData\Anaconda3\lib\site-packages\joblib\parallel.py in __call__(self, iterable)
   1059 
   1060             with self._backend.retrieval_context():
-> 1061                 self.retrieve()
   1062             # Make sure that we get a last message telling us we are done
   1063             elapsed_time = time.time() - self._start_time

C:\ProgramData\Anaconda3\lib\site-packages\joblib\parallel.py in retrieve(self)
    938             try:
    939                 if getattr(self._backend, 'supports_timeout', False):
--> 940                     self._output.extend(job.get(timeout=self.timeout))
    941                 else:
    942                     self._output.extend(job.get())

C:\ProgramData\Anaconda3\lib\site-packages\joblib\_parallel_backends.py in wrap_future_result(future, timeout)
    540         AsyncResults.get from multiprocessing."""
    541         try:
--> 542             return future.result(timeout=timeout)
    543         except CfTimeoutError as e:
    544             raise TimeoutError from e

C:\ProgramData\Anaconda3\lib\concurrent\futures\_base.py in result(self, timeout)
    430                 raise CancelledError()
    431             elif self._state == FINISHED:
--> 432                 return self.__get_result()
    433 
    434             self._condition.wait(timeout)

C:\ProgramData\Anaconda3\lib\concurrent\futures\_base.py in __get_result(self)
    386     def __get_result(self):
    387         if self._exception:
--> 388             raise self._exception
    389         else:
    390             return self._result

ValueError: at least one array or dtype is required
```

vsyrgkanis:
Even simpler: It seems I cannot specify an empty heterogeneity list, which should be a valid option

```python
A = np.hstack([T.reshape(-1, 1), X, W])
Xcols = [f'X{i}' for i in range(n_x)] 
Wcols = [f'W{i}' for i in range(n_w)]
Adf = pd.DataFrame(A, columns=['T'] + Xcols + Wcols)
ca = CausalAnalysis(feature_inds=['T'],
                    heterogeneity_inds=[],
                    categorical=[])
ca.fit(Adf, Y)
```

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1044 @@
...
+
+    Parameters
+    ----------
+    feature_inds: array-like of int, str, or bool
+        The features for which to estimate causal effects, expressed as either column indices,
+        column names, or boolean flags indicating which columns to pick
+    categorical: array-like of int, str, or bool
+        The features which are categorical in nature, expressed as either column indices,
+        column names, or boolean flags indicating which columns to pick
+    heterogeneity_inds: array-like of int, str, or bool, or None or list of array-like elements or None, default None
```
vsyrgkanis, line 215: To avoid confusion I think we should make the default of heterogneity inds be 'all', which means use all variables, and None or empty list, should mean use no variables for heterogeneity.

vsyrgkanis (changes requested):

vsyrgkanis:
This code should be raising an error of incompatible input lengths but it doesn't
```python 
A = np.hstack([T.reshape(-1, 1), X, W])
Xcols = [f'X{i}' for i in range(n_x)] 
Wcols = [f'W{i}' for i in range(n_w)]
Adf = pd.DataFrame(A, columns=['T'] + Xcols + Wcols)
ca = CausalAnalysis(feature_inds=['T'],
                    heterogeneity_inds=[Xcols, Xcols, Xcols],
                    categorical=[])
ca.fit(Adf, Y)
```

econml/solutions/causal_analysis/_causal_analysis.py
heimengqi: if it's an empty list, forest option will not work? it requires X? and also all the local view will not work neither

vsyrgkanis:
It seems weird that when we have a single feature_ind, then the global causal effect is returned as:
image
with "value" as index, while when we have multiple feature_inds, the name of the treatment is used. 
image

The name of the treatment should always be used, even with one treatment:

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1044 @@
...
+        T0 = _safe_indexing(X, feature_index, axis=1)
+        T1 = Xnew
+        result = self._safe_result_index(X, feature_index)
+        inf = result.estimator.effect_inference(
+            X=result.X_transformer.transform(X), T0=T0, T1=T1)
+
+        # we want to offset the inference object by the baseline estimate of y
+        inf.translate(y)
+
+        return inf
```
vsyrgkanis, line 844: I think we should be returning a pandas data frame again in the same spirit as we return pandas data frames in all other results. So that the user doesn't need to know our inference results object. 

So we should make alpha part of the input and return a pandas data frame with that alpha

vsyrgkanis (changes requested):

heimengqi (changes requested):

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: We can't return an inference object in the other cases because we've got one-per-feature; here, it's easy to get a data frame out if the user wants to by just getting the summary off of the inference object.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1044 @@
...
+        model, then only the features in this array will be used for heterogeneity. If a 2d
+        array then its first dimension should be len(feature_inds) and whenever estimating
+        a local causal effect for target feature feature_inds[i], then only features in
+        heterogeneity_inds[i] will be used for heterogeneity. If heterogeneity_inds[i]=None, then all features
+        are used for heterogeneity when estimating local causal effect for feature_inds[i]. If heterogeneity_ind=None
+        then all features are used for heterogeneity for all features.
+    feature_names: list of str, default None
+        The names for all of the features in the data.  Not necessary if the input will be a dataframe.
+        If None and the input is a plain numpy array, generated feature names will be ['X1', 'X2', ...].
+    upper_bound_on_cat_expansion: int, default 5
```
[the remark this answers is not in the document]
kbattocchi: Yeah, there's a TODO later noting that it's not yet implemented but should be.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1044 @@
...
+            "The only supported nuisance models are 'linear' and 'automl', "
+            f"but was given {self.nuisance_models}")
+
+        assert self.heterogeneity_model in ['linear', 'forest'], (
+            "The only supported heterogeneity models are 'linear' and, 'forest but received "
+            f"{self.heterogeneity_model}")
+
+        assert np.ndim(X) == 2, f"X must be a 2-dimensional array, but here had shape {np.shape(X)}"
+
+        if warm_start:
```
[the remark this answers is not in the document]
kbattocchi: Good catch.  This is due to not transforming Y to be compatible with the regression wrapper if we're in the warm start logic.  I've fixed this in my next push.

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: I know but they need to know about our inferenceresults object and it's properties. That's ok for now, but I would make it more uniform in api to always return a standard object like a pandas frame or an array

kbattocchi:
> It seems weird that when we have a single feature_ind, then the global causal effect is returned as:
> image
> with "value" as index, while when we have multiple feature_inds, the name of the treatment is used.
> image
> 
> The name of the treatment should always be used, even with one treatment:

@vsyrgkanis Ok

vsyrgkanis:
> @vsyrgkanis Ok

I might even do the same uniformization even for local effects and maintain the feature_index inner index with the single feature name that we run the analysis for.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Yeah, you're right; on the other hand I'm not sure how much of a problem it is to have short lived workers that don't receive any work.  Going to punt for now given more pressing issues.

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: 0.05 should be the default thorughout
90% confidence is too low.
95% confidence is more standard

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1056 @@
...
+            # perform model selection for the Y model using all X, not on a per-column basis
+            self._x_transform = ColumnTransformer([('encode',
+                                                    OneHotEncoder(
+                                                        drop='first', sparse=False),
+                                                    self.categorical)],
+                                                  remainder='passthrough')
+            allX = self._x_transform.fit_transform(X)
+
+            if self.classification:
+                self._model_y = _first_stage_clf(
```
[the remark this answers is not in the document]
kbattocchi: Needed for multiclass since you'll get multiple outputs otherwise, which the classifier won't handle correctly.  Added a comment to that effect.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Okay, in that case I'll do that in a separate change to come in the next release, but it will affect a *lot* of defaults.

vsyrgkanis: yeap makes sense

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Ok, we can change this in the next release.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1071 @@
...
+        def make_dataframe(props):
+
+            to_include = OrderedDict([(key, value.reshape(-1))
+                                      for key, value in props])
+
+            # TODO: enrich outcome logic for multi-class classification when that is supported
+            index = pd.MultiIndex.from_tuples([(i, outcome, res.feature_name, f"{lvl}v{res.feature_baseline}"
+                                                if res.feature_baseline is not None
+                                                else lvl)
+                                               for i in range(n)
```
vsyrgkanis, line 614: My feeling is that the large extra overhead in local causal effect is stemming from the for i in range(n) loop which is a terrible for loop of 5k samples for instance in native python. We should find a way to avoid a pytrhon for loop over the samples if possible. Also we should first make sure that it is indeed coming from this.

heimengqi (approved):

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -0,0 +1,1071 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
```
vsyrgkanis, line 2: The following raises a bug:
```python
from econml.solutions.causal_analysis import CausalAnalysis
import numpy as np
from itertools import product
import pandas as pd
import matplotlib.pyplot as plt

%matplotlib inline

# Treatment effect function
def exp_te(X):
    return np.exp(2*X[:, 0])
# DGP constants
np.random.seed(123)
n = 5000
n_w = 30
support_size = 5
n_x = 2
# Outcome support
support_Y = np.random.choice(range(n_w), size=support_size, replace=False)
coefs_Y = np.random.uniform(0, 1, size=support_size)
epsilon_sample = lambda n: np.random.uniform(-1, 1, size=n)
# Treatment support 
support_T = support_Y
coefs_T = np.random.uniform(0, 1, size=support_size)
eta_sample = lambda n: np.random.uniform(-1, 1, size=n) 

# Generate controls, covariates, treatments and outcomes
W = np.random.normal(0, 1, size=(n, n_w))
X = np.random.uniform(0, 1, size=(n, n_x))
# Heterogeneous treatment effects
TE = exp_te(X)
Tind = np.dot(W[:, support_T], coefs_T) + eta_sample(n)
T = (Tind>0)
Y = TE * T + np.dot(W[:, support_Y], coefs_Y) + epsilon_sample(n)
T = np.empty(Tind.shape, dtype='str')
T[Tind>0] = 'a'
T[Tind<=0] = 'b'
# Y = (Y > 0)
X_test = np.array(list(product(np.arange(0, 1, 0.01), repeat=n_x)))
A = np.hstack([T.reshape(-1, 1), X, W])
Xcols = [f'X{i}' for i in range(n_x)] 
Wcols = [f'W{i}' for i in range(n_w)]
Adf = pd.DataFrame(A, columns=['T'] + Xcols + Wcols)
ca = CausalAnalysis(feature_inds=['T'],
                    heterogeneity_inds=[Xcols],
                    categorical=['T'],
                    classification=False)
ca.fit(Adf, Y)
ca.feature_inds = ['T', 'X0', f'W{support_Y[0]}']
ca.heterogeneity_inds = [Xcols, Xcols, Xcols]
ca.fit(Adf, Y, warm_start=True)
```

```
---------------------------------------------------------------------------
_RemoteTraceback                          Traceback (most recent call last)
_RemoteTraceback: 
"""
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\externals\loky\process_executor.py", line 431, in _process_worker
    r = call_item()
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\externals\loky\process_executor.py", line 285, in __call__
    return self.fn(*self.args, **self.kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\_parallel_backends.py", line 595, in __call__
    return self.func(*args, **kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\parallel.py", line 262, in __call__
    return [func(*args, **kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\joblib\parallel.py", line 262, in <listcomp>
    return [func(*args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\solutions\causal_analysis\_causal_analysis.py", line 438, in process_feature
    est.fit(y, T, X=X_xf, W=W, cache_values=True)
  File "c:\users\vasy\documents\econml\econml\utilities.py", line 1262, in m
    return to_wrap(*args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\dml\dml.py", line 659, in fit
    return super().fit(Y, T, X=X, W=W,
  File "c:\users\vasy\documents\econml\econml\utilities.py", line 1262, in m
    return to_wrap(*args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\dml\dml.py", line 505, in fit
    return super().fit(Y, T, X=X, W=W, sample_weight=sample_weight, freq_weight=freq_weight,
  File "c:\users\vasy\documents\econml\econml\utilities.py", line 1262, in m
    return to_wrap(*args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\dml\_rlearner.py", line 369, in fit
    return super().fit(Y, T, X=X, W=W,
  File "c:\users\vasy\documents\econml\econml\utilities.py", line 1262, in m
    return to_wrap(*args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\_cate_estimator.py", line 130, in call
    m(self, Y, T, *args, **kwargs)
  File "c:\users\vasy\documents\econml\econml\_ortho_learner.py", line 638, in fit
    nuisances, fitted_models, new_inds, scores = self._fit_nuisances(
  File "c:\users\vasy\documents\econml\econml\_ortho_learner.py", line 767, in _fit_nuisances
    nuisances, fitted_models, fitted_inds, scores = _crossfit(self._ortho_learner_model_nuisance, folds,
  File "c:\users\vasy\documents\econml\econml\_ortho_learner.py", line 170, in _crossfit
    nuisance_temp = model_list[idx].predict(*args_test, **kwargs_test)
  File "c:\users\vasy\documents\econml\econml\dml\_rlearner.py", line 75, in predict
    T_res = T - T_pred.reshape(T.shape)
TypeError: unsupported operand type(s) for -: 'str' and 'float'
"""

The above exception was the direct cause of the following exception:

TypeError                                 Traceback (most recent call last)
<ipython-input-89-f52cbe5cafa7> in <module>
      2 ca.heterogeneity_inds = [Xcols, Xcols, Xcols]
      3 # ca.nuisance_models = 'automl'
----> 4 ca.fit(Adf, Y, warm_start=True)

c:\users\vasy\documents\econml\econml\solutions\causal_analysis\_causal_analysis.py in fit(self, X, y, warm_start)
    500 
    501         cache_updates = dict(zip(new_inds,
--> 502                                  joblib.Parallel(n_jobs=self.n_jobs,
    503                                                  verbose=1)(joblib.delayed(process_feature)(feat_name, feat_ind)
    504                                                             for feat_name, feat_ind in zip(new_feat_names, new_inds))))

C:\ProgramData\Anaconda3\lib\site-packages\joblib\parallel.py in __call__(self, iterable)
   1059 
   1060             with self._backend.retrieval_context():
-> 1061                 self.retrieve()
   1062             # Make sure that we get a last message telling us we are done
   1063             elapsed_time = time.time() - self._start_time

C:\ProgramData\Anaconda3\lib\site-packages\joblib\parallel.py in retrieve(self)
    938             try:
    939                 if getattr(self._backend, 'supports_timeout', False):
--> 940                     self._output.extend(job.get(timeout=self.timeout))
    941                 else:
    942                     self._output.extend(job.get())

C:\ProgramData\Anaconda3\lib\site-packages\joblib\_parallel_backends.py in wrap_future_result(future, timeout)
    540         AsyncResults.get from multiprocessing."""
    541         try:
--> 542             return future.result(timeout=timeout)
    543         except CfTimeoutError as e:
    544             raise TimeoutError from e

C:\ProgramData\Anaconda3\lib\concurrent\futures\_base.py in result(self, timeout)
    437                 raise CancelledError()
    438             elif self._state == FINISHED:
--> 439                 return self.__get_result()
    440             else:
    441                 raise TimeoutError()

C:\ProgramData\Anaconda3\lib\concurrent\futures\_base.py in __get_result(self)
    386     def __get_result(self):
    387         if self._exception:
--> 388             raise self._exception
    389         else:
    390             return self._result

TypeError: unsupported operand type(s) for -: 'str' and 'float'
```

vsyrgkanis (changes requested):

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: I think when you hstack the columns, the resulting array has a weird type because the first column has strings in it, so things get messed up from there.

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: Yea let's put that in the list of todos then

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: Let's put that in the TODOs

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: Let's add performance improvement on this in the TODOs for next release

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: Sounds good. Let's just elaborate a bit more in the docstring that to use no features for heterogneiety use the empty list.

vsyrgkanis (approved):
LGTM

PR #459, opened 2021-05-05
Fix tree exporter utilities.MissingModule import

gregorybchris:
If matplotlib is not available in the environment this import will raise an exception


kbattocchi (approved):
Thanks for the report/fix!

PR #462, opened 2021-05-10
fixed max_samples=.5 in default grid for cfddml.tune

vsyrgkanis:
max samples=.5 doesn't work with inference, so default grid changed to [.3, .45]


kbattocchi (changes requested):
LGTM, but please add a test case that would fail under the old settings

PR #473, opened 2021-05-24
Improve causal analysis functionality

kbattocchi:
This makes a few changes:
* Makes `whatif` return a dataframe to harmonize with the other methods
* Supports exporting trees as dictionaries
* Removes policy treatment values; returns always treat and policy values instead
* Adds an `individualized_policy` method (and dictionary counterpart)

### DIFF
--- a/econml/solutions/causal_analysis/_causal_analysis.py
+++ b/econml/solutions/causal_analysis/_causal_analysis.py
@@ -16,6 +16,7 @@ from sklearn.ensemble import GradientBoostingClassifier, RandomForestClassifier,
 from sklearn.linear_model import Lasso, LassoCV, LogisticRegression, LogisticRegressionCV
 from sklearn.pipeline import make_pipeline, Pipeline
 from sklearn.preprocessing import OneHotEncoder, PolynomialFeatures, StandardScaler
+from sklearn.tree import _tree
 from sklearn.utils.validation import column_or_1d
 from ...cate_interpreter import SingleTreeCateInterpreter, SingleTreePolicyInterpreter
 from ...dml import LinearDML, CausalForestDML
@@ -104,60 +105,64 @@ def _get_metadata_causal_insights_keys():
             _CausalInsightsConstants.ViewKey]
 
 
-def _first_stage_reg(X, y, *, automl=True):
+def _first_stage_reg(X, y, *, automl=True, random_state=None, verbose=0):
     if automl:
-        model = GridSearchCVList([make_pipeline(StandardScaler(), LassoCV()),
+        model = GridSearchCVList([make_pipeline(StandardScaler(), LassoCV(random_state=random_state)),
                                   RandomForestRegressor(
-                                      n_estimators=100, random_state=123, min_samples_leaf=10),
-                                  lgb.LGBMRegressor(num_leaves=32)],
+                                      n_estimators=100, random_state=random_state, min_samples_leaf=10),
+                                  lgb.LGBMRegressor(num_leaves=32, random_state=random_state)],
                                  param_grid_list=[{},
                                                   {'min_weight_fraction_leaf':
                                                       [.001, .01, .1]},
                                                   {'learning_rate': [0.1, 0.3], 'max_depth': [3, 5]}],
-                                 cv=2,
-                                 scoring='neg_mean_squared_error')
+                                 cv=3,
+                                 scoring='r2',
+                                 verbose=verbose)
         best_est = model.fit(X, y).best_estimator_
         if isinstance(best_est, Pipeline):
-            return make_pipeline(StandardScaler(), Lasso(alpha=best_est.steps[1][1].alpha_))
+            return make_pipeline(StandardScaler(), Lasso(alpha=best_est.steps[1][1].alpha_, random_state=random_state))
         else:
             return best_est
     else:
-        model = make_pipeline(StandardScaler(), LassoCV(cv=5)).fit(X, y)
-        return make_pipeline(StandardScaler(), Lasso(alpha=model.steps[1][1].alpha_))
+        model = make_pipeline(StandardScaler(), LassoCV(cv=5, random_state=random_state)).fit(X, y)
+        return make_pipeline(StandardScaler(), Lasso(alpha=model.steps[1][1].alpha_, random_state=random_state))
 
 
-def _first_stage_clf(X, y, *, make_regressor=False, automl=True):
+def _first_stage_clf(X, y, *, make_regressor=False, automl=True, random_state=None, verbose=0):
     if automl:
-        model = GridSearchCVList([make_pipeline(StandardScaler(), LogisticRegression(max_iter=1000)),
-                                  RandomForestClassifier(
-                                      n_estimators=100, random_state=123),
-                                  GradientBoostingClassifier(random_state=123)],
+        model = GridSearchCVList([make_pipeline(StandardScaler(), LogisticRegression(max_iter=1000,
+                                                                                     random_state=random_state)),
+                                  RandomForestClassifier(n_estimators=100, min_samples_leaf=10,
+                                                         random_state=random_state),
+                                  lgb.LGBMClassifier(num_leaves=32, random_state=random_state)],
                                  param_grid_list=[{'logisticregression__C': [0.01, .1, 1, 10, 100]},
-                                                  {'max_depth': [3, 5],
-                                                   'min_samples_leaf': [10, 50]},
-                                                  {'n_estimators': [50, 100],
-                                                   'max_depth': [3],
-                                                   'min_samples_leaf': [10, 30]}],
-                                 cv=5,
-                                 scoring='neg_log_loss')
+                                                  {'max_depth': [3, None],
+                                                   'min_weight_fraction_leaf': [.001, .01, .1]},
+                                                  {'learning_rate': [0.1, 0.3], 'max_depth': [3, 5]}],
+                                 cv=3,
+                                 scoring='neg_log_loss',
+                                 verbose=verbose)
         est = model.fit(X, y).best_estimator_
     else:
-        model = make_pipeline(StandardScaler(), LogisticRegressionCV(cv=5, max_iter=1000)).fit(X, y)
-        est = make_pipeline(StandardScaler(), LogisticRegression(C=model.steps[1][1].C_[0]))
+        model = make_pipeline(StandardScaler(), LogisticRegressionCV(
+            cv=5, max_iter=1000, random_state=random_state)).fit(X, y)
+        est = make_pipeline(StandardScaler(), LogisticRegression(
+            C=model.steps[1][1].C_[0], random_state=random_state))
     if make_regressor:
         return _RegressionWrapper(est)
     else:
         return est
 
 
-def _final_stage():
-    return GridSearchCVList([WeightedLasso(),
-                             RandomForestRegressor(n_estimators=100, random_state=123)],
+def _final_stage(*, random_state=None, verbose=0):
+    return GridSearchCVList([WeightedLasso(random_state=random_state),
+                             RandomForestRegressor(n_estimators=100, random_state=random_state, verbose=verbose)],
                             param_grid_list=[{'alpha': [.001, .01, .1, 1, 10]},
                                              {'max_depth': [3, 5],
                                               'min_samples_leaf': [10, 50]}],
-                            cv=5,
-                            scoring='neg_mean_squared_error')
+                            cv=3,
+                            scoring='neg_mean_squared_error',
+                            verbose=verbose)
 
 
 # simplification of sklearn's ColumnTransformer that encodes categoricals and passes through selected other columns
@@ -173,8 +178,9 @@ class _ColumnTransformer(TransformerMixin):
         cat_cols = _safe_indexing(X, self.categorical, axis=1)
         if cat_cols.shape[1] > 0:
             self.has_cats = True
-            self.one_hot_encoder = OneHotEncoder(
-                drop='first', sparse=False).fit(cat_cols)
+            # NOTE: set handle_unknown to 'ignore' so that we don't throw at runtime if given a novel value
+            self.one_hot_encoder = OneHotEncoder(sparse=False,
+                                                 handle_unknown='ignore').fit(cat_cols)
         else:
             self.has_cats = False
         self.d_x = X.shape[1]
@@ -210,15 +216,172 @@ def _sanitize(obj):
     else:
         try:
             return [_sanitize(item) for item in obj]
-        except e:
+        except Exception:
             raise ValueError(f"Could not sanitize input {obj}")
+
+
+# Convert SingleTreeInterpreter to a python dictionary
+def _tree_interpreter_to_dict(interp, features, leaf_data=lambda t, n: {}):
+    tree = interp.tree_model_.tree_
+
+    def recurse(node_id):
+        if tree.children_left[node_id] == _tree.TREE_LEAF:
+            return {'leaf': True, 'n_samples': tree.n_node_samples[node_id], **leaf_data(tree, node_id)}
+        else:
+            return {'leaf': False, 'feature': features[tree.feature[node_id]], 'threshold': tree.threshold[node_id],
+                    'left': recurse(tree.children_left[node_id]),
+                    'right': recurse(tree.children_right[node_id])}
+
+    return recurse(0)
 
 
 # named tuple type for storing results inside CausalAnalysis class;
 # must be lifted to module level to enable pickling
 _result = namedtuple("_result", field_names=[
     "feature_index", "feature_name", "feature_baseline", "feature_levels", "hinds",
-    "X_transformer", "W_transformer", "estimator", "global_inference"])
+    "X_transformer", "W_transformer", "estimator", "global_inference", "treatment_value"])
+
+
+def _process_feature(name, feat_ind, verbose, categorical_inds, categories, heterogeneity_inds, y, X,
+                     nuisance_models, h_model, random_state, model_y):
+    if verbose > 0:
+        print(f"CausalAnalysis: Feature {name}")
+
+    discrete_treatment = feat_ind in categorical_inds
+    if discrete_treatment:
+        cats = categories[categorical_inds.index(feat_ind)]
+    else:
+        cats = 'auto'  # just leave the setting at the default otherwise
+
+    hinds = heterogeneity_inds[feat_ind]
+    WX_transformer = ColumnTransformer([('encode', OneHotEncoder(drop='first', sparse=False),
+                                         [ind for ind in categorical_inds
+                                             if ind != feat_ind]),
+                                        ('drop', 'drop', feat_ind)],
+                                       remainder='passthrough')
+    W_transformer = ColumnTransformer([('encode', OneHotEncoder(drop='first', sparse=False),
+                                        [ind for ind in categorical_inds
+                                            if ind != feat_ind and ind not in hinds]),
+                                       ('drop', 'drop', hinds),
+                                       ('drop_feat', 'drop', feat_ind)],
+                                      remainder='passthrough')
+    # Use _ColumnTransformer instead of ColumnTransformer so we can get feature names
+    X_transformer = _ColumnTransformer([ind for ind in categorical_inds
+                                        if ind != feat_ind and ind in hinds],
+                                       [ind for ind in hinds
+                                        if ind != feat_ind and ind not in categorical_inds])
+
+    # Controls are all other columns of X
+    WX = WX_transformer.fit_transform(X)
+    # can't use X[:, feat_ind] when X is a DataFrame
+    T = _safe_indexing(X, feat_ind, axis=1)
+
+    # TODO: we can't currently handle unseen values of the feature column when getting the effect;
+    #       we might want to modify OrthoLearner (and other discrete treatment classes)
+    #       so that the user can opt-in to allowing unseen treatment values
+    #       (and return NaN or something in that case)
+
+    W = W_transformer.fit_transform(X)
+    X_xf = X_transformer.fit_transform(X)
+    if W.shape[1] == 0:
+        # array checking routines don't accept 0-width arrays
+        W = None
+
+    if X_xf.shape[1] == 0:
+        X_xf = None
+
+    if verbose > 0:
+        print("CausalAnalysis: performing model selection on T model")
+
+    # perform model selection
+    model_t = (_first_stage_clf(WX, T, automl=nuisance_models == 'automl',
+                                random_state=random_state, verbose=verbose)
+               if discrete_treatment else _first_stage_reg(WX, T, automl=nuisance_models == 'automl',
+                                                           random_state=random_state,
+                                                           verbose=verbose))
+
+    if X_xf is None and h_model == 'forest':
+        warnings.warn(f"Using a linear model instead of a forest model for feature '{name}' "
+                      "because forests don't support models with no heterogeneity indices")
+        h_model = 'linear'
+
+    if h_model == 'linear':
+        est = LinearDML(model_y=model_y,
+                        model_t=model_t,
+                        discrete_treatment=discrete_treatment,
+                        fit_cate_intercept=True,
+                        linear_first_stages=False,
+                        categories=cats,
+                        random_state=random_state)
+    elif h_model == 'forest':
+        est = CausalForestDML(model_y=model_y,
+                              model_t=model_t,
+                              discrete_treatment=discrete_treatment,
+                              n_estimators=4000,
+                              min_var_leaf_on_val=True,
+                              categories=cats,
+                              random_state=random_state,
+                              verbose=verbose)
+
+        if verbose > 0:
+            print("CausalAnalysis: tuning forest")
+        est.tune(y, T, X=X_xf, W=W)
+    if verbose > 0:
+        print("CausalAnalysis: training causal model")
+    est.fit(y, T, X=X_xf, W=W, cache_values=True)
+
+    # Prefer ate__inference to const_marginal_ate_inference(X) because it is doubly-robust and not conservative
+    if h_model == 'forest' and discrete_treatment:
+        global_inference = est.ate__inference()
+    else:
+        # convert to NormalInferenceResults for consistency
+        inf = est.const_marginal_ate_inference(X=X_xf)
+        global_inference = NormalInferenceResults(d_t=inf.d_t, d_y=inf.d_y,
+                                                  pred=inf.mean_point,
+                                                  pred_stderr=inf.stderr_mean,
+                                                  mean_pred_stderr=None,
+                                                  inf_type='ate')
+
+    # Set the dictionary values shared between local and global summaries
+    if discrete_treatment:
+        cats = est.transformer.categories_[0]
+        baseline = cats[est.transformer.drop_idx_[0]]
+        cats = cats[np.setdiff1d(np.arange(len(cats)),
+                                 est.transformer.drop_idx_[0])]
+        d_t = len(cats)
+        insights = {
+            _CausalInsightsConstants.TypeKey: ['cat'] * d_t,
+            _CausalInsightsConstants.RawFeatureNameKey: [name] * d_t,
+            _CausalInsightsConstants.CategoricalColumnKey: cats.tolist(),
+            _CausalInsightsConstants.EngineeredNameKey: [
+                f"{name} (base={baseline}): {c}" for c in cats]
+        }
+        treatment_value = 1
+    else:
+        d_t = 1
+        cats = ["num"]
+        baseline = None
+        insights = {
+            _CausalInsightsConstants.TypeKey: ["num"],
+            _CausalInsightsConstants.RawFeatureNameKey: [name],
+            _CausalInsightsConstants.CategoricalColumnKey: [name],
+            _CausalInsightsConstants.EngineeredNameKey: [name]
+        }
+        # calculate a "typical" treatment value, using the mean of the absolute value of non-zero treatments
+        treatment_value = np.mean(np.abs(T[T != 0]))
+
+    result = _result(feature_index=feat_ind,
+                     feature_name=name,
+                     feature_baseline=baseline,
+                     feature_levels=cats,
+                     hinds=hinds,
+                     X_transformer=X_transformer,
+                     W_transformer=W_transformer,
+                     estimator=est,
+                     global_inference=global_inference,
+                     treatment_value=treatment_value)
+
+    return insights, result
 
 
 class CausalAnalysis:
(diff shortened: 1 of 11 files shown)


econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -997,12 +1023,21 @@ def _tree(self, is_policy, Xtest, feature_index, *, treatment_cost=0,
...
+            if result.feature_baseline is None:  # continuous treatment, so apply a treatment level 10% of typical
+                treatment_level = result.treatment_value * 0.1
+            else:
+                treatment_level = 1
+            # TODO: this calculation is correct only if treatment consts are marginal costs,
+            #       because then scaling the difference between treatment value and treatment costs is the
+            #       same as scaling the treatment value and subtracting the scaled treatment cost.
+            #       If treatment costs are absolute, then we'll need to modify the policy tree code itself to handle
+            #       treatment levels other than 1.
+            policy_values = intrp.policy_value_ * treatment_level, intrp.always_treat_value_ * treatment_level
```
heimengqi, line 1035: Here seems still unclear to me for both continue and discrete case.

```
        all_y = np.hstack([np.zeros((y_pred.shape[0], 1)), np.atleast_1d(y_pred)])

        self.tree_model_.fit(X, all_y)
        self.policy_value_ = np.mean(np.max(self.tree_model_.predict_value(X), axis=1))
        self.always_treat_value_ = np.mean(y_pred, axis=0)
``` 
The code above is how we get `policy_value_`, it's the average of our best treatment policy, where some subgroup we treat, some subgroup we don't treat, why we give a treatment level for all? should we only give a treatment level for treated sub group?

I might misunderstand it...please let me know how to explain it.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -1136,10 +1176,116 @@ def _heterogeneity_tree_string(self, Xtest, feature_index, *,
...
+                                                                        axis=1)
+
+        Xtest = result.X_transformer.transform(Xtest)
+        if Xtest.shape[1] == 0:
+            Xtest = None
+
+        effect = result.estimator.const_marginal_effect_inference(Xtest)
+        if result.feature_baseline is None:
+            # apply 10% of a typical treatment for this feature
+            effect.scale(result.treatment_value * 0.1)
```
heimengqi, line 1225: Instead of add a scale function under Inference class, why not learn the effect direction index from const_marginal_effect, and they create two vector T0 as a vector of `treatment_value` and T1 as a vector of either (treatment_value*1.1) or (treatment_value*0.9) depends on the index of increase or decrease, they call effect(X,T0=T0, T1=T1)? it will automatically give you the right point estimate and CI?

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -1136,10 +1176,116 @@ def _heterogeneity_tree_string(self, Xtest, feature_index, *,
...
+            eff = np.take_along_axis(all_effs, eff_ind, 1).reshape(-1)
+            eff_lb = np.take_along_axis(all_eff_lbs, eff_ind, 1).reshape(-1)
+            eff_ub = np.take_along_axis(all_eff_ubs, eff_ind, 1).reshape(-1)
+
+        df = pd.DataFrame.from_dict({'Treatment': rec,
+                                     'Effect of treatment': eff,
+                                     'Effect of treatment lower bound': eff_lb,
+                                     'Effect of treatment upper bound': eff_ub})
+
+        return df.join(orig_df).sort_values('Effect of treatment',
```
heimengqi, line 1262: Why we want to save the `orig_df` here which **excludes** the heterogeneous features?

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -1136,10 +1176,116 @@ def _heterogeneity_tree_string(self, Xtest, feature_index, *,
...
+            # set the effect bounds; for positive treatments these agree with
+            # the estimates; for negative treatments, we need to invert the interval
+            eff_lb, eff_ub = est_lb, est_ub
+            eff_lb[est <= 0], eff_ub[est <= 0] = -eff_ub[est <= 0], -eff_lb[est <= 0]
+            # the effect is now always positive since we decrease treatment when negative
+            eff = np.abs(est)
+        else:
+            # for discrete treatment, stack a zero result in front for control
+            zeros = np.zeros((est.shape[0], 1))
+            all_effs = np.hstack([zeros, est])
```
heimengqi, line 1246: As we discussed in our sync today, here the assumption is in discrete case, different treatments have equal cost.

heimengqi (changes requested):
Looks good! There are a couple of places I think the logic looks strange to me, especially related to policy interpreter, I leave the comments and please help correct me if my understanding is wrong.

vsyrgkanis:
Currently the DR correction in the CausalForestDML has too small of a covariance clipping here:
https://github.com/microsoft/EconML/blob/dbbd9488712f6e92566c2df588675dbe8fd0f73f/econml/dml/causal_forest.py#L77

causing great instability in the ate__inference and very large variances. 

Can we increase that to 1e-2 in this PR

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: What does this function return when treatment is categorical with many categories? Has this been tested and has it been documented in the docstring what is the form of the output?

vsyrgkanis: We should also document in the docstring that for continuous features, the treatment_cost is the "per unit treatment cost, or marginal treatment cost", i.e. that each extra unit of treatment costs that, i.e. the total cost is "treatment_cost * treatment_level". 

Does the docstring also document what values 'treatment_cost" can take when treatment is categorical with multiple categories? Can I pass a list which contains the cost for each non-baseline treatment?

vsyrgkanis: Also for categorical features, the "always treat" should be a list containing the values of always treatment with each separate non baseline treatment. Shall we then maybe also be returning the associated "names of those non-baseline treatment". So that then the UX can say: gains over always treatment wit T="A", gains over always treatment wit T="B" etc. for each bar in the bar plot

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -1065,18 +1100,23 @@ def _policy_tree_output(self, Xtest, feature_index, *, treatment_cost=0,
...
+        (intrp, feature_names, treatment_names,
+            (policy_val, always_trt)) = self._tree(True, Xtest, feature_index,
+                                                   treatment_cost=treatment_cost,
+                                                   max_depth=max_depth,
+                                                   min_samples_leaf=min_samples_leaf,
+                                                   min_impurity_decrease=min_value_increase,
+                                                   alpha=alpha)
+
+        def policy_data(tree, node_id):
+            return {'treatment': treatment_names[np.argmax(tree.value[node_id])]}
```
vsyrgkanis, line 1118: I think we should also be returning the local average policy value at each node (maybe not with confidence intervals). This value is being displayed in the graphviz tree. You can use the node_dict_ that is constructed at fit time by  the interpreter. 
https://github.com/microsoft/EconML/blob/dbbd9488712f6e92566c2df588675dbe8fd0f73f/econml/cate_interpreter/_interpreters.py#L469

This is what is being used by the graphviz tree. Then you can just do: node_dict_[node_id]['mean'], node_dict_[node_id]['ci']

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -214,11 +215,26 @@ def _sanitize(obj):
...
+
 # named tuple type for storing results inside CausalAnalysis class;
 # must be lifted to module level to enable pickling
 _result = namedtuple("_result", field_names=[
     "feature_index", "feature_name", "feature_baseline", "feature_levels", "hinds",
-    "X_transformer", "W_transformer", "estimator", "global_inference"])
+    "X_transformer", "W_transformer", "estimator", "global_inference", "treatment_value"])
 
 
 class CausalAnalysis:
```
vsyrgkanis, line 240: If the only thing that has changed in between runs for a particular feature are it's corresponding "heterogeneity_inds", then we should not repeat the model selection for model_t (if warm_start=True) as it is exactly the same call, since WX has not changed.

So we should add this logic in line 494.

vsyrgkanis: Also why are we retraining a model_y if self.heterogeneity_model has changed? The call to model_y model selection is exactly the same.

Also even if we want to go with that, then why are we not also re-training model_y if hetergogeneity_inds has changed but heterogeneity_model hasnt.

I think in either case retraining model_y is not needed.

vsyrgkanis (changes requested):

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: Actually the node_dict just gives you the local CATE information. 
The tree.value[node_id] also contains the cost, i.e. it is the average: CATE-cost.
So maybe we should return:
'treatment': ..
'gains': ... (value - cost from tree.value[node_id])
'effect': ... (from node_dict_[node_id]['mean'])
'effect_ci': ... (from node_dict)

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: Also weirdly I got the warning:
`Using a linear model instead of a forest model for feature 'IsSupport' because forests don't support models with no heterogeneity indices`

Even when I specified `heterogeneity_model='linear'` and an empty list for hterogeneity_inds. This should only be raised if `hetergeneity='forest'`

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -1136,10 +1176,116 @@ def _heterogeneity_tree_string(self, Xtest, feature_index, *,
...
+                                                max_depth=max_depth,
+                                                min_samples_leaf=min_samples_leaf,
+                                                min_impurity_decrease=min_impurity_decrease,
+                                                alpha=alpha)
+
+        def hetero_data(tree, node_id):
+            return {'effect': _sanitize(tree.value[node_id])}
+        return _tree_interpreter_to_dict(intrp, feature_names, hetero_data)
+
+    def individualized_policy(self, Xtest, feature_index, *, n_rows=None, treatment_costs=0, alpha=0.1):
```
vsyrgkanis, line 1189: When I use this function all of the other features are somehow coming up as NaN when the input data frame has an index.
```python
inds = np.arange(1000)
np.random.shuffle(inds)
X = pd.DataFrame(np.random.normal(0, 1, size=(1000, 2)), columns=['A', 'B'], index=inds)
y = np.random.normal(0, 1, size=1000)
ca_test = CausalAnalysis(feature_inds=['A'], categorical=[])
ca_test.fit(X, y)
ind_policy = ca_test.individualized_policy(X[:50], feature_index='A')
ind_policy
```

vsyrgkanis (changes requested):

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis, line 240: Shouldn't we also have a "random_state" input to the object, so that people will be able to reproduce results by rerunning and so that this random_state if not None, is propagated to all parts of the process (e.g. model selection and CATE estimation)

vsyrgkanis (changes requested):

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -214,11 +215,26 @@ def _sanitize(obj):
...
+
+    return recurse(0)
+
+
 # named tuple type for storing results inside CausalAnalysis class;
 # must be lifted to module level to enable pickling
 _result = namedtuple("_result", field_names=[
     "feature_index", "feature_name", "feature_baseline", "feature_levels", "hinds",
-    "X_transformer", "W_transformer", "estimator", "global_inference"])
+    "X_transformer", "W_transformer", "estimator", "global_inference", "treatment_value"])
```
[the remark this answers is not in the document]
kbattocchi: Sorry, for the latter you mean in [causal_forest.py](../blob/master/econml/dml/causal_forest.py)?

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Yes, done.

kbattocchi: I have fixed the last issue with the warning.

At the moment the logic is relatively simple: we have a single "warm_start" flag which we choose not to honor if certain things have changed which would invalidate part of the already-computed results, and we have feature processing code that we run in its entirety for every feature that has been added/updated.  (But note that contra the second half of your second comment, we already *do not* retrain the y_model if only the heterogeneity indices have changed, we only ignore the warm_start flag if the nuisance models or  heterogeneity model has changed, since that invalidates all existing estimators)

It is possible to add additional checks to optimize out even more computation as you've suggested here, but I fear that it will be harder to verify the correctness of the resulting logic (e.g. how do you add a test that verifies that the y_model is trained only when necessary?).  How critical is optimizing the warm_start logic to eliminate all redundant computation, as opposed to having it be a convenience for adding a few more features?

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -537,6 +554,9 @@ def process_feature(name, feat_ind):
                     _CausalInsightsConstants.CategoricalColumnKey: [name],
                     _CausalInsightsConstants.EngineeredNameKey: [name]
                 }
+                # calculate a "typical" treatment value, using the mean of the absolute value of non-zero treatments
+                treatment_value = np.mean(np.abs(T[T != 0]))
```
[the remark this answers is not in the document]
kbattocchi: Yes: if for some units it's positive and some it's negative, we don't want to average it to zero: the "typical" treatment value is the magnitude, and when we learn a policy we'll either increase or decrease by some fraction of the typical (currently 10%)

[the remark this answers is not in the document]
kbattocchi: You're right that if the treatment is never positive, then having the "typical" treatment also be negative might be more natural in some sense, but since our policy is expressed in terms of "increase" vs. "decrease", we want to be multiplying our policy values by a positive number to maintain the correct sign.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -997,12 +1023,21 @@ def _tree(self, is_policy, Xtest, feature_index, *, treatment_cost=0,
         if is_policy:
             intrp.interpret(result.estimator, Xtest,
                             sample_treatment_costs=treatment_cost)
-            treat = intrp.treat(Xtest)
-        else:  # no treatment cost for CATE trees
+            if result.feature_baseline is None:  # continuous treatment, so apply a treatment level 10% of typical
+                treatment_level = result.treatment_value * 0.1
+            else:
+                treatment_level = 1
```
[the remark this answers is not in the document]
kbattocchi: I've changed this logic so that we don't do any multiplication in the discrete treatment case because it's not necessary.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: It turns out this need not be the case: treatment costs can be specified as an (n, d_t-1) array of costs that vary by individual and treatment.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: I think it would be confusing to include that feature since we're saying to increase or decrease it (treating it as a treatment), but if you think it's better to include it we could.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: I think the scale function is generally useful, but I agree that effect could be used here.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Both fixed.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: The docstring has been updated and tests added; I have not yet added a return for the feature names.

kbattocchi: I've also expanded the explanation of what we compute here and why the math is correct.

econml/tests/test_causal_analysis.py
```diff
@@ -481,3 +490,86 @@ def test_over_cat_limit(self):
...
+        ca = CausalAnalysis(inds, cats, hinds, heterogeneity_model='linear')
+        ca.fit(X, y)
+        df = ca.individualized_policy(X, 'a')
+        self.assertEqual(df.shape[0], 500)  # all rows included by default
+        self.assertEqual(df.shape[1], 4 + X.shape[1] - 1)  # new cols for policy, effect, upper and lower bounds
+        df = ca.individualized_policy(X, 'b', n_rows=5)
+        self.assertEqual(df.shape[0], 5)
+        self.assertEqual(df.shape[1], 4 + X.shape[1] - 1)  # new cols for policy, effect, upper and lower bounds
+        # verify that we can use a scalar treatment cost
+        df = ca.individualized_policy(X, 'c', treatment_costs=100)
```
gregorybchris, line 513: Will you standardize `treatment_cost` vs `treatment_costs`? I think this class originally used `treatment_cost` which has a nice vector-value interpretation. If the default is `0` with optional individualized cost, then the singular "cost" makes sense to me.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -1136,10 +1274,120 @@ def _heterogeneity_tree_string(self, Xtest, feature_index, *,
...
+            treatment_arr = np.array([result.feature_baseline] + [lvl for lvl in result.feature_levels], dtype=object)
+            rec = treatment_arr[eff_ind]
+            eff_ind = eff_ind.reshape(-1, 1)
+            eff = np.take_along_axis(all_effs, eff_ind, 1).reshape(-1)
+            eff_lb = np.take_along_axis(all_eff_lbs, eff_ind, 1).reshape(-1)
+            eff_ub = np.take_along_axis(all_eff_ubs, eff_ind, 1).reshape(-1)
+
+        df = pd.DataFrame({'Treatment': rec,
+                           'Effect of treatment': eff,
+                           'Effect of treatment lower bound': eff_lb,
```
gregorybchris, line 1360: Out of scope of PR, but you may want to make these constants publicly accessible. Maybe just `CausalAnalysis.TREATMENT_EFFECT_LOWER`? Then EconML users don't have to redefine their own constants.

econml/tests/test_causal_analysis.py
kbattocchi: Good point, thanks, I'll switch it to `treatment_cost` here, too.

kbattocchi: Er, actually I'm going to do the opposite and standardize on `treatment_costs` because that's what's already used in our interpreters, and if we have multiple discrete treatments then there really can be multiple costs per sample.

gregorybchris: Makes sense! Thanks

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -104,60 +105,67 @@ def _get_metadata_causal_insights_keys():
...
                                   RandomForestRegressor(
-                                      n_estimators=100, random_state=123, min_samples_leaf=10),
-                                  lgb.LGBMRegressor(num_leaves=32)],
+                                      n_estimators=100, random_state=random_state, min_samples_leaf=10),
+                                  lgb.LGBMRegressor(num_leaves=32, random_state=random_state)],
                                  param_grid_list=[{},
                                                   {'min_weight_fraction_leaf':
                                                       [.001, .01, .1]},
                                                   {'learning_rate': [0.1, 0.3], 'max_depth': [3, 5]}],
                                  cv=2,
```
vsyrgkanis, line 118: why do we have cv=2 in the regression and cv=5 in the classification?

vsyrgkanis (changes requested):

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: Also it seems that neg_mse is a very cryptic score and we might want to switch to r2score for regression and log loss for classification .

vsyrgkanis: or even simpler, switch to r2score at least for both, so that the score is normalized in [0, 1]

econml/solutions/causal_analysis/_causal_analysis.py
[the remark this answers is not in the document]
vsyrgkanis: yeap

vsyrgkanis (approved):
LGTM

heimengqi (approved):

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Switched to `cv=3` for both, and switched regression to r2score; not sure r2score makes sense for a classifier so didn't change that one now, but can revisit if you feel strongly.

econml/solutions/causal_analysis/_causal_analysis.py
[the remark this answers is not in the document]
kbattocchi: Done

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -104,60 +105,67 @@ def _get_metadata_causal_insights_keys():
...
+                                      n_estimators=100, random_state=random_state, min_samples_leaf=10),
+                                  lgb.LGBMRegressor(num_leaves=32, random_state=random_state)],
                                  param_grid_list=[{},
                                                   {'min_weight_fraction_leaf':
                                                       [.001, .01, .1]},
                                                   {'learning_rate': [0.1, 0.3], 'max_depth': [3, 5]}],
-                                 cv=2,
-                                 scoring='neg_mean_squared_error')
+                                 cv=3,
+                                 scoring='r2_score',
```
vsyrgkanis, line 119: This should be 'r2', 'r2_score' throws an error
https://scikit-learn.org/stable/modules/model_evaluation.html

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -938,7 +1014,7 @@ def whatif(self, X, Xnew, feature_index, y):
 
         return inf
 
-    def _whatif_dict(self, X, Xnew, feature_index, y, alpha=0.1):
+    def whatif(self, X, Xnew, feature_index, y, *, alpha=0.1):
```
vsyrgkanis, line 1017: I think it would be good if the user passes a pandas df, that we maintain the same index in our results that we return, so that they can potentially do subsequent merging later on. 

Currently we get rid of the index and return a df with a 0, 1, 2, ... index

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -955,58 +1031,107 @@ def _whatif_dict(self, X, Xnew, feature_index, y, alpha=0.1):
...
         else:
             treatment_names = [f"{result.feature_baseline}"] + \
                 [f"{lvl}" for lvl in result.feature_levels]
 
-        if len(treatment_names) > 2 and is_policy:
-            raise AssertionError("Can't create policy trees for multi-class features, "
-                                 f"but this feature has values {treatment_names}")
-
         TreeType = SingleTreePolicyInterpreter if is_policy else SingleTreeCateInterpreter
         intrp = TreeType(include_model_uncertainty=True,
```
vsyrgkanis, line 1091: Since we are not surfacing the uncertainty I would either set this to False (as it severely impacts run time especially in forests) or more ideally, I would make `include_uncertainty` as an extra option, with default = False, in all the policy methods and set this equal to the passed `include_uncertainty` option

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -362,7 +549,7 @@ def fit(self, X, y, warm_start=False):
                               "since the previous call to fit")
                 warm_start = False
 
-            # TODO: bail out also if categorical columns, classification changed?
+            # TODO: bail out also if categorical columns, classification, random_state changed?
```
vsyrgkanis, line 552: I suspect we should not retrain a model_y when only heterogeneity_model has changed. 

If that is the case, then why are we also not checking if "heterogeneity_inds has changed here.

vsyrgkanis (approved):

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -104,60 +105,67 @@ def _get_metadata_causal_insights_keys():
...
     if automl:
-        model = GridSearchCVList([make_pipeline(StandardScaler(), LogisticRegression(max_iter=1000)),
+        model = GridSearchCVList([make_pipeline(StandardScaler(), LogisticRegression(max_iter=1000,
+                                                                                     random_state=random_state,
+                                                                                     verbose=verbose)),
                                   RandomForestClassifier(
-                                      n_estimators=100, random_state=123),
-                                  GradientBoostingClassifier(random_state=123)],
+                                      n_estimators=100, random_state=random_state, verbose=verbose),
+                                  GradientBoostingClassifier(random_state=random_state)],
```
vsyrgkanis, line 138: Why are we using a GradientBoostingClassifier here and not an lgbm.LGBClassifier, analogous to what we are doing in the regression with similar hyperparameter search as in the regression?

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Fixed, thanks

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -104,60 +105,67 @@ def _get_metadata_causal_insights_keys():
...
+def _first_stage_clf(X, y, *, make_regressor=False, automl=True, random_state=None, verbose=0):
     if automl:
-        model = GridSearchCVList([make_pipeline(StandardScaler(), LogisticRegression(max_iter=1000)),
+        model = GridSearchCVList([make_pipeline(StandardScaler(), LogisticRegression(max_iter=1000,
+                                                                                     random_state=random_state,
+                                                                                     verbose=verbose)),
                                   RandomForestClassifier(
-                                      n_estimators=100, random_state=123),
-                                  GradientBoostingClassifier(random_state=123)],
+                                      n_estimators=100, random_state=random_state, verbose=verbose),
```
vsyrgkanis, line 137: I think we should remove the verbosity from the RF classifier and from the logistic regresison. Especially the RF classifier pringts too many things. 

For instance, we have not added these  verbosities in the regression case for the RF regressor. So also for uniformity

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Not sure I understand the question; however I have now changed the logic so that we do not retrain a y model when heterogeneity_model changes

kbattocchi: But maybe a partial answer to your question is that even if some heterogeneity_inds have changed (so that models need to be retrained there), others may not have and so we want to keep most of the warm_start logic intact.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Done

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: @vsyrgkanis Any thoughts on this?

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Ok, done.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -104,60 +105,67 @@ def _get_metadata_causal_insights_keys():
             _CausalInsightsConstants.ViewKey]
 
 
-def _first_stage_reg(X, y, *, automl=True):
+def _first_stage_reg(X, y, *, automl=True, random_state=None, verbose=0):
     if automl:
-        model = GridSearchCVList([make_pipeline(StandardScaler(), LassoCV()),
+        model = GridSearchCVList([make_pipeline(StandardScaler(), LassoCV(random_state=random_state)),
```
vsyrgkanis, line 110: Since now we are scaling the outcome, we should not use LassoCV, but rather just Lasso and add a grid of hyperparameters. This way we are selecting best on the CV of the grid search and not just running LassoCV for each fold. 

Then we also don't need to do the thing you have afterwards where we replace lassocv with lasso.

We could also use the sklearn function to generate the alpha grid:
https://github.com/scikit-learn/scikit-learn/blob/15a949460dbf19e5e196b8ef48f9712b72a3b3c3/sklearn/linear_model/_coordinate_descent.py#L81

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -104,60 +105,67 @@ def _get_metadata_causal_insights_keys():
...
-        return make_pipeline(StandardScaler(), Lasso(alpha=model.steps[1][1].alpha_))
+        model = make_pipeline(StandardScaler(), LassoCV(cv=5, random_state=random_state)).fit(X, y)
+        return make_pipeline(StandardScaler(), Lasso(alpha=model.steps[1][1].alpha_, random_state=random_state))
 
 
-def _first_stage_clf(X, y, *, make_regressor=False, automl=True):
+def _first_stage_clf(X, y, *, make_regressor=False, automl=True, random_state=None, verbose=0):
     if automl:
-        model = GridSearchCVList([make_pipeline(StandardScaler(), LogisticRegression(max_iter=1000)),
+        model = GridSearchCVList([make_pipeline(StandardScaler(), LogisticRegression(max_iter=1000,
```
vsyrgkanis, line 133: Lets' replace this for better uniformity with regression and so as to use lgb, which is better and faster for boosting, with:
```python
model = GridSearchCVList([make_pipeline(StandardScaler(), LogisticRegression(max_iter=1000,
                                                                                     random_state=random_state)),
                                  RandomForestClassifier(
                                      n_estimators=100, min_samples_leaf=10, random_state=random_state),
                                  lgb.LGBMClassifier(num_leaves=32, random_state=random_state)],
                                 param_grid_list=[{'logisticregression__C': [0.01, .1, 1, 10, 100]},
                                                  {'max_depth': [3, None],
                                                   'min_weight_fraction_leaf': [.001, .01, .1]},
                                                  {'learning_rate': [0.1, 0.3], 'max_depth': [3, 5]}],
```

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -892,30 +991,7 @@ def _safe_result_index(self, X, feature_index):
         (result,) = results
```
vsyrgkanis, line 991: On line 740 we should not be calling get_inference separately for every attribute, as getting inference can be computationally very costly.

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: Scratch this. If we do sth like this, then the gcv, wont warm start lasso for each alpha point in the grid, and it will be much much slower than running lassocv. So let's leave the lassocv as is.

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: I think it would be good to include the current value of the treatment, maybe changing the field name: "current treatment" This way they can see what was the base of the modification,

Also why is @heimengqi  referring to "hetergoeneous features"? Are these also excluded?

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: ok

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: done

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Addressed by your suggestion above

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Currently only the single specific feature under consideration is being dropped (not all of them), but I'll keep it around based on the feedback from the two of you.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Good catch, fixed.

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: CAn you also make the max_depth [3, None] instead of [3, 5]

kbattocchi: Yes, done.  (BTW, it's easier to see things you want addressed if you unresolve the conversation when you make them)

kbattocchi: @vsyrgkanis You just mean the one for the RandomForestClassifier, not the one for LGBMClassifier, which is still [3,5], right?

vsyrgkanis:
Yeap

PR #476, opened 2021-06-01
Correct default value of alphas

mtanghu:
The PopulationSummaryResults class should to default to self.alpha when the alpha parameter is not passed. This is not the case currently since some methods use the statement 'alpha = self.alpha if alpha is None else alpha', however alpha's default value in the methods is .1, thus self.alpha will be ignored even if the client does not pass the alpha parameter.

Without the fix ate_interval(), marginal_ate_interval() and const_marginal_ate_interval() all ignore the alpha parameter.


heimengqi:
Thanks @mtanghu ! Great catch! I will address that thoroughly in a separate PR.

heimengqi:
@mtanghu We are happy to have more external contributors. I think your current fix have already partially addressed this issue, and we have the similar issue for `value`, `tol` and `decimals` as well, it would be great if you'd like to implement those changes as well.

The logic will be in the initializer we have some default `alpha`, `value`, `tol`, `decimals`  since call `PopulationSummaryResult` will output the summary table, and if users still want to call `PopulationSummaryResult.conf_int_mean()` for instance, by default it will use the same alpha set from initializer, but if they call `.conf_int_mean(alpha=bla)`, it will update the CI with new alpha. In this case, default for `alpha`, `value`, `tol`, `decimals` in each method has to be None.

mtanghu:
@heimengqi I'd love to! I'll do that this weekend!

econml/inference/_inference.py
```diff
@@ -1310,7 +1310,7 @@ def percentile_point(self, *, alpha=.1):
         upper_percentile_point = np.percentile(self.pred, (1 - alpha / 2) * 100, axis=0)
         return lower_percentile_point, upper_percentile_point
 
-    def conf_int_point(self, *, alpha=.1, tol=.001):
+    def conf_int_point(self, *, alpha=None, tol=.001):
```
kbattocchi, line 1313: As @heimengqi pointed out, it would be great to also address `tol` while you're at it.

econml/inference/_inference.py
```diff
@@ -1242,7 +1242,7 @@ def pvalue(self, *, value=0):
         pvalue = norm.sf(np.abs(self.zstat(value=value)), loc=0, scale=1) * 2
         return pvalue
 
-    def conf_int_mean(self, *, alpha=.1):
+    def conf_int_mean(self, *, alpha=None):
```
kbattocchi, line 1245: Could you also please add a test to test_inference.py that fails with the current implementation but will be fixed by your changes?  For example, I believe that a test that trains an estimator, calls `ate_interval` with two different values of alpha, and asserts that the interval from the smaller alpha is wider would do this.

kbattocchi (changes requested):
Thanks, this mostly looks good, but I've added a few extra suggestions (one of which was also made by @heimengqi).

Once those are addressed I'd be happy to approve this.

kbattocchi:
@mtanghu Are you still interested in following this through, or should someone on our team do the remaining work to get this PR cleaned up and then merged?

mtanghu:
@kbattocchi  Oh yes sorry, I'm definitely interested and just caught up in other work. I should be able to get this PR cleaned up soon!

mtanghu:
This should be ready to merge, lmk if there's anything you else you'd like me to help with

econml/inference/_inference.py
mtanghu: Should be done with the recent commits

kbattocchi (approved):
Looks good to me, thanks!

PR #482, opened 2021-06-10
Set default alpha to 0.05 everywhere

kbattocchi:
Previously many places were using 0.1, but 0.05 is more standard.


heimengqi (approved):
Looks good! Have you also checked our notebooks? I feel we will have some place need to update as well (e.g. comments or code)

PR #483, opened 2021-06-18
Keep track of failed features instead of throwing

kbattocchi:


econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -628,21 +658,54 @@ def fit(self, X, y, warm_start=False):
...
+                    invalid_inds.append((ind, 'upper_bound_on_cat_expansion'))
+
+                elif counts[min_ind] < _CAT_LIMIT:
+                    if self.skip_cat_limit_checks and counts[min_ind] >= 2:
+                        # train the model, but warn
+                        warnings.warn(f"Column {ind}'s value {cats[min_ind]} has only {counts[min_ind]} instances in "
+                                      f"the training dataset, which is less than {_CAT_LIMIT}. A model will be fit "
+                                      "because 'skip_cat_limit_checks' is True, but this model may not be robust.")
+                        min_counts[ind] = counts[min_ind]
+                    elif counts[min_ind] < 2:
```
heimengqi: two questions here:
1. Is 2 enough? I heard this morning you said we might have tuning for CausalForestDML will split again internally?
2. the warning below should say the minimum instances is less than 2 explicitly? No matter check cat or not, this can't be trained.

heimengqi (approved):
Looks good!

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Both of these issues have been addressed in the latest commit.

PR #487, opened 2021-06-24
Upgrade shap dependency

gregorybchris:
Upgrades to the current latest version of shap

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,7 +40,7 @@ install_requires =
     statsmodels >= 0.10
     graphviz
     pandas
-    shap ~= 0.38.1
+    shap >= 0.38.1, <= 0.39.0
     dowhy
     lightgbm
 test_suite = econml.tests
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -40,7 +40,7 @@ install_requires =
     statsmodels >= 0.10
     graphviz
     pandas
-    shap ~= 0.38.1
+    shap >= 0.38.1, <= 0.39.0
```
kbattocchi, line 43: I think we'd expect point releases of shap to remain okay if the 0.39.0 release is okay, so a more liberal dependency of

```suggestion
    shap >= 0.38.1, < 0.40
```

would be okay, too.

kbattocchi (approved):
Seems fine to me, as long as it works.

setup.cfg
gregorybchris: You got it!

PR #489, opened 2021-06-29
Set categorical to empty list if not passed

gregorybchris:


econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -507,7 +507,7 @@ def __init__(self, feature_inds, categorical, heterogeneity_inds=None, feature_n
                  upper_bound_on_cat_expansion=5, nuisance_models='linear', heterogeneity_model='linear', *,
                  categories='auto', n_jobs=-1, verbose=0, skip_cat_limit_checks=False, random_state=None):
         self.feature_inds = feature_inds
-        self.categorical = categorical
+        self.categorical = categorical if categorical is not None else []
```
gaugup, line 510: I think down stream code should handle this. We shouldn't override the user config.

gaugup (changes requested):

gregorybchris:
Closing PR and @kbattocchi will consider another fix

PR #493, opened 2021-07-08
Add additional causal analysis flags

kbattocchi:
This PR changes three things:
* Better validation of init args of CausalAnalysis class
* Allow row-wise dictionary output for effects
* Allow preserving all levels in dataframe outputs instead of suppressing single-valued ones


vsyrgkanis (approved):
Make the internal methods also key based only. OW LGTM

PR #504, opened 2021-07-29
Changed the meaning of effect of treatment in individualized policy

vsyrgkanis:
made individualized policy return the delta from the current treatment as effect of recommendation, as compared to delta with respect to baseline. this seems more interpretable.

No other change is made. Schemas remain identical.

Also added the option for confidence intervals in the _heterogeneity_tree_output, since we want the heterogeneity tree to print the confidence intervals too, if we want to integrate it in the UI.

### DIFF
--- a/econml/solutions/causal_analysis/_causal_analysis.py
+++ b/econml/solutions/causal_analysis/_causal_analysis.py
@@ -1587,7 +1587,8 @@ class CausalAnalysis:
             current_treatment = orig_df['Current treatment'].values
             if np.ndim(treatment_costs) >= 2:
                 # remove third dimenions potentially added
-                treatment_costs = treatment_costs.reshape(treatment_costs.shape[:2])
+                if multi_y:  # y was an array, not a vector
+                    treatment_costs = np.squeeze(treatment_costs, 1)
                 assert treatment_costs.shape[1] == len(treatment_arr) - 1, ("If treatment costs are an array, "
                                                                             " they must be of shape (n, d_t-1),"
                                                                             " where n is the number of samples"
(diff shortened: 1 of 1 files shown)


econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -1571,14 +1579,46 @@ def individualized_policy(self, Xtest, feature_index, *, n_rows=None, treatment_
...
             rec = treatment_arr[eff_ind]
-            eff_ind = eff_ind.reshape(-1, 1)
-            eff = np.take_along_axis(all_effs, eff_ind, 1).reshape(-1)
-            eff_lb = np.take_along_axis(all_eff_lbs, eff_ind, 1).reshape(-1)
-            eff_ub = np.take_along_axis(all_eff_ubs, eff_ind, 1).reshape(-1)
+            # we need to call effect_inference to get the correct CI between the two treatment options
+            effect = result.estimator.effect_inference(Xtest, T0=orig_df['Current treatment'], T1=rec)
+            # we now need to construct the delta in the cost between the two treatments and translate the effect
+            current_treatment = orig_df['Current treatment'].values
+            if np.ndim(treatment_costs) >= 2:
```
kbattocchi, line 1588: I don't love all of the array dimensionality munging, weird boolean indexing, etc. here and further below, but I spent a while trying to simplify it without much luck so maybe there's no easy way around it.  Ideally add a TODO to take a closer look at whether we can simplify some of the logic in a later batch of work.

kbattocchi (approved):
Added some minor suggestions but looks good

PR #505, opened 2021-08-04
Causal analysis improvements

kbattocchi:


econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -138,25 +138,30 @@ def _first_stage_reg(X, y, *, automl=True, random_state=None, verbose=0):
...
 def _first_stage_clf(X, y, *, make_regressor=False, automl=True, min_count=None, random_state=None, verbose=0):
     if min_count is None:
         min_count = _CAT_LIMIT  # we have at least this many instances
     if automl:
-        model = GridSearchCVList([make_pipeline(StandardScaler(), LogisticRegression(max_iter=1000,
-                                                                                     random_state=random_state)),
+        # use same Cs as would be used by default by LogisticRegressionCV
+        # NOTE: we don't use LogisticRegressionCV inside the grid search because of the nested stratification
+        #       which could affect how many times each distinct Y value needs to be present in the data
+        cs = LogisticRegressionCV().fit(np.empty((10, 1)), 5 * [1] + 5 * [0]).Cs_
```
vsyrgkanis, line 157: To avoid cases where sklearn switches to an adaptive setting of the Cs array based on the values of the input variables, which in this case will be weird, it might be better to just use an explicit log-space array which is implemented internally. We could in principle hard code the same array when we use LogisticRegressionCV() for consistency.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -494,6 +507,20 @@ class CausalAnalysis:
         Degree of parallelism to use when training models via joblib.Parallel
     verbose : int, default=0
         Controls the verbosity when fitting and predicting.
+    cv: int, cross-validation generator or an iterable, default 5
```
vsyrgkanis, line 510: It might be helpful to explain that this is for the cross-fitting causal estimation methods that cross-fit auxiliary models.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -715,15 +745,27 @@ def fit(self, X, y, warm_start=False):
         # assume we'll be able to train former failures this time; we'll add them back if not
         invalid_inds = [(ind, reason) for (ind, reason) in invalid_inds if ind not in new_inds]
 
+        self._has_column_names = True
```
vsyrgkanis, line 748: Why are these lines moved up here?

vsyrgkanis (approved):
LGTM. Only minor comments

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: This was necessary so that `_format_col` could be used to produce the warning messages below, since we need to have access to the column names in order to decide how to format the column

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Good suggestion, I'll update the text.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: Okay, I'll use a fixed log-space array both here and above.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -715,15 +747,27 @@ def fit(self, X, y, warm_start=False):
         # assume we'll be able to train former failures this time; we'll add them back if not
```
vsyrgkanis, line 747: On line 685: you are not scaling the continuous variables when you are running the model selection for the y outcome!

kbattocchi: Fixed

PR #509, opened 2021-08-07
Causal analysis fixes

kbattocchi:
This PR fixes two issues:
* individual policy results should not fail when `hinds` is empty
* scaling the continuous features in X should be done when fitting nuisance models but should not affect the units of the features reflected in policy trees, etc.


econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -340,6 +364,17 @@ def _process_feature(name, feat_ind, verbose, categorical_inds, categories, hete
 
         W = W_transformer.fit_transform(X)
         X_xf = X_transformer.fit_transform(X)
+
+        # HACK: this is slightly ugly because we rely on the fact that DML passes [X;W] to the first stage models
+        #       and so we can just peel the first columns off of that combined array for rescaling in the pipeline
+        X_scaler = ColumnTransformer([('scale', StandardScaler(),
```
vsyrgkanis, line 370: I feel this is going to be very problematic at some point in the future and we will never notice it. We need to find a solution that does not rely on the internals of dml. I fear that if we push this solution now we will never look back :(

vsyrgkanis: Viewing dml as a black box with its public api, it seems to me that a solution should be calling dml with scaled variables, getting back the tree and then using the cate_feature_names which is part of the public api, to rescale back the tree numbers. The last part is hard, which I guess is the reason why you went with this version.

kbattocchi: Yes, I couldn't find a clean way to get that back into the tree (we can manipulate the tree-as-dictionary that we return from `_policy_tree_dict` without too much trouble, but not the actual tree as used by `plot_policy_tree`).

But also what if we later decide to do something else that uses the coefficients rather than getting out the effect (maybe SHAP or something like that)?  Not scaling X seems like the "correct" thing to do to make sure that everything downstream remains straightforward.

kbattocchi: I agree with your concern, however - so I want to try to come up with a test that would break if this assumption is invalidated.  I just haven't done that yet.

vsyrgkanis: A breaking test seems like a good first order solution. Especially if we have a clear text when it fails that explains why we need to ensure what we are testing. 

Another potential approach to the problem: 
- I agree with you that maybe not scaling X but only pipelining a preprocessor to the nuisances is a good thing. 
- We could do a simple modification to our original design where instead of passing a StandardScaler, we create a transformer, that first checks which columns are categorical by running an np.unique and then applies the standard scaler only to the remainder of the columns.
- Then we could use this "ContinuousStandardScaler" as the first transformer in the pipeline and don't change anything else in the code. 
- I think the biggest problem was that we were scaling categoricals and not that we were computing the std within each fold. So this might be a solution that solves the instabilities and doesn't change much as compared to what we originally had. 

However, this solution might be too computationally costly for large datasets as it will be checking the number of unique entries of each column and so not ideal either.

econml/solutions/causal_analysis/_causal_analysis.py
```diff
@@ -8,10 +8,11 @@
...
 import lightgbm as lgb
+from numba.core.utils import erase_traceback
 import numpy as np
 from numpy.lib.function_base import iterable
 import pandas as pd
-from sklearn.base import TransformerMixin
+from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.compose import ColumnTransformer
 from sklearn.ensemble import GradientBoostingClassifier, RandomForestClassifier, RandomForestRegressor
 from sklearn.linear_model import Lasso, LassoCV, LogisticRegression, LogisticRegressionCV
```
vsyrgkanis, line 18: _CAT_LIMIT seems undefined!

vsyrgkanis: The final logistic regression model that we return in the case of "linear", does not have `max_iter=1000`

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: Also what if we want to add say a LinearDML with a 2-degree featurizer. Then this design will fail. 

Also suppose we add a SparseLinearDML. Then having a standardized X even in the final stage model, will be a good thing, while now we will be using the un-normalized in the final stage.

kbattocchi: I'd be happy to switch to the ContinuousStandardScaler idea, but I believe that would imply that the scaling would then be done using only the subset of data in the cross-fit split rather than on the entire data set.  As long as we're okay with that, then I can switch to that implementation.

econml/solutions/causal_analysis/_causal_analysis.py
kbattocchi: > _CAT_LIMIT seems undefined!

It is: https://github.com/microsoft/EconML/blob/80ea4d5d48a04b16c651624725a263cf327ff8c7/econml/solutions/causal_analysis/_causal_analysis.py#L492-L494

kbattocchi: > The final logistic regression model that we return in the case of "linear", does not have `max_iter=1000`

Fixed in next push.

econml/solutions/causal_analysis/_causal_analysis.py
vsyrgkanis: Seems a reasonable alternative. Maybe we should make sure that it yields the same stability and if so seitch to that?

vsyrgkanis: Let's just go with your current solution. I find equally weird the checking of unique elements per column. We should just add a test that fails if someone ever changes the order that the variables are passed to nuisances. We could for instance create an X that has variance 1 and W that has variance 1million. Then if the order of X and W is changed, X will be divided by 1million and if y = T + X + .1*W, then most prob lasso will only choose X and not W and will yield biased results. Sth along these lines. 

An even better solution, that might actually be helpful down the line:
we can add to our DML api sth like:
model_y_feature_names(X_feature_names, W_feature_names)
model_t_feature_names(X_feature_names, W_feature_names)

This will return the inputs to model_y and model_t respectively. 

This can be helpful if people want to understand what are the important predictive variables.

This also then ties the DML hands, so that it commits to the public what the inputs to model_y and model_t are.

Then we can add a test that X always goes first, as part of the TestCausalAnalysis suite. 

We could do that post this release, but add a TODO for our 0.12 release.

vsyrgkanis: Even easier test: creare a transformer that zeros out all variables after the first n_x variables. So it zeros out W. (Say also that n_w =n_x)

pass an example where W is irrelevant and X is confounder.

as long as dml doesnt change the order of the inputs, then things should be good. Otherwise X would be zeroed out and results will be imvalid

vsyrgkanis: Thiugh I’d also add a todo to add the model_y_feature names api too.

we could even add a model_y_transformer api that takes as input X,W and returns the variables that go into model_y.

should be very useful for people to debug and understand nuisance models

vsyrgkanis (approved):
LGTM

PR #546, opened 2021-12-12
Enable newer versions of python

kbattocchi:


kbattocchi:
I've merged these changes into #518 rather than using this PR.

PR #568, opened 2022-02-04
updates dependency dowhy to v0.7

xrowan:
* Updates dependency requirements to allow for use of `dowhy` up to (and including) version 0.7, which is the latest version of that package
* Version 0.7 `dowhy` package changed the method for calculating a refutation estimate with the ([random common cause method](https://github.com/microsoft/dowhy/blob/master/dowhy/causal_refuters/random_common_cause.py) compared to version 0.6. The relevant change is that the method now runs simulations to estimate an average effect of the randomly generated cofounder. 
* This change required explitly specifying a reasonable (3-10) number of simulations in several places where EconML calls `refute_estimate` with the `random_common_cause` method to avoid using the default number of simulations (100) which leads to long running tests (> 1hr).


xrowan (commented):
I noticed there is already instructions in the notebooks to increase the number of simulations when calculating the refutation estimate.

kbattocchi (approved):
LGTM

PR #571, opened 2022-02-08
Tests running workflows with GitHub Actions

xrowan:


.github/workflows/main-workflow.yml
```diff
@@ -0,0 +1,259 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - id: check_changes
+        run: |
+          echo "=============== list modified files ==============="
+          editedFiles=`git diff --name-only HEAD^ --name-only`
```
kbattocchi, line 20: On Azure Pipelines, the commit that is checked out is the merge commit for the PR as described [here](https://stackoverflow.com/questions/63594658/git-refs-merge-vs-head-in-pull-request), the first parent of which is the target branch (so `HEAD^` gets the correct thing to diff against).

It looks like GitHub actions doesn't do things quite the same way, so I think you'll need to use something like this instead (based on https://stackoverflow.com/questions/62331829/how-to-get-the-target-branch-of-the-github-pull-request-from-an-actions):

 ```suggestion
           editedFiles=`git diff --name-only ${{ github.event.pull_request.base.ref }}`
```

**Comment by xrowan** [comment]
It is currently not possible to re-run a single job (or all failed jobs). 
https://github.com/github/roadmap/issues/271

File: .github/workflows/main-workflow.yml [hunk 2]
```diff
@@ -0,0 +1,258 @@
...
+      - name: Install Jupyter
+        run: pip install jupyter jupyter-client
+      - name: Install nb
+        run: pip install nbconvert nbformat 
+      - name: Install misc.   
+        run: pip install seaborn xgboost tqdm
+      - name: Unit Tests
+        run: python setup.py pytest
+        env: 
+          PYTEST_ADDOPT: '-m "notebook"'
```
kbattocchi, line 141:
```suggestion
          PYTEST_ADDOPTS: '-m "notebook"'
```

.github/workflows/main-workflow.yml
```diff
@@ -0,0 +1,258 @@
...
+      # shap 0.40 has a bug in waterfall (https://github.com/slundberg/shap/issues/2283) that breaks our main tests
+      # but fixes the interaction here...
+      - name: Upgrade shap
+        run: pip install -U shap~=0.40.0
+      - name: Install test dependencies
+        run: pip install pytest pytest-runner jupyter jupyter-client nbconvert nbformat seaborn xgboost tqdm
+      - name: Unit Tests
+        run: python setup.py pytest
+        env: 
+          PYTEST_ADDOPT: '-m "notebook"'
```
kbattocchi, line 171:
```suggestion
          PYTEST_ADDOPTS: '-m "notebook"'
```

kbattocchi:
Closing in favor of #746

PR #575, opened 2022-02-17
Profiling Unit Tests 

xrowan:


econml/tests/test_driv.py
```diff
@@ -1,161 +1,177 @@
...
+                                  est=est):
+
+                    # TODO: serializing/deserializing for every combination -- is this necessary?
+                    # ensure we can serialize unfit estimator
+                    # pickle.dumps(est)
+
+                    est.fit(y, T, Z=Z, X=X, W=W)
+
+                    # ensure we can serialize fit estimator
+                    # pickle.dumps(est)
```
kbattocchi, line 135: We should always ensure we can serialize the fit estimator - different settings will result in different attributes being set (e.g. for DML, with a discrete treatment we create a one-hot encoder, and we also combine this with the featurizer if that is not None)

```suggestion
                    pickle.dumps(est)
```

kbattocchi: Note that we are not keeping the result around, we just verify that calling `pickle.dumps` does not fail, so hopefully this should not result in too much extra memory usage since the resulting string should be eligible for garbage collection right away.

pyproject.toml
```diff
@@ -9,11 +9,14 @@ requires = [
 build-backend = "setuptools.build_meta"
 
 [tool.pytest.ini_options]
-addopts = "--junitxml=junit/test-results.xml -n auto --strict-markers --cov-config=setup.cfg --cov=econml --cov-report=xml"
+# addopts = "--junitxml=junit/test-results.xml -n auto --strict-markers --cov-config=setup.cfg --cov=econml --cov-report=xml"
```
kbattocchi, line 12: Why is this being commented out? As a result, test results don't show up in the Azure Pipelines results, and code coverage results are not being generated and published.

azure-pipelines.yml
```diff
@@ -270,6 +270,43 @@ jobs:
...
+      - script: 'pip install pytest pytest-runner && python setup.py pytest'
+        displayName: 'Unit tests'
+        env:
+          PYTEST_ADDOPTS: '-m "cate_api" -n auto'
+          COVERAGE_PROCESS_START: 'setup.cfg'
+      - script: 'python setup.py pytest'
+        displayName: 'Unit tests'
+        env:
+          PYTEST_ADDOPTS: '-m "cate_api_dml" -n auto'
+          COVERAGE_PROCESS_START: 'setup.cfg'
```
kbattocchi, line 297: Is there a problem with running these all together instead of one after the other?  If not, you could use `-m "cate_api or cate_api_dml"` in a single call to pytest.

econml/tests/test_causal_analysis.py
```diff
@@ -670,21 +674,27 @@ def test_random_state(self):
...
-                    glo = ca.global_causal_effect()
 
-                    ca2 = CausalAnalysis(inds, cats, hinds, classification=classification,
-                                         nuisance_models=n_model, heterogeneity_model=h_model, random_state=123)
-                    ca2.fit(X, y)
-                    glo2 = ca.global_causal_effect()
+        for n_model, h_model, classification in\
+            itertools.product(['linear', 'automl'],
+                              ['linear', 'forest'],
+                              [True, False]):
```
kbattocchi, line 681: I think this (and the related change in test_driv) definitely improves readability - good call.

kbattocchi: Consider doing the same for test_dmliv as well.

kbattocchi (changes requested):
Mostly looks good, but we need to re-enable the pytest flags that make the tests show up in Azure, at least until moving everything to GitHub Actions.

pyproject.toml
xrowan: Ah, I just wanted to test running a single test and I wasn't able to do it with those options.

azure-pipelines.yml
xrowan: Makes sense.

kbattocchi (approved):

PR #587, opened 2022-03-11
Update README.md

kbattocchi:
Update readme to mention upcoming workshop


ewdillon (approved):
switch "solved using DoWhy+EconML" to "solved using DoWhy and/or EconML"

PR #596, opened 2022-04-01
Fix RScorer circular import

fverac:


kbattocchi:
If possible, add a test that would have failed prior to the fix.

kbattocchi (approved):
Looks good.

PR #598, opened 2022-04-04
Replace Boston housing analysis with Ames housing

fverac:


notebooks/Solutions/Causal Interpretation for Ames Housing Price.ipynb
```diff
@@ -0,0 +1,3055 @@
...
+    "from econml.solutions.causal_analysis import CausalAnalysis\n",
+    "\n",
+    "# initialize heterogeneity features\n",
+    "hetero_cols = ['AgeAtSale', 'MSZoning']\n",
+    "ca = CausalAnalysis(\n",
+    "    feature_inds= [\n",
+    "        'HasFireplace',\n",
+    "        'HasPorch',\n",
+    "        'HasDeck',\n",
+    "    ],\n",
```
kbattocchi, line 2114: Why these three features?  Overall, this new notebook is good, but one thing that's missing compared to the old one is any finding that one of top predictive features may not be causally important.

Maybe the dataset doesn't support such a finding, and that would be fine to point out too (something like "We find that not only are features X, Y, and Z predictive of high prices, but that even when accounting for other features they appear to be causally linked.")  But right now there doesn't seem to be a connection between the predictive and causal analysis.

kbattocchi, line 2114: Why these three features?  Overall, this new notebook is good, but one thing that's missing compared to the old one is any finding that one of top predictive features may not be causally important.

Maybe the dataset doesn't support such a finding, and that would be fine to point out too (something like "We find that not only are features X, Y, and Z predictive of high prices, but that even when accounting for other features they appear to be causally linked.")  But right now there doesn't seem to be a connection between the predictive and causal analysis.

kbattocchi (changes requested):
Sorry for the delay in getting this reviewed.  Overall, I think it's really good, but I think there's room to improve the connection between the predictive and causal analyses.

notebooks/Solutions/Causal Interpretation for Ames Housing Price.ipynb
```diff
@@ -2041,9 +2049,16 @@
     "### Direct Causal Effect --  Do certain house characteristics have a direct effect on home value?"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We've seen which features are good predictors of home value, but are these features also causally signficiant? Below we use the CausalAnalysis class to assess the causal effects of certain house characteristics on home value."
```
kbattocchi, line 2056: This is good, but it would be nice to have a parallel statement after calling CausalAnalysis which summarizes whether it agrees or disagrees with the predictive analysis.

kbattocchi (approved):
Looks great, thanks!

Issue #185, opened 2019-11-22
StatsModelsLinearRegression needs check_input


moprescu:
If the numpy arrays passed are of type `object` and not `float`, then an exception is thrown in `param, _, rank, _ = np.linalg.lstsq(WX, wy, rcond=None)`

kbattocchi:
Simple reproduction:

```
from econml.sklearn_extensions.linear_model import StatsModelsLinearRegression
from sklearn.linear_model import LinearRegression
import numpy as np

X = np.random.normal(size=(100,5)).astype('O')
y = np.random.normal(size=100)

LinearRegression().fit(X,y)
StatsModelsLinearRegression().fit(X,y)
```

PR #599, opened 2022-04-04
improve input processing for StatsModelsLinearRegression

fverac:


econml/utilities.py
```diff
@@ -514,7 +514,7 @@ def check_inputs(Y, T, X, W=None, multi_output_T=True, multi_output_Y=True):
     return Y, T, X, W
 
 
-def check_input_arrays(*args, validate_len=True, force_all_finite=True):
+def check_input_arrays(*args, validate_len=True, force_all_finite=True, dtype=None):
```
kbattocchi, line 517: This is a good way to address the issue.

kbattocchi (changes requested):
Looks good, but please add a test (you can probably just take the repro in #185 and just verify that the point estimates are the same to within a small tolerance for both the sklearn LinearRegression model and ours).

kbattocchi (approved):
Looks good.  I'll rebase on top of main once the test stability fixes are in.

Issue #557, opened 2022-01-07
Using FLAML in tune() methods


EgorKraevTransferwise:
It's great to have tune() methods in some of the model classes (though sadly not all as yet). But they all seem to do tuning via grid search, which is quite wasteful. 
Could you possibly add an option to use FLAML instead of grid search, to search the hyperparameter space - should be quite easy if there's already a spec in the code for the parameter space of each class to be tuned.

kbattocchi:
Thanks for the suggestion - we'll look into it.

EgorKraevTransferwise:
A solution for some usecases (for now, CATE for binary treatments, extensions pending) is implemented in our [auto-causality](https://github.com/transferwise/auto-causality) library, which searches over both individual model's parameters and across models to optimize a given score.
Requests for extending the usecases it supports (via raising issues) are very welcome!

kbattocchi:
I think ideally we should be providing the ranges in an optimizer-agnostic way, so that users can use FLAML if they want, but could use something like hyperopt instead if that suits their environment better.

If there is a need to use FLAML-specific functionality somewhere, then you may be able to try to import it but use the `MissingModule` class from `econml.utilities` in its place if it can't be imported - see e.g. https://github.com/microsoft/EconML/blob/ac12f54f4304c730d1501edaaeae2b96e9af4507/econml/_tree_exporter.py#L18-L28

EgorKraevTransferwise:
How about I wrap the FLAML imports into  `MissingModule` for now, and add an optimizer-agnostic spec as soon as someone raises an issue asking for it?

PR #635, opened 2022-06-15
Add FLAML search spaces to models

EgorKraevTransferwise:
Adressing https://github.com/microsoft/EconML/issues/557
Just did ForestDRLearner to see if you agree with the way I'm doing it, will add the others if so.

Is there a reason you don't support dowhy versions from 0.8?

### DIFF
--- a/econml/dr/_drlearner.py
+++ b/econml/dr/_drlearner.py
@@ -1433,6 +1433,56 @@ class ForestDRLearner(ForestModelFinalCateEstimatorDiscreteMixin, DRLearner):
 
     def _gen_ortho_learner_model_final(self):
         return _ModelFinal(self._gen_model_final(), self._gen_featurizer(), False)
+    
+    @classmethod
+    def search_space(cls) -> dict:
+        """
+        Specify the hyperparameter search space for FLAML
+
+        Returns
+        -------
+        A FLAML-compatible search space for __init__ args we want to search over
+        """
+        return {
+            "min_propensity": tune.loguniform(1e-6, 1e-1),
+            # "mc_iters": tune.randint(0, 10), # is this worth searching over?
+            "n_estimators": tune.randint(2, 200),
+            # "max_depth": is tune.choice([None, tune.randint(2, 1000)]) the right syntax?
+            "min_samples_split": tune.randint(2, 20),
+            "min_samples_leaf": tune.randint(1, 25),
+            "min_weight_fraction_leaf": tune.uniform(0, 0.5),
+            "max_features": tune.choice(["auto", "sqrt", "log2", None]),
+            "min_impurity_decrease": tune.uniform(0, 10),
+            "max_samples": tune.uniform(0, 0.5),
+            "min_balancedness_tol": tune.uniform(0, 0.5),
+            "honest": tune.choice([0, 1]),
+            "subforest_size": tune.randint(2, 10),
+        }
+
+    @classmethod
+    def search_start(cls) -> dict:
+        """
+        Specify the default start point for hyperparameter search
+
+        Returns
+        -------
+        A dict of hyperparameter values consistent with the output from search_space()
+        """
+
+        return {
+            "min_propensity": 1e-6,
+            "n_estimators": 100,  # default is 1000 but that seems excessive
+            "min_samples_split": 5,
+            "min_samples_leaf": 5,
+            "min_weight_fraction_leaf": 0.0,
+            "max_features": "auto",
+            "min_impurity_decrease": 0.0,
+            "max_samples": 0.45,
+            "min_balancedness_tol": 0.45,
+            "honest": True,
+            "subforest_size": 4,
+        }
+
 
     def fit(self, Y, T, *, X=None, W=None, sample_weight=None, groups=None,
             cache_values=False, inference='auto'):
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,8 +39,9 @@ install_requires =
     statsmodels >= 0.10
     pandas
     shap >= 0.38.1, < 0.41.0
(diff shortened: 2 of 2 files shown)


immu4989:
Thanks for putting this up @EgorKraevTransferwise , I would like to help unblock it.
I am am working in the FLAML codebase concurrently and was looking at how its
search space API gets consumed downstream, which led me here. A few things
worth raising before any further code work, since the PR is blocked at a
direction question that never got answered in 2022.

### Direction question (for @kbattocchi)

In [#557](https://github.com/py-why/EconML/issues/557) you asked for an
optimizer-agnostic spec so users could plug in hyperopt as easily as FLAML.
The PR as filed is FLAML-specific. Before Egor invests more in this branch,
is your guidance still optimizer-agnostic? Two viable shapes:

1. **Keep it FLAML shaped now**, accept the tight coupling, document that the
   API is provisional (e.g., `_search_space` rather than `search_space`).
2. **Optimizer agnostic spec** — return a plain dict like
   `{"n_estimators": {"type": "randint", "low": 2, "high": 200}, ...}` and
   ship a small `econml.tuning.flaml_adapter` translator. More work but
   matches your 2022 ask and avoids hard-pinning FLAML.

If (2), the PR should move the API onto a mixin / base class so it lands
once and gets inherited by `LinearDML`, `CausalForestDML`, `DRLearner`,
etc., rather than being defined per-estimator.

### Must fix correctness (regardless of direction)

1. **`tune` is not imported in the diff.** `econml/dr/_drlearner.py` uses
   `tune.loguniform`, `tune.randint`, etc., but the patch contains no
   `from flaml import tune`. The classmethods would raise `NameError` on
   first call. (Whatever direction you take above, the import must land
   somewhere.)

2. **`max_samples: tune.uniform(0, 0.5)`.** sklearn's `max_samples` requires
   the open interval `(0, 1]`; sampling exactly `0` will raise. Suggest
   `tune.uniform(0.05, 0.95)` (or `0.05, 0.5` if there's a reason to cap at
   half — but justify the cap, sklearn's own default is `None` ≈ 1.0).

3. **Patch targets `setup.cfg`.** EconML migrated to `pyproject.toml` (see
   [#1538](https://github.com/py-why/EconML/pull/1538)-era cleanup);
   `setup.cfg` no longer drives dependency resolution. Move the FLAML add
   into `pyproject.toml`'s `dependencies` (or `[project.optional-dependencies]`
   under a `tuning` extra — see point 4).

4. **Hard FLAML dependency.** Your own [June 2022
   reply](https://github.com/py-why/EconML/issues/557#issuecomment-1162197474)
   promised wrapping the import via `econml.utilities.MissingModule` to keep
   FLAML optional. The PR currently makes it a hard `install_requires`. With
   FLAML's transitive footprint (xgboost, lightgbm, sklearn pins), making it
   mandatory will degrade install ergonomics for users who don't tune.
   Either `MissingModule` + lazy import, or move FLAML under a `tuning`
   extra.

5. **Stale base + merge conflict.** PR opens against `ac12f54f` (June 2022);
   GitHub reports `mergeable_state: dirty`. `_drlearner.py` has had
   substantial churn since (notably the `model_selection`-based `'auto'`
   nuisance refactor). Needs rebase before further review.

6. **No tests.** A new public classmethod exposed on `ForestDRLearner` with
   no coverage. Suggest minimal: instantiate, call `cls.search_space()`,
   sample one config from each domain, pass it through `cls(**search_start)`
   to confirm every key is a valid `__init__` arg.

### Design feedback

- **`min_impurity_decrease: tune.uniform(0, 10)`.** Upper bound of 10 is far
  outside the practical range — sklearn's default is `0.0`, and values
  above `~0.1` typically prune trees to stumps. Suggest `tune.loguniform(1e-4, 1e-1)`
  or `tune.uniform(0, 0.1)`. Worth comparing against ranges used in
  established AutoML libs (auto-sklearn, optuna's sklearn integration).

- **`honest: tune.choice([0, 1])` vs `search_start: honest: True`.**
  Type mismatch — sampler returns `int`, start returns `bool`. EconML
  passes the value through to a constructor expecting `bool`; mostly OK
  via Python's truthiness, but cleaner as `tune.choice([False, True])`.

- **`subforest_size: tune.randint(2, 10)` searched independently of
  `n_estimators`.** ForestDRLearner expects `n_estimators % subforest_size == 0`
  (see `RegressionForest`); independent sampling will hit invalid configs.
  Either constrain via a conditional, or document the invariant and let
  the constructor raise.

- **Commented-out lines.** `mc_iters` and `max_depth` carry your "is this
  worth searching?" / "is this the right syntax?" notes. Decide before
  merge:
  - `mc_iters`: typically a fidelity dial, not a tuning knob — leave out.
  - `max_depth`: FLAML's `tune.choice([None, tune.randint(2, 1000)])` is
    not valid (`tune.choice` expects concrete values). Use a single
    `tune.randint(2, 50)` and accept that the unbounded `None` case isn't
    searched (or use a sentinel like `0` and translate).

### Smaller things

- `search_start: n_estimators: 100` vs. constructor default `1000`. Reasonable
  for a *search* starting point (faster), but worth a one-line docstring
  saying "intentionally lower than the model default to keep search budgets
  bounded."

- `tune.choice(["auto", "sqrt", "log2", None])` — `"auto"` is sklearn-1.3-deprecated
  for `RandomForestRegressor`, but EconML's `RegressionForest` may still
  accept it (it routes to `n_features` per the docstring). Worth a quick
  test that all four values still work end-to-end under the project's
  current sklearn pin (`>=1.0,<1.9`).

### Offer

If direction (1) or (2) gets confirmed, happy to take the rebase + the
items above as a follow up PR (with credit to the original work) or
leave it with you, whichever you prefer.

PR #638, opened 2022-06-21
fix _preprocess_data to align with sklearn refactor

fverac:
Making changes to some preprocessing in our sklearn extensions to align with sklearn refactoring. These are necessary changes to align with sklearn 1.1+.

Relevant sklearn changes here: https://github.com/scikit-learn/scikit-learn/pull/22762


setup.cfg
```diff
@@ -33,7 +33,7 @@ packages = find_namespace:
 install_requires =
     numpy
     scipy > 1.4.0
-    scikit-learn > 0.22.0, < 1.1
+    scikit-learn > 0.22.0
```
kbattocchi, line 36: Overall, this looks good, but I'd just bump the upper limit:

```suggestion
    scikit-learn > 0.22.0, < 1.2
```

so that if there are further breaking changes in 1.2 we can handle them accordingly and then update the limit again.  We tend not to put upper bounds on dependencies that rarely break us for convenience, but sklearn seems to cause us to break somewhat more often (probably because we rely on some internals) so I think it's worth being more conservative here.

kbattocchi (changes requested):
Mostly looks good.  One minor but important suggestion.

PR #653, opened 2022-07-11
update econml to shap v0.41.0

imatiach-msft:
Update econml to support shap v0.41.0.
Also updates the CI tests to use latest shap 0.41.0.


imatiach-msft:
nvm it looks like it's already on 0.41.0!

PR #660, opened 2022-08-05
Update _interpreters.py support criterion mse and squared_error in different sklearn version

00helloworld:
support criterion `mse` and `squared_error` in different sklearn version


kbattocchi:
Thanks, I've cherry-picked this commit into #712

PR #662, opened 2022-08-12
Create panel subpackage

kbattocchi:
This creates an econml.panel package, adds a few utilities to it, and moves DynamicDML from econml.dynamic.dml to econml.panel.dml


fverac:
Do we want to restructure/relabel some of our docs now to include a panel section?

I believe at the moment we currently have a "Dynamic Treatment Regime" section, but maybe that can be housed under a broader "Panel Data" section?
https://econml.azurewebsites.net/spec/estimation_dynamic.html

econml/panel/utilities.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def long(x):
+    """
+    Reshape panel data to long format, i.e. (n_units * n_periods, d_x) or (n_units * n_periods,)
+
+    Parameters
+    ----------
+    x : array-like
+        Panel data in wide format
```
fverac, line 12: I think it could be helpful to be more specific about what we mean by wide format. 

Does this mean x can be either (n_units, n_periods, d_x) shaped or (n_units, n_periods)?

The two examples I see for wide format in the `wide` docstring are 2-dim arrays, so it's not clear to me why we need to deal with the case when x is not 2 dimensions.

fverac:
> Do we want to restructure/relabel some of our docs now to include a panel section?
> 
> I believe at the moment we currently have a "Dynamic Treatment Regime" section, but maybe that can be housed under a broader "Panel Data" section? https://econml.azurewebsites.net/spec/estimation_dynamic.html

Bumping this. It seems to me like it could be worthwhile to rename the "Estimation Methods for Dynamic Treatment Regimes" section in our docs to something like "Estimation Methods for Panel Data" in here since this PR is sort of prepping for future panel methods.

fverac (approved):

fverac (approved):

kbattocchi:
I've renamed the documentation section as suggested.

PR #672, opened 2022-09-14
update some pipeline images (macos-12, ubuntu20), fix pycodestyle E275

fverac:
macos-10.15 will soon be deprecated in GitHub Actions and Azure Devops so we are upgrading the version of macOS that we use.

Relevant issue:

https://github.com/actions/runner-images/issues/5583

Additionally updating the deprecated ubuntu 18 image to ubuntu 20.

Fixing multiple new E275 style errors with new pycodestyle update.

### DIFF
--- a/azure-pipelines-steps.yml
+++ b/azure-pipelines-steps.yml
@@ -5,7 +5,7 @@
 
 parameters:
   package: '-e .'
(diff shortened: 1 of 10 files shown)


kbattocchi (approved):
Looks good.  Consider adding a comment to the PR description indicating that the spacing changes are due to a pycodestyle update in case we ever need to go back through the history of those lines.

PR #680, opened 2022-09-20
add matplotlib version ceiling

fverac:
It seems that a recently released matplotlib version has broken some things for the shap library, which in turn causes some of our tests/functionality to fail.

Capping the version to avoid for now.

Relevant issues:
https://github.com/slundberg/shap/issues/2687
https://github.com/SeldonIO/alibi/issues/774

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -68,7 +68,8 @@ all =
     tensorflow > 1.10, < 2.3
     ; Version capped due to tensorflow incompatibility
     protobuf < 4
-    matplotlib
+    ; Version capped due to shap incompatibility
+    matplotlib < 3.6.0
     
 [options.packages.find]
 include =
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -69,7 +69,8 @@ all =
     tensorflow > 1.10, < 2.3
     ; Version capped due to tensorflow incompatibility
     protobuf < 4
-    matplotlib
+    ; Version capped due to shap incompatibility
+    matplotlib < 3.6.0
```
kbattocchi, line 73: Consider using 

```suggestion
    matplotlib != 3.6.0
```

if this seems like a matplotlib issue that will be fixed soon.

kbattocchi, line 73: Note that whatever change you make here, the same change should also be made to the `plt` extra.

kbattocchi (approved):
LGTM

PR #705, opened 2022-11-21
create standard dgp for metric aggregation

fverac:


econml/data/dgps.py
```diff
@@ -93,3 +94,182 @@ def _process_ihdp_sim_data():
...
+        self.discrete_instrument = discrete_instrument
+        self.squeeze_T = squeeze_T
+        self.squeeze_Y = squeeze_Y
+
+        if callable(nuisance_Y):
+            self.nuisance_Y = nuisance_Y
+        else:  # else must be dict
+            if nuisance_Y is None:
+                nuisance_Y = {'support': self.d_x, 'degree': 1}
+            nuisance_Y['k'] = self.d_x
```
kbattocchi, line 135: Consider asserting the expected type here, something like:

```suggestion
            assert(isinstance(nuisance_Y, dict), f"nuisance_Y must be a callable or dict, but got {type(nuisance_Y)}")
            nuisance_Y['k'] = self.d_x
```

(and likewise for the other similar arguments)

econml/data/dgps.py
```diff
@@ -93,3 +94,182 @@ def _process_ihdp_sim_data():
...
+            y_of_t['k'] = self.d_t
+            self.y_of_t, self.y_of_t_coefs = self.gen_nuisance(**y_of_t)
+
+        self.x_eps = x_eps
+        self.y_eps = y_eps
+        self.t_eps = t_eps
+
+    def gen_Y(self):
+        self.y_noise = np.random.normal(size=(self.n, self.d_y), scale=self.y_eps)
+        self.Y = self.theta(self.X) * self.y_of_t(self.T) + self.nuisance_Y(self.X) + self.y_noise
```
kbattocchi, line 179: It looks like y_of_t is really a treatment featurizer, is that right?  If so, change the name accordingly.

econml/data/dgps.py
kbattocchi, line 135: Setting `nuisance_Y['k']` is not strictly necessary since the default behavior of `gen_nuisance` will do this for you, but perhaps it's worth being explicit.

More importantly, making this assignment will mutate the dictionary that was passed in, which is possibly undesirable; probably better to do something like: 
```suggestion
            nuisance_Y = { **nuisance_Y, 'k':self.d_x }
```
which will create a new dictionary instead of altering the old one.

Furthermore, what if the dictionary already had a `'k'` entry - do you really want to ignore and override it?

econml/data/dgps.py
```diff
@@ -93,3 +94,182 @@ def _process_ihdp_sim_data():
...
+        return self.Y
+
+    def gen_X(self):
+        self.X = np.random.normal(size=(self.n, self.d_x), scale=self.x_eps)
```
kbattocchi, line 183: Consider allowing heavier-tailed distributions than just normal (or incorporate outliers in some other way), here and  wherever else you're using randomness.

econml/data/dgps.py
```diff
@@ -93,3 +94,182 @@ def _process_ihdp_sim_data():
...
+        else:
+            self.T = self.nuisance_T(self.X) + self.nuisance_TZ(self.Z) + self.T_noise
+            return self.T
+
+    def gen_Z(self):
+        if self.d_z:
+            if self.discrete_instrument:
+                # prob_Z = expit(np.random.normal(size=(self.n, self.d_z)))
+                # self.Z = np.random.binomial(1, prob_Z, size=(self.n, 1))
+                # self.Z = np.random.binomial(1, prob_Z)
```
kbattocchi, line 204: Remove dead code unless it has a purpose (like presenting an alternative we might want to use in some specific scenario, in which case also document what that purpose is), in which case also document what that purpose is

econml/data/dgps.py
```diff
@@ -93,3 +94,182 @@ def _process_ihdp_sim_data():
...
+                y_of_t = {'support': self.d_t, 'degree': 1, 'bounds': [1, 1]}
+            y_of_t['k'] = self.d_t
+            self.y_of_t, self.y_of_t_coefs = self.gen_nuisance(**y_of_t)
+
+        self.x_eps = x_eps
+        self.y_eps = y_eps
+        self.t_eps = t_eps
+
+    def gen_Y(self):
+        self.y_noise = np.random.normal(size=(self.n, self.d_y), scale=self.y_eps)
```
kbattocchi, line 178: Consider allowing the user to specify a seed so that results are reproducible.

econml/data/dgps.py
```diff
@@ -93,3 +94,182 @@ def _process_ihdp_sim_data():
     # Append a column of ones as intercept
     X = np.insert(X, 0, np.ones(X.shape[0]), axis=1)
     return T, X
+
+
+class StandardDGP():
```
kbattocchi, line 99: This should probably support generating and using W as well.

kbattocchi, line 99: This should probably support using a user-specified featurizer on X.

kbattocchi, line 99: Add tests to at least verify that the API can actually be used and that you get arrays of the right dimensions out, etc., even if there aren't good checks for the actual data (but perhaps there are useful checks there, too)

kbattocchi, line 99: Do we expect users to use this class to test things?  If so, perhaps also add a DGP section to the documentation (and at least add it to the module docs).  But if this is mainly for internal use, add a comment to the docstring indicating that instead.

kbattocchi (changes requested):
Looks like a good start but there are a few things that should be cleaned up.

Issue #708, opened 2022-12-13
Obtaining [T0][T1] or [Y0][T0] in causalforest shap values


fobembe:
I am trying to implement the shap plots to obtain the relevant factors for my predictive analytics.  I have done this with causalml without any problem, but I have been struggling to get this done with econml.  Does anyone have a clue on what I am doing wrongly in the following code?

I'll appreciate any help. Thanks.

![image](https://user-images.githubusercontent.com/38144116/207242725-95ffc850-a2b3-4dc6-bf88-2f7527521cbf.png)

vsyrgkanis:
Most prob you have a binary treatment and you need to use the value of the treatment itself as keyword. You can print the keys of the dictionary shap_values[Y0] to see what they right second keyword is instead of T0

vsyrgkanis:
Check cells 6,7,8 here
https://github.com/microsoft/EconML/blob/main/notebooks/Interpretability%20with%20SHAP.ipynb

i think you need to pass T0_1 if your treatments where taking values 0/1 and you are using the drlearner, or metalearners

fobembe:
Thanks a lot for your reply.  I am using causal forest and I used the following but still have error
![image](https://user-images.githubusercontent.com/38144116/207265658-c55393bb-c0a6-4323-a700-bf924fa921f2.png)

fobembe:
![image](https://user-images.githubusercontent.com/38144116/207369648-d67bd2ae-1628-4f67-b151-8b9448616f64.png)
![image](https://user-images.githubusercontent.com/38144116/207369790-216cec70-2d7d-4f78-9496-b494a2571d57.png)

fobembe:
I really appreciate this.  My treatment is binary 0/1 and my outcome is continuous (Sales).  I reran the whole code to provide a complete picture of what is happening.  Thanks again for your assistance.


![image](https://user-images.githubusercontent.com/38144116/207428024-a8b58b75-8443-463b-9610-4eaa32eebb5a.png)
![image](https://user-images.githubusercontent.com/38144116/207427229-8bcdbe24-537b-4711-943a-e79396f4ae57.png)
![image](https://user-images.githubusercontent.com/38144116/207427439-32802d57-4a18-4a5b-9d1e-cc741558987a.png)
![image](https://user-images.githubusercontent.com/38144116/207427539-0789c94f-8a6a-47d6-bcf7-fd66e9830bf3.png)

fobembe:
Pip list
package                       Version
[...]

PR #709, opened 2022-12-13
Avoid returning defaultdict directly

kbattocchi:
This should address the usability issue that led to #708


fverac:
Looks good to me.

fverac (approved):

PR #716, opened 2023-01-13
Bootstrapping Final Model

S-Karnik:
This PR will add the capability to bootstrap the final model only for estimators with final models.


econml/inference/_bootstrap.py
```diff
@@ -86,44 +94,76 @@ def fit(self, *args, **named_args):
...
-                         **{arg: convertArg(named_args[arg], inds) for arg in named_args})
-            for obj, inds in zip(self._instances, indices)
-        )
+            if isinstance(arg, tuple):
+                converted_arg = []
+                for arg_param in arg:
+                    converted_arg.append(convertArg_(arg_param, inds))
+                return tuple(converted_arg)
+            return convertArg_(arg, inds)
+
```
kbattocchi, line 121: I think it's probably worth reworking the API here a bit to pull this and some of the neighboring code out to a a new top-level function in this module (say, `fit_with_subsets(clones, strata, *args, **kwargs)`) that calls fit on each clone in parallel with the correctly modified arguments.  This could then be used both here as well as directly within OrthoLearner for the final_only case.

econml/inference/_bootstrap.py
```diff
@@ -139,8 +179,16 @@ def __getattr__(self, name):
...
+                instance_results = []
+                obj = clone(self._wrapped, safe=False)
+                for i in range(self._n_bootstrap_samples):
+                    if self._only_final:
+                        obj._set_current_cloned_ortho_learner_model_final(i)
+                    else:
+                        obj = self._instances[i]
+                    instance_results.append(f(obj, name))
+                instance_results = np.array(instance_results)
+                results = instance_results, f(self._wrapped, name)
```
kbattocchi, line 191: I think it would be better to continue to use parallelism here if possible.

kbattocchi (changes requested):
Overall, the progress here is great.  However, there are a number of small items to address.

econml/_ortho_learner.py
```diff
@@ -794,21 +817,59 @@ def _fit_nuisances(self, Y, T, X=None, W=None, Z=None, sample_weight=None, group
...
+            def convertArg(arg, inds):
+                def convertArg_(arg, inds):
+                    arr = np.asarray(arg)
+                    if arr.ndim > 0:
+                        return arr[inds]
+                    else:  # arg was a scalar, so we shouldn't have converted it
+                        return arg
+                if arg is None:
+                    return None
+                if isinstance(arg, tuple):
```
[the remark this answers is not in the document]
S-Karnik: The purpose of the handling of tuples is that the nuisances are passed in as tuples, which means that the individual nuisances in the tuple need to also be indexed.

econml/inference/_bootstrap.py
S-Karnik: Based on our discussion, it seems like keeping this in `_bootstrap.py` and using parallel processes or threads will not be possible. In the future, we would like to make this call in `_ortho_learner.py`, just as we do with fit.

S-Karnik:
> @microsoft-github-policy-service agree company="Microsoft Research"

@microsoft-github-policy-service agree company="Microsoft Research"

PR #750, opened 2023-03-21
Minor fixes

kbattocchi:


.github/workflows/ci.yml
```diff
@@ -224,7 +226,7 @@ jobs:
     steps:
     - run: exit 1
       name: At least one check failed or was cancelled
-      if: ${{ !(success()) }}
+      if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}
```
fverac, line 229: I don't understand the reason for the change here, could you explain?

econml/_ortho_learner.py
```diff
@@ -899,7 +899,7 @@ def score(self, Y, T, X=None, W=None, Z=None, sample_weight=None, groups=None):
                     nuisances = [np.zeros((n_iters * n_splits,) + nuis.shape) for nuis in nuisance_temp]
 
                 for it, nuis in enumerate(nuisance_temp):
-                    nuisances[it][i * n_iters + j] = nuis
+                    nuisances[it][j * n_iters + i] = nuis
```
fverac, line 902: Should we link this to the inciting issue.

kbattocchi: The name of that commit mentions it, and I'll rebase rather than squash when merging so that it will be maintained.

.github/workflows/ci.yml
kbattocchi: At the step level, success is checking that each previous step succeeded, not that the required jobs succeeded, see e.g. https://github.com/py-why/EconML/actions/runs/4461230788/jobs/7835479732

fverac (approved):

PR #759, opened 2023-04-20
Adding functionality to EconML Package

AnthonyCampbell208:


econml/sklearn_extensions/model_selection.py
```diff
@@ -255,7 +260,112 @@ def get_n_splits(self, X, y, groups=None):
...
+                    self._search_list.append(temp_search)
+                else:
+                    if is_polynomial_pipeline(estimator=estimator):
+                        temp_search.fit(X, y, groups=groups, linear__sample_weight=sample_weight)
+                    elif is_mlp(estimator=estimator):
+                        temp_search.fit(X, y, groups=groups)
+                    else:
+                        temp_search.fit(X, y,  groups=groups, sample_weight=sample_weight)
+                    self._search_list.append(temp_search)
+            except (ValueError, TypeError, FitFailedWarning) as e:
```
kbattocchi, line 343: Why only these three errors?  Won't any other errors that are generated cause the same issues?

econml/sklearn_extensions/model_selection.py
```diff
@@ -255,7 +260,112 @@ def get_n_splits(self, X, y, groups=None):
...
+                warning_msg = f"Warning: {e} for estimator {estimator} and param_grid {param_grid}"
+                warnings.warn(warning_msg, category=UserWarning)
+            if not hasattr(temp_search, 'cv_results_'):
```
kbattocchi, line 346: I would have thought that this line will only be hit if the above exception handling code also triggers; in that case it seems like we'll get two similar but non-identical warnings.  Is there a reason to issue them both?

econml/sklearn_extensions/model_selection.py
```diff
@@ -255,7 +260,112 @@ def get_n_splits(self, X, y, groups=None):
...
+                warning_msg = f"Warning: estimator {estimator} and param_grid {param_grid} failed"
+                warnings.warn(warning_msg, category=FitFailedWarning)
+        self.best_ind_ = np.argmax([search.best_score_ for search in self._search_list])
+        self.best_estimator_ = self._search_list[self.best_ind_].best_estimator_
+        self.best_score_ = self._search_list[self.best_ind_].best_score_
+        self.best_params_ = self._search_list[self.best_ind_].best_params_
+        print(f'Best estimator {self.best_estimator_} and best score {self.best_score_} and best params {self.best_params_}')
+        # pdb.set_trace()
```
kbattocchi, line 354: Remove dead code
```suggestion
```

econml/sklearn_extensions/model_selection.py
```diff
@@ -255,7 +260,112 @@ def get_n_splits(self, X, y, groups=None):
...
+
+        # TODO Add in more functionality by checking if it's an empty list. If it's just 1 dictionary then we're going to need to turn it into a list
+        # Just do more cases
+        if param_grid_list == 'auto':
+            self.param_grid_list = auto_hyperparameters(estimator_list=self.complete_estimator_list, is_discrete=is_discrete)
+        elif (param_grid_list == None):
+            self.param_grid_list = len(self.complete_estimator_list) * [{}]
+        else:
+            self.param_grid_list = param_grid_list
+        # self.categorical_indices = categorical_indices
```
kbattocchi, line 279: I think we do need to track categorical indices, since depending on the model we may need to one-hot-encode them and we also should exclude them from scaling.

econml/sklearn_extensions/model_selection_utils.py
```diff
@@ -0,0 +1,434 @@
...
+
+# For classification problems
+models_classification = [
+    LogisticRegressionCV(),
+    RandomForestClassifier(),
+    GradientBoostingClassifier(),
+    MLPClassifier()
+]
+
+scaling_lst =  [StandardScaler(), MinMaxScaler(), RobustScaler(), MaxAbsScaler()]
```
kbattocchi, line 48: This doesn't appear to be used anywhere

econml/sklearn_extensions/model_selection_utils.py
```diff
@@ -0,0 +1,434 @@
...
+
+    Returns:
+    is_linear (bool): True if the estimator is a linear model, False otherwise.
+    """
+
+    # Check if the estimator is a polynomial regression
+    if isinstance(estimator, Pipeline):
+        has_poly_feature_step = any(isinstance(step[1], PolynomialFeatures) for step in estimator.steps)
+        if has_poly_feature_step:
+            return True
```
kbattocchi, line 366: This seems wrong; you're checking whether there is a PolynomialFeatures transformer anywhere in the pipeline, but that doesn't mean it's a linear model, right?  A user could pipeline PolynomialFeatures with a neural network or something.

econml/sklearn_extensions/model_selection_utils.py
```diff
@@ -0,0 +1,434 @@
...
+
+    # Check if the estimator is a linear regression or related model
+    if hasattr(estimator, 'fit_intercept') and hasattr(estimator, 'coef_'):
+        return True
```
kbattocchi, line 370: Again this seems like a roundabout test - I think checking for specific classes is probably more reasonable.

kbattocchi (changes requested):
This is a good start.  I've made some suggestions - most important among them is that you should add tests that cover this new functionality at a low level.  Furthermore, many checks are failing, so take a look at those and try to resolve the issues that are happening there.

kbattocchi:
@AnthonyCampbell208 @ShrutiRM97 I've updated main with the fixes to our build system (although there's one flaky notebook test that I'm still investigating) - please merge main into this branch and then we can see what test failures remain.

kbattocchi:
This was merged via a separate branch/PR

PR #766, opened 2023-05-10
Improve the CI workflow

kbattocchi:


.github/workflows/ci.yml
```diff
@@ -125,35 +139,53 @@ jobs:
...
+
   tests:
     name: "Run tests"
     needs: [eval]
     if: ${{ needs.eval.outputs.testCode == 'True' }}
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: ['3.7', '3.8', '3.9', '3.10']
```
fverac, line 183: Does changing the python versions to be strings do anything?

kbattocchi: They should always have been strings (and in many cases YAML interprets things as strings even without quotes, but since these can be parsed as floats that isn't true here); now it matters because `3.10` as a float is `3.1` which is not what we want as a version of python.

setup.cfg
```diff
@@ -68,8 +68,8 @@ dowhy =
     dowhy < 0.9
 all =
     azure-cli
-    keras < 2.4
-    tensorflow > 1.10, < 2.3
+    keras < 2.4;python_version < '3.9'
```
fverac, line 71: Is it possible to add a comment here explaining the python version cap? Are there relevant issues to link to?

kbattocchi: These lines are just copied from the `tf` section, which does contain such a comment - when the `tf` section was updated I forgot to also make a parallel update here.  But I'll copy a version of the comment here too.

.github/workflows/ci.yml
kbattocchi: There are a few other places in this file where versions are (incorrectly but benignly for now) specified as floats which I'll also change now.

fverac (approved):

PR #781, opened 2023-06-12
Get datasets from web instead of blob

kbattocchi:
Signed-off-by: Keith Battocchi <kebatt@microsoft.com>


kbattocchi:
Cherry picking into #766 instead, closing this PR

PR #783, opened 2023-06-21
Fix authentication for website publishing

kbattocchi:
We now need to use a service principal rather than a publish profile due to changes on the backend.


fverac (approved):
Thanks for figuring this out. 

Do you now understand why it broke in the first place? If so, sharing a relevant link or explanation would be helpful if you have the time.

PR #791, opened 2023-07-14
add arg to allow missing values in W and sometimes X

fverac:
Summary of changes:

* Add enable_missing arg to most CATE estimators. 
	_OrthoLearner and subclasses
	Metalearners
	OrthoForest models
* Note of the estimators that accept missing values, most only accept missing in W. 
* Fix bug where DoWhyWrapper did not work with DMLOrthoForest when discrete_treatment=True


kbattocchi (changes requested):
Looks good, but please add at least one test exercising this functionality.

esbraun:
Greetings! Per the office hours call - may we let the user opt in to nan's for non parametric DML, the double robust learner, and the S, T and X learners? All of these models potentially allow for a cate model able to handle nan's.

kbattocchi (approved):
These updated changes look good to me.

PR #794, opened 2023-07-18
Pin cython<3

kbattocchi:
We're getting several compilation errors with Cython 3, which pinning it will work around until we're able to address them


fverac:
Looks good. Do we also have to change 
https://github.com/py-why/EconML/blob/main/.github/workflows/publish-package.yml#L94
and 
https://github.com/py-why/EconML/blob/main/.github/workflows/publish-documentation.yml#L66
?

fverac (approved):

kbattocchi:
> Looks good. Do we also have to change https://github.com/py-why/EconML/blob/main/.github/workflows/publish-package.yml#L94 and https://github.com/py-why/EconML/blob/main/.github/workflows/publish-documentation.yml#L66 ?

Yes, good catch

evan-magnusson:
Has there been any progress (or plans) to upgrade to Cython 3?

PR #803, opened 2023-08-04
Add LKG generation to CI

kbattocchi:
This will generate updated LKG files and store them as an artifact as part of our nightly builds.

### DIFF
--- a/.github/workflows/generate_lkg.py
+++ b/.github/workflows/generate_lkg.py
@@ -1,0 +1,217 @@
+# Copyright (c) PyWhy contributors. All rights reserved.
+# Licensed under the MIT License.
+
+import argparse
+import re
+from collections import defaultdict, namedtuple
+from os import listdir, path
+
+import packaging.version
+from packaging.version import Version
+
+# We have a list of requirements files, one per python version and OS.
+# We want to generate a single requirements file that specifies the requirements
+# for each package contained in any of those files, along with the constraints on python version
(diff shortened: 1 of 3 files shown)


.github/workflows/generate_lkg.py
```diff
@@ -0,0 +1,216 @@
...
+                #  (platform_system=='os2' and (python_version=='py3' or ...) ..."
+                #
+                # that is, we will OR together constraints grouped by os
+                # for some oses, we might find a nice representation for their python version constraints in the os_map
+                # (e.g. "python_version=='3.8'", or "python_version<'3.8'"), in which case we'll use that;
+                # for others, we'll have to OR together all of the relevant individual versions
+                os_constraints = []
+
+                os_versions = defaultdict(set)  # dictionary from os to set of python versions
+                for fp in combos:
```
fverac, line 179: What do you think about changing occurrences of `fp` into `combo`. Since fp refers to `FileParts` which is now called `Combo` 

For this line specifically it could be 
```
for combo in combos:
```

fverac: Or maybe even `c` if `combo` is too verbose. Mostly I just think `fp` is vestigial

kbattocchi: Fixed, good catch.

fverac (approved):
Looks good!

PR #804, opened 2023-08-04
Several minor improvements

kbattocchi:
* Support direct covariance fitting for DRIV
* Ensure that groups can be passed to DMLIV and DRIV
* Dependency cleanup:
    * Enable newer versions of shap, matplotlib, seaborn, and dowhy
    * Drop support for sklearn<1.0 and enable support for sklearn 1.3
* CI improvements:
    * Run doctests as part of build
    * Don't fail fast when building packages fails on one platform
    * Store test output in an artifact


econml/iv/dml/_dml.py
```diff
@@ -526,7 +526,7 @@ def score(self, Y, T, Z, X=None, W=None, sample_weight=None):
             The MSE of the final CATE model on the new data.
         """
         # Replacing score from _OrthoLearner, to enforce Z to be required and improve the docstring
-        return super().score(Y, T, X=X, W=W, Z=Z, sample_weight=sample_weight)
+        return super().score(Y, T, X=X, W=W, Z=Z, sample_weight=sample_weight, groups=None)
```
fverac, line 529: should it be groups=groups here?

econml/iv/dr/_dr.py
```diff
@@ -318,7 +318,7 @@ def predict(self, X=None):
         X = self._transform_X(X, fitting=False)
         return self._model_final.predict(X).reshape((-1,) + self.d_y + self.d_t)
 
-    def score(self, Y, T, X=None, W=None, Z=None, nuisances=None, sample_weight=None):
+    def score(self, Y, T, X=None, W=None, Z=None, nuisances=None, sample_weight=None, groups=None):
```
fverac, line 321: groups=groups?

econml/tests/test_dmliv.py
```diff
@@ -187,3 +187,54 @@ def true_fn(X):
...
+                discrete_treatment=True,
+                discrete_instrument=True,
+                model_y_xw=LinearRegression(),
+                model_t_xw=LogisticRegression(),
+                model_t_xwz=LogisticRegression(),
+            ),
+        ]
+
+        for est in est_list:
+            est.fit(y, T, Z=Z, X=X, W=W, groups=groups)
```
fverac, line 238: Is there a way to make sure the groups are actually being used here? To avoid problems like when groups is accidentally left as None in the call to super().score() instead of threaded through from the args.

econml/iv/dml/_dml.py
kbattocchi: Yes, good catch.  (It doesn't affect the results since groups are never used in scoring, but I'll fix it in the next set of changes).

econml/iv/dr/_dr.py
kbattocchi: This is the method definition, so groups=None is correct.

econml/iv/dml/_dml.py
```diff
@@ -526,7 +526,7 @@ def score(self, Y, T, Z, X=None, W=None, sample_weight=None):
             The MSE of the final CATE model on the new data.
```
fverac, line 526: minor since groups aren't really used for scoring but they are not included in the docstring as parameters

econml/tests/test_dml.py
```diff
@@ -1151,7 +1151,7 @@ def test_groups(self):
             est.fit(y, t, groups=groups)
 
         # test outer grouping
-        est = LinearDML(model_y=LinearRegression(), model_t=LinearRegression(), cv=GroupKFold(2))
```
fverac, line 1154: Is it worth adding some check to verify that a GroupKFold splitter was used under the hood?

econml/tests/test_driv.py
```diff
@@ -201,34 +209,137 @@ def dgp(n, p, true_fn):
...
-        n = 1000
-        p = 10
-        true_ate = 10
-
-        def true_fn(X):
-            return true_ate
-        y, T, Z, X = dgp(n, p, true_fn)
         for est in ests_list:
             with self.subTest(est=est):
+                # no heterogeneity
```
fverac, line 214: Minor question but is there a benefit to moving this inside the for loop?

notebooks/CustomerScenarios/Case Study - Multi-investment Attribution at A Software Company.ipynb
fverac, line 1: The diff is hard to parse here for some reason even though the actual changes are minimal just like the econml+dowhy version of the notebook. 
Not sure why. Different jupyter version?

econml/tests/test_drlearner.py
```diff
@@ -793,9 +793,12 @@ def test_groups(self):
             est.fit(y, t, W=w, groups=groups)
 
         # test outer grouping
-        # NOTE: we should ideally use a stratified split with grouping, but sklearn doesn't have one yet
+        # NOTE: StratifiedGroupKFold has a bug when shuffle is True where it doesn't always stratify properly
```
fverac, line 796: Is this bug worth worrying about for our users since we use crossfit uses StratifiedGroupKFold with shuffle=True?

econml/tests/test_driv.py
kbattocchi: The test passes :-).  The default is for fit_cov_directly to be True, which means that the previous random seed doesn't generate identical results to what they were before, which lead to a marginal failure on this test.

Logically, I think this makes more sense anyway: it's weird to have different loops creating two sets of identical subtests that test different things; if you run the tests locally via `unittest` you'll see one result per subtest but there won't be any way to tell which was which.

notebooks/CustomerScenarios/Case Study - Multi-investment Attribution at A Software Company.ipynb
kbattocchi: That was indeed very weird; fixed.

econml/tests/test_drlearner.py
kbattocchi: Hopefully it will be fixed in sklearn and then it will have the right behavior, but until then it's possible that users can run into it (although the buggy behavior only occurs with certain datasets, so hopefully it works most of the time).

However I don't think there's any good fix on our end - in general we do want to shuffle, it's just that for the purposes of this one test we can ignore that, but it wouldn't be an appropriate substitute in general.

econml/iv/dml/_dml.py
kbattocchi: Upon further consideration, I've removed `groups` from the DMLIV and DRIV scoring methods, because they are never used and so there's no point in including them.

The `groups` argument needs to exist on the nuisance models, because the signatures for fit, predict, and score all need to be compatible for how we do cross-fitting, but there's no need for them to pollute the estimators themselves, and indeed our existing classes like LinearDML do not have groups on their scoring methods.

econml/iv/dml/_dml.py
kbattocchi: As mentioned in a previous comment, removed groups from scoring on the estimator since they do nothing

econml/iv/dml/_dml.py
```diff
@@ -837,7 +837,7 @@ def fit(self, Y, T, *, Z, X=None, W=None, sample_weight=None, freq_weight=None,
                            sample_weight=sample_weight, freq_weight=freq_weight, sample_var=sample_var, groups=groups,
                            cache_values=cache_values, inference=inference)
 
-    def score(self, Y, T, Z, X=None, W=None, sample_weight=None):
+    def score(self, Y, T, Z, X=None, W=None, sample_weight=None, groups=None):
         """
         Score the fitted CATE model on a new data set. Generates nuisance parameters
         for the new data set based on the fitted residual nuisance models created at fit time.
```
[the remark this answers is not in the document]
kbattocchi: As mentioned, removed groups from this method.

econml/tests/test_dml.py
```diff
@@ -1151,7 +1151,7 @@ def test_groups(self):
...
+        est = LinearDML(model_y=LinearRegression(), model_t=LinearRegression())
         est.fit(y, t, groups=groups)
```
fverac, line 1156:
```suggestion
assert isinstance(est.splitter, GroupKFold)
```
What about adding something like this. Just to protect against the case where groups isn't actually used under the hood.

fverac: Though seems like currently we don't save the splitter to our ests as an attribute

fverac (approved):
Looks good

PR #808, opened 2023-08-18
Enable model selection for first stage models

kbattocchi:


econml/dml/dml.py
```diff
@@ -29,74 +29,85 @@
...
-            Target = inverse_onehot(Target)
 
-        if sample_weight is not None:
-            fit_with_groups(self._model, self._combine(X, W, Target.shape[0]), Target, groups=groups,
-                            sample_weight=sample_weight)
-        else:
-            fit_with_groups(self._model, self._combine(X, W, Target.shape[0]), Target, groups=groups)
-        return self
+class _FirstStageWrapper:
+    def __init__(self, model, discrete_target):
```
fverac, line 48: I like the refactoring. There is a consequence of not being able to make model-specific (Y vs T) warnings though (like the binary outcomes PR tries to do in this class).

econml/dml/dml.py
```diff
@@ -29,74 +29,85 @@
...
+                         cross_product, deprecated,
                          hstack, inverse_onehot, ndim, reshape,
                          reshape_treatmentwise_effects, shape, transpose,
                          get_feature_names_or_default, filter_none_kwargs)
 from .._shap import _shap_explain_model_cate
+from ..sklearn_extensions.model_selection import get_selector, ModelSelector, SingleModelSelector
 
 
-class _FirstStageWrapper:
-    def __init__(self, model, is_Y, featurizer, linear_first_stages, discrete_treatment):
```
fverac, line 40: Is linear_first_stages used anymore?

econml/sklearn_extensions/model_selection.py
```diff
@@ -256,6 +270,536 @@ def get_n_splits(self, X, y, groups=None):
         return self.n_splits
 
 
+class ModelSelector(metaclass=abc.ABCMeta):
+    """
+    This class enables a two-stage fitting process, where first a model is selected
+    by calling `train` with `is_selecting=True`, and then the selected model is fit (presumably
```
fverac, line 276: I am wondering why we opt for a 'train' method over a 'fit' method

econml/sklearn_extensions/model_selection.py
```diff
@@ -256,6 +270,536 @@ def get_n_splits(self, X, y, groups=None):
...
+        return self.best_model.predict_proba(*args, **kwargs)
+
+    def score(self, *args, **kwargs):
+        if hasattr(self.best_model, 'score'):
+            return self.best_model.score(*args, **kwargs)
+        else:
+            return None
+
+
+def _fit_with_groups(model, X, y, *, groups, **kwargs):
```
fverac, line 335: marking for discussion

econml/dml/dml.py
```diff
@@ -716,13 +741,19 @@ class LinearDML(StatsModelsCateEstimatorMixin, DML):
 
     def __init__(self, *,
                  model_y='auto', model_t='auto',
```
fverac, line 743: We should probably update our docstrings to say that users can pass something like 'forest' to model_y right?

econml/_ortho_learner.py
```diff
@@ -202,13 +204,17 @@ def predict(self, X, y, W=None):
 
     """
     model_list = []
+
+    kwargs = filter_none_kwargs(**kwargs)
+    model.train(True, *args, **kwargs)
```
fverac, line 209: I think in some earlier conversations we were thinking about giving the users the option to do "dirty crossfitting" i.e. picking a good est from all data before cross fitting. Am I correct in my understanding that this PR just does "dirty crossfitting" by default?

econml/dml/dml.py
fverac, line 40: Also do we no longer featurizer our first stage models?

fverac:
General comment not just about this PR but of process of adding new features - I feel like a demo notebook (even if brief) would help users get up to speed with new features we add. Maybe we can add demo notebooks before moving from a beta release to an official release?
.e.g. how does one make use of the new param_list arg?

econml/dml/dml.py
kbattocchi: Let's think about the cleanest way to do that as part of your PR; one thing to keep in mind is that we also use this wrapper for some Z models, so it's not just T vs. Y, which is what we assumed when we first implemented this logic when we only had DML but not the IV classes.

econml/dml/dml.py
kbattocchi: No; linear_first_stages has always been an awkward hack and this gives us a good opportunity to get rid of it.  Given

T = f(X, W) + error
Y = T <alpha phi(X)> + g(X,W) + error

if both f and g are linear, then because we multiply T by theta in the equation for Y, Y is not actually a linear function of X and W, we need to include interaction terms.  So we introduced linear_first_stages because our default y and t models were linear and we wanted to make sure we could estimate the correct model if f and g were linear.  Now, since we do model selection, the y and t models are not (necessarily) linear, so this makes less sense.

There's also a potentially big performance cost if users accidentally leave linear_first_stages set to True, because we can end up generating a huge number of interaction terms, and we see this happen fairly often in practice.

Additionally, linear models aren't the only time that specifying the right models for f and g as your t- and y-models wouldn't give the correct estimate - if f and g are both degree-n polynomials then you'd need to interact the degree-n interaction terms of [X;W] with the featurized X, for example, and we've never supported that; users can always supply a more complicated pipelined model that does something like that themselves if they'd like, but I'd argue it doesn't make sense to try to solve it on our end.

econml/sklearn_extensions/model_selection.py
kbattocchi: At first I used `fit`, but this makes the error messages much easier to understand if a selector is being passed where an estimator is expected or vice-versa.

econml/dml/dml.py
kbattocchi: Yes, I think we should probably create a whole new documentation section on how to specify model selectors, because there are a number of possibilities now.  But I was planning on deferring that until after this PR.

econml/_ortho_learner.py
kbattocchi: Yes, and that's definitely something we could consider making easier for users.

It's possible, though not straightforward, to do non-dirty crossfitting now, by wrapping a CV estimator in a FixedModelSelector, which will always use the estimator as is for both selecting and fitting.  However, there are some changes we could make to make this more efficient, since then the selecting step is unnecessary and so we could just skip it.

I'd propose tabling that for now and implementing that as one of several future enhancements to the model selection logic.

econml/dml/dml.py
kbattocchi: We only ever used featurization in the first stage models to support linear_first_stages, since then we needed to interact phi(X) with [X;W].

Logically, the featurization is only about the final model, the first stage models just get the raw X and W and can fit whatever kind of model they'd like on them, including one that pipelines featurization with any other logic.

fverac (approved):

PR #816, opened 2023-09-22
Improve discreteness handling, allow binary outcomes

fverac:
Adds a binary_outcome keyword arg to most estimators, where if True then the outcome nuisance model will be a classifier.


econml/dr/_drlearner.py
```diff
@@ -101,12 +103,18 @@ def predict(self, Y, T, X=None, W=None, *, sample_weight=None, groups=None):
         n = T.shape[0]
         Y_pred = np.zeros((T.shape[0], T.shape[1] + 1))
         T_counter = np.zeros(T.shape)
-        Y_pred[:, 0] = self._model_regression.predict(np.hstack([XW, T_counter])).reshape(n)
+        if self._binary_outcome and hasattr(self._model_regression, 'predict_proba'):
```
kbattocchi, line 106: Is it valid to have a regression model without a `predict_proba` method if `binary_outcome` is True?

econml/dr/_drlearner.py
```diff
@@ -101,12 +103,18 @@ def predict(self, Y, T, X=None, W=None, *, sample_weight=None, groups=None):
...
+            Y_pred[:, 0] = self._model_regression.predict_proba(np.hstack([XW, T_counter]))[:, 1].reshape(n)
+        else:
+            Y_pred[:, 0] = self._model_regression.predict(np.hstack([XW, T_counter])).reshape(n)
         Y_pred[:, 0] += (Y.reshape(n) - Y_pred[:, 0]) * np.all(T == 0, axis=1) / propensities[:, 0]
         for t in np.arange(T.shape[1]):
             T_counter = np.zeros(T.shape)
             T_counter[:, t] = 1
-            Y_pred[:, t + 1] = self._model_regression.predict(np.hstack([XW, T_counter])).reshape(n)
+            if self._binary_outcome and hasattr(self._model_regression, 'predict_proba'):
```
kbattocchi, line 114: Same question here

econml/tests/test_bootstrap.py
```diff
@@ -270,8 +270,8 @@ def test_internal_options(self):
 
     def test_stratify(self):
         """Test that we can properly stratify by treatment"""
-        T = [1, 0, 1, 2, 0, 2]
-        Y = [1, 2, 3, 4, 5, 6]
+        T = np.array([1, 0, 1, 2, 0, 2])
+        Y = np.array([1, 2, 3, 4, 5, 6])
```
kbattocchi, line 274: Is this change necessary or just for consistency with X?  If the latter that's fine, but if the former then that's not ideal since it's nice to allow lists instead of proper numpy vectors when possible for convenience.

econml/tests/test_ortho_learner.py
```diff
@@ -1,484 +1,487 @@
-# Copyright (c) PyWhy contributors. All rights reserved.
```
kbattocchi, line 1: Any idea why GitHub thinks every line has changed?

kbattocchi (changes requested):
Mostly looks great.  I've suggested a few minor changes.

Related to this PR, I think lines 52-57 of econml/dml/dml.py should be made more robust - what happens if the target is discrete and there is no predict_proba method, or target is not discrete and there is a predict_proba method?  I think in either case, we should at least warn the user that they are passing a classifier where a regressor is expected or vice versa.  When there is no predict_proba method but one is expected, there doesn't seem to be much harm in falling back to calling predict instead as long as the user is warned; in the opposite scenario it's less clear to me that calling predict_proba instead of just calling predict as usual is a good idea, but at least if we warn the user they can change the discreteness and get that behavior if they want.

Whatever we decide here, if there's a non-trivial amount of logic we do something similar with other estimators that don't use this first stage wrapper; probably that should happen in a new utility method that this module and others can all use consistently (perhaps something like `get_prediction(estimator, expected_discrete)`).

econml/tests/test_ortho_learner.py
fverac: Seems that the line endings got switched to crlf from lf at some point. Will try and switch back

econml/_ortho_learner.py
```diff
@@ -516,11 +519,20 @@ def _gen_ortho_learner_model_final(self):
     """
 
     def __init__(self, *,
-                 discrete_treatment, treatment_featurizer,
-                 discrete_instrument, categories, cv, random_state,
-                 mc_iters=None, mc_agg='mean', allow_missing=False, use_ray=False, ray_remote_func_options=None):
-        self.actors = []
```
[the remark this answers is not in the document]
fverac: Yes. Seemed like it got added from the Ray PR, and never used. And removing didn't affect any tests.

econml/_ortho_learner.py
```diff
@@ -327,6 +327,9 @@ class _OrthoLearner(TreatmentExpansionMixin, LinearCateEstimator):
 
     Parameters
     ----------
+    binary_outcome: bool
+        Whether the outcome should be treated as binary
+
```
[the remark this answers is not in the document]
fverac: I don't think we've written it off completely, but definitely out of scope for this PR. I'll change it to discrete_outcome.

econml/dr/_drlearner.py
fverac: I was thinking it's not ideal but it should be fine (since this is the way it's been done all the way until now so maybe we should leave the option to use a regressor when y is binary). I will add a warning to communicate this is not the intended type of model to pass when discrete_outcome=True.

fverac:
> Mostly looks great. I've suggested a few minor changes.
> 
> Related to this PR, I think lines 52-57 of econml/dml/dml.py should be made more robust - what happens if the target is discrete and there is no predict_proba method, or target is not discrete and there is a predict_proba method? I think in either case, we should at least warn the user that they are passing a classifier where a regressor is expected or vice versa. When there is no predict_proba method but one is expected, there doesn't seem to be much harm in falling back to calling predict instead as long as the user is warned; in the opposite scenario it's less clear to me that calling predict_proba instead of just calling predict as usual is a good idea, but at least if we warn the user they can change the discreteness and get that behavior if they want.
> 
> Whatever we decide here, if there's a non-trivial amount of logic we do something similar with other estimators that don't use this first stage wrapper; probably that should happen in a new utility method that this module and others can all use consistently (perhaps something like `get_prediction(estimator, expected_discrete)`).

Adding a warning when first stage is discrete target but model does not have predict_proba,
and raising an error when first stage target is continuous but model does have predict_proba.

econml/_ortho_learner.py
```diff
@@ -739,15 +713,17 @@ def fit(self, Y, T, *, X=None, W=None, Z=None, sample_weight=None, freq_weight=N
...
                 self.outcome_transformer = LabelEncoder()
                 self.outcome_transformer.fit(Y)
                 if Y.shape[1:] and Y.shape[1] > 1:
                     raise ValueError(
-                        f"Only one outcome variable is supported when binary_outcome=True. Got Y of shape {Y.shape}")
+                        f"Only one outcome variable is supported when discrete_outcome=True. Got Y of shape {Y.shape}")
                 if len(self.outcome_transformer.classes_) > 2:
                     raise AttributeError(
-                        "More than 2 outcome classes detected. This method currently only supports binary outcomes")
+                        f"({self.outcome_transformer.classes_} outcome classes detected. \
```
kbattocchi, line 724: I think you're including the classes themselves rather than their count here.

econml/utilities.py
```diff
@@ -1482,3 +1482,29 @@ def jacify_featurizer(featurizer):
        a function for calculating the jacobian
     """
     return _TransformerWrapper(featurizer)
+
+
+def single_strata_from_discrete_arrays(arrs):
```
kbattocchi, line 1487: [minor] The singular for strata is stratum, so this name seems slightly weird.  I think `strata_from_discrete_arrays` (since this gets the strata for all of the rows at once) is shorter and just as clear.

kbattocchi (approved):
This looks good; I've made a couple of minor suggestion but you can merge as soon as you are comfortable with it.

econml/_ortho_learner.py
fverac: Good catch

PR #818, opened 2023-10-04
Kgao/federal learning

kgao:
Draft PR, please review the documents first.


econml/sklearn_extensions/federated_learning.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) PyWhy contributors. All rights reserved.
+# Licensed under the MIT License.
+
+import numpy as np
+from econml.dml import LinearDML
+from econml.sklearn_extensions.linear_model import StatsModelsLinearRegression
+from typing import List
+
+
+class FederatedLearner:
```
kbattocchi, line 10: It would probably be better to have this class implement the whole LinearCateEstimator API using the aggregated final model.  However, we could defer that for now if you'd like.

But in any case, there should at least be a docstring that describes the basics of the class, like what its purpose is and what attributes are present.

econml/sklearn_extensions/federated_learning.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) PyWhy contributors. All rights reserved.
```
kbattocchi, line 1: I don't think `sklearn_extensions` is the best place to put this; I think `econml.federated_learning.FederatedEstimator` is probably a better location and name - so move this file up one directory and add `federated_learning` to the list in `econml\__init__.py`.

econml/sklearn_extensions/local_test_combined.py
```diff
@@ -0,0 +1,76 @@
+import numpy as np
```
kbattocchi, line 1: Remove this file

econml/sklearn_extensions/local_test_splitted.py
```diff
@@ -0,0 +1,62 @@
+import numpy as np
```
kbattocchi, line 1: Remove this file

econml/tests/test_federated_learning.py
```diff
@@ -0,0 +1,89 @@
...
+
+import numpy as np
+import unittest
+from econml.dml import LinearDML
+from econml.inference import StatsModelsInference
+from econml.sklearn_extensions.federated_learning import FederatedLearner
+
+
+class FunctionRegressor:
+    def __init__(self, func):
```
kbattocchi, line 12: Add a docstring just to make it even more clear what's happening here.

```suggestion
    """A simple model that ignores the data it is fitted on, always just using the specified function to predict"""
    def __init__(self, func):
```

kbattocchi (changes requested):
Mostly looks good, I've suggested a few minor changes to the code.  You also need to fix the linting and documentation building CI failures.  I didn't review the .rst file yet, since I'll wait until the docs are generated and look at the resulting HTML

econml/sklearn_extensions/federated_learning.py
kgao: Agreed, we could discuss how we would like to do as implementation of the whole LinearCateEstimator API; For now, I just add the docstring for the class.

econml/sklearn_extensions/federated_learning.py
kgao: Done.

econml/sklearn_extensions/local_test_combined.py
kgao: Done.

econml/sklearn_extensions/local_test_splitted.py
kgao: Done.

econml/tests/test_federated_learning.py
kgao: Agreed, updated.

econml/sklearn_extensions/federated_learning.py
kgao: Related linting should be fixed too.

doc/spec/federated_learning.rst
```diff
@@ -0,0 +1,353 @@
+Federated Learning in Microsoft EconML Library
```
kbattocchi, line 1: This documentation has some helpful content but also seems very long, with lots of boilerplate/fluff, most of which is not related to our library.

Please trim it down a lot - we don't need an introduction on how to use Spark in general, for example.

Most importantly, this documentation does not include any reference to the actual classes in our library that implement the functionality.  The "Combining Final Results" section looks completely generic and refers to a whole bunch of techniques that have nothing to do with our library.

kbattocchi (changes requested):
Please clean up the documentation, removing all of the generic content and mainly focusing on the actual classes that perform the work in our library.  Explaining that this can be used with Spark is helpful, but should not go into a full intro to Spark (and one benefit of how we've implemented this functionality is that it could also be used with any other distributed computing platform).

Also, please merge the latest changes from main into your branch.

kbattocchi (approved):
Now looks good, will merge once all checks pass.

Issue #827, opened 2023-11-08
Issue with #791: Missing data allowed when training, but not inferencing


esbraun:
#791 ought to allow missing data when inferencing if allow_missing = True, but currently an error is thrown when the inference data includes missing data. RE: below for a working example based on the "Metalearner Examples" notebook for a demonstration of the issue.

https://colab.research.google.com/drive/1QT-IdwRWRUiQNDwqJ0EXPXSftDuIg3uE?usp=sharing

Ftrejo23:
I'm seeing the same issue

fverac:
Thanks for raising this issue. In the original PR we originally only considered missing data during training, but it seems there is interest in extending this to inferencing as well.

erasedcitizen11:
Hi there, just wondering if there any update on this as running into the same problems (doesn't allow missing values when computing CATE/ATE).

esbraun:
@erasedcitizen11 I created a fork that disables the missing value checks which is a requirement for our use cases. Use at your own risk: https://github.com/esbraun/EconML/tree/main

pgeertsema:
@EconML team. Great product!

Any progress in remediating this issue? It appears that scikit-learn's default input validation checks remain in place even when ```allow_missing=True``` is set in DML() and the base learners support missing values (such as LightGBM). This is inconsistent with the documentation (at https://econml.azurewebsites.net/_autosummary/econml.dml.DML.html, see allow_missing (bool)). A self-contained minimal working example follows: 

```
import numpy as np
import scipy.special as sc
from econml.dml import DML
from econml.sklearn_extensions.linear_model import StatsModelsLinearRegression
from lightgbm import LGBMRegressor, LGBMClassifier

np.random.seed(123)
X = np.random.normal(size=(1000, 5))
T = np.random.binomial(1, sc.expit(X[:, 0]))

mask = np.random.rand(*X.shape) < 0.05
X[mask] = np.nan

y = (1 + .5*X[:, 0]) * T + X[:, 0] + np.random.normal(size=(1000,))
est = DML(
    # both base learner models support missing values out of the box
    model_y=LGBMRegressor(),
    model_t=LGBMClassifier(),
    model_final=StatsModelsLinearRegression(fit_intercept=False),
    # should allow discrete treatment
    discrete_treatment=True,
    allow_missing=True,
)
est.fit(y, T, X=X, W=None)

```

It terminates with

```
Traceback (most recent call last):
[...]

PR #828, opened 2023-11-15
Meta Learner allow N/A  

xhulianoThe1:
Fixing and adding to sklearn utils function that checks array the ability to allow NA values when do effect estimation. 
This based on the issue noted here: https://github.com/py-why/EconML/issues/827

### DIFF
--- a/econml/metalearners/_metalearners.py
+++ b/econml/metalearners/_metalearners.py
@@ -134,7 +134,7 @@ class TLearner(TreatmentExpansionMixin, LinearCateEstimator):
             the corresponding singleton dimensions in the output will be collapsed
         """
         # Check inputs
-        X = check_array(X)
+        X = check_array(X, force_all_finite='allow-nan' if 'X' in self._gen_allowed_missing_vars() else True)
         taus = []
         for ind in range(self._d_t[0]):
             taus.append(self.models[ind + 1].predict(X) - self.models[0].predict(X))
@@ -259,7 +259,7 @@ class SLearner(TreatmentExpansionMixin, LinearCateEstimator):
         # Check inputs
         if X is None:
             X = np.zeros((1, 1))
-        X = check_array(X)
+        X = check_array(X, force_all_finite='allow-nan' if 'X' in self._gen_allowed_missing_vars() else True)
         Xs, Ts = broadcast_unit_treatments(X, self._d_t[0] + 1)
         feat_arr = np.concatenate((Xs, Ts), axis=1)
         prediction = self.overall_model.predict(feat_arr).reshape((-1, self._d_t[0] + 1,) + self._d_y)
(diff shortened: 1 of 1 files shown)


fverac:
Thanks for the PR! For completeness, would you also be able to make changes that would allow missing values during effect estimation for other ests that support missing values in X, like DRLearner, DML, and NonParamDML as well?

xhulianoThe1:
Yes! Will update PR.

fverac:
Hey, I never followed up on this. 

Seems your latest commit broke some tests. I haven't quite looked at the errors in the tests, but speaking to the commit itself, I would generally recommend against adding the additional check_inputs lines you've added, because these checks already exist somewhere in the control flow (you can look at the errors/traceback generated in the colab [notebook linked](https://colab.research.google.com/drive/1QT-IdwRWRUiQNDwqJ0EXPXSftDuIg3uE?usp=sharing#scrollTo=sL1o0ZqYBwCE) in #827). 

So in order to allow missing values during effect estimation for DRLearner, DML, and NonParamDML, we'd have to address the already-existing missing value checks in the control flow (via expand_treatments, for instance). 

Another thing I'm noticing is that while your metalearner-specific fixes may have allowed for calling const_marginal_effect with missing values in X, I still don't think it would work when calling effect (or marginal effect, const_marginal_ate, etc.) with missing values in X (because of the checks that expand_treatments does, as can be referred in the colab notebook linked)

So it seems the next step is to tackle how we might allow expand_treatments to allow missing values in X.

Finally, another thing we'd want to do before we merge is create corresponding tests to for the functionality we've added (via adding to econml/tests/test_missing_values.py). So tests that actually double check that we can call XLearner(...).fit(..).effect(X) when X has missing values.

I know I just shared a lot of comments, and I don't know how much bandwidth/interest you have in bringing this PR to completion, so feel free to let me know if you'd rather the EconML team try to wrap the PR up by picking up where you left off (though I can't say when we'd get to it).

xhulianoThe1:
> Hey, I never followed up on this.
> 
> Seems your latest commit broke some tests. I haven't quite looked at the errors in the tests, but speaking to the commit itself, I would generally recommend against adding the additional check_inputs lines you've added, because these checks already exist somewhere in the control flow (you can look at the errors/traceback generated in the colab [notebook linked](https://colab.research.google.com/drive/1QT-IdwRWRUiQNDwqJ0EXPXSftDuIg3uE?usp=sharing#scrollTo=sL1o0ZqYBwCE) in #827).
> 
> So in order to allow missing values during effect estimation for DRLearner, DML, and NonParamDML, we'd have to address the already-existing missing value checks in the control flow (via expand_treatments, for instance).
> 
> Another thing I'm noticing is that while your metalearner-specific fixes may have allowed for calling const_marginal_effect with missing values in X, I still don't think it would work when calling effect (or marginal effect, const_marginal_ate, etc.) with missing values in X (because of the checks that expand_treatments does, as can be referred in the colab notebook linked)
> 
> So it seems the next step is to tackle how we might allow expand_treatments to allow missing values in X.
> 
> Finally, another thing we'd want to do before we merge is create corresponding tests to for the functionality we've added (via adding to econml/tests/test_missing_values.py). So tests that actually double check that we can call XLearner(...).fit(..).effect(X) when X has missing values.
> 
> I know I just shared a lot of comments, and I don't know how much bandwidth/interest you have in bringing this PR to completion, so feel free to let me know if you'd rather the EconML team try to wrap the PR up by picking up where you left off (though I can't say when we'd get to it).

Hey, sorry I have had very limited bandwidth and need to deep dive in the intricacies of the code base some more. I will find try and find time to plug the holes based on what you commented and reupdate the draft as I go along. Then if I can't finish in a timely fashion whenever one of the members has bandwidth they can take it on.

PR #846, opened 2024-01-31
Make federation opt in, plus misc. fixes

kbattocchi:
This PR does a few things:
* Improve federation by making it opt in, and improve documentation
* Fix a pytest build issue, plus upgrade some GitHub Actions components that are generating warnings
* Remove the code that causes some plots not to show up in our notebooks
* Remove deprecated `linear_first_stages` arguments from docs
* Remove deprecated location for DynamicDML


econml/tests/test_federated_learning.py
```diff
@@ -74,9 +74,12 @@ def test_lineardrlearner(self):
 
             for cov_type in ['HC0', 'HC1', 'nonrobust']:
                 with self.subTest(n_t=n_t, cov_type=cov_type):
-                    est_all = LinearDRLearner(model_propensity=t_model, model_regression=y_model)
-                    est_h1 = LinearDRLearner(model_propensity=t_model, model_regression=y_model)
-                    est_h2 = LinearDRLearner(model_propensity=t_model, model_regression=y_model)
+                    est_all = LinearDRLearner(model_propensity=t_model, model_regression=y_model,
```
fverac, line 77: Should we also test that aggregation is not possible when enable_federation=False?

econml/tests/test_federated_learning.py
```diff
@@ -150,9 +153,12 @@ def test_splitting_works(self):
                                 with self.subTest(d_x=d_x, d_t=d_t, d_y=d_y,
                                                   weights=(weights is not None), fw=(freq_weights is not None),
                                                   cov_type=cov_type):
-                                    est_all = LinearDML(model_t=t_model, model_y=y_model, linear_first_stages=False)
-                                    est_h1 = LinearDML(model_t=t_model, model_y=y_model, linear_first_stages=False)
-                                    est_h2 = LinearDML(model_t=t_model, model_y=y_model, linear_first_stages=False)
+                                    est_all = LinearDML(model_t=t_model, model_y=y_model, linear_first_stages=False,
+                                                        enable_federation=True)
+                                    est_h1 = LinearDML(model_t=t_model, model_y=y_model, linear_first_stages=False,
```
fverac, line 158: Is there any reason to keep linear_first_stages=False around? Seems like we could start removing its presence from our tests (it is present in other tests as well e.g. test_refit)

notebooks/Choosing First Stage Models.ipynb
fverac, line 1: Seems some of the markdown for this file is outdated now with our new model selection capabilities. Probably worth updating

e.g. outdated description: 
"Disadvantages:

The EconML estimator will take longer to run due to an internal cross-validation step for computing the residuals. Further, the CV estimator will be trained on 
 data points which might not be suitable for small datasets.
Requires special CV estimator to choose among many classes of estimators (e.g. Lasso and GradientBoostingForest, see section 2.2. for workaround)"

econml/tests/test_federated_learning.py
kbattocchi: Good suggestion, will address in next revision.

econml/tests/test_federated_learning.py
kbattocchi: My thinking was that it's deprecated but not removed I'd leave it in place, but since it's completely ignored even when specified I guess there's no benefit to this, so I'll remove it from tests as well (not just user-facing documentation).

notebooks/Choosing First Stage Models.ipynb
kbattocchi: Good point but I think that'll fit better with the next PR.

fverac:
Looks good. I would additionally remove ` dml.linear_first_stages = True` in test_refit.py as well. Not a big deal though. I'll approve and let you decide if it's worth making another commit on this PR

fverac (approved):

PR #848, opened 2024-02-02
Improve model selection

kbattocchi:
Fixes two-stage model selection used by IV models, so that we first select and train nuisances for y, t, etc., and then select and train the covariance model.  Added central documentation for how to specify how to select models and simplifies the documentation for each estimator by pointing there.


econml/sklearn_extensions/model_selection.py
```diff
@@ -599,9 +600,20 @@ def get_selector(input, is_discrete, *, random_state=None, cv=None, wrapper=Grid
     named_models = {
         'linear': (LogisticRegressionCV(random_state=random_state, cv=cv) if is_discrete
                    else WeightedLassoCVWrapper(random_state=random_state, cv=cv)),
+        'poly': ([make_pipeline([PolynomialFeatures(d),
```
fverac, line 603: Seems there could potentially be some weird interactions if a user passes 'poly' and also has something like PolynomialFeatures as an X featurizer. 

Not sure if we should do anything to protect or warn against this though

fverac:
Should we add a test that sklearn pipelines work correctly?

econml/sklearn_extensions/model_selection.py
kbattocchi: I don't think this is a problem: featurization affects the final model but not the first stage models, whereas this is for selecting a good first stage model.

kbattocchi:
> Should we add a test that sklearn pipelines work correctly?

Good point, I'll add several tests covering this functionality.

econml/_ortho_learner.py
```diff
@@ -1035,26 +1083,45 @@ def score(self, Y, T, X=None, W=None, Z=None, sample_weight=None, groups=None):
...
-        n_splits = len(self._models_nuisance[0])
+        # self._models_nuisance will be a list of lists or a list of list of lists
+        # so we use self._ortho_learner_model_nuisance to determine the nesting level
+        expand_list = not isinstance(self._ortho_learner_model_nuisance, list)
+        if expand_list:
+            n_selectors = 1
+            n_splits = len(self._models_nuisance[0])
+        else:
+            n_selectors = len(self._models_nuisance[0])
+            n_splits = len(self._models_nuisance[0][1])
```
fverac, line 1094: I'm having trouble understanding the 1 index here. Is it arbitrary? i.e. could the 1 index be replaced with a 0 index and also work fine?

kbattocchi: Yes, I'll change to 0 - there should be the same number of folds for each nuisance model so it's arbitrary, and in pratice DRIV is the only set of estimators where we have more than one nuisance model so 1 works okay, but 0 is a more natural choice.

econml/sklearn_extensions/model_selection.py
```diff
@@ -386,19 +389,31 @@ def _fit_with_groups(model, X, y, *, sub_model=None, groups, **kwargs):
...
+    def train(self, is_selecting, folds: Optional[List], X, y, groups=None, **kwargs):
+        if is_selecting:
+            # since needs_fit is False, is_selecting will only be true if
+            # the score needs to be compared to another model's
+            # so we don't need to fit the model itself, just get the out-of-sample score
+            assert hasattr(self.model, 'score'), (f"Can't select between a fixed {type(self.model)} model and others "
+                                                  "because it doesn't have a score method")
+            scores = []
+            for train, test in folds:
+                # use _fit_with_groups instead of just fit to handle nested grouping
```
fverac, line 407: I'm a little confused here. Why do we need to do nested grouping for a FixedModelSelector?

I'm having trouble grasping why we seem to be doing two levels of splitting here. (one level within this train function, and another level within _fit_with_groups).

Couldn't we get away with only one level of splitting?

kbattocchi: In practice `_fit_with_groups` almost always just calls plain fit on all of the data without doing any further splitting; the exception is if there are groups and the model that we've wrapped has a `cv` attribute, in which case the model is going to be doing its own cross-validation when we call fit no matter what, so we just want to make sure that we update the `cv` attribute to respect the groups.

In the typical case, if a user passes a known sklearn model that supports cross validation, we won't wrap it in a FixedModelSelector anyway, but this gives users an escape hatch if they want to be doing cross-validation within each fold for some reason.

econml/tests/test_model_selection.py
```diff
@@ -0,0 +1,135 @@
...
+
+            model_t = 'some_random_string'
+            with self.subTest(model_t=model_t, discrete_treatment=True):
+                mdl = LinearDML(model_t=model_t,
+                                discrete_treatment=discrete_treatment,
+                                model_y=LinearRegression())
+                with self.assertRaises(ValueError):
+                    mdl.fit(Y, T, X=X, W=W)
+
+    def test_list_selection(self):
```
fverac, line 92: Is it worth checking other possible permutations like [str, str], [str, SklearnCVselector], [fixedSelector, str].

kbattocchi: Wouldn't hurt, but I believe the current tests already cover all of the relevant code paths.

econml/sklearn_extensions/model_selection.py
```diff
@@ -578,18 +604,19 @@ def __init__(self, models, unwrap=True):
...
-                model.train(is_selecting, *args, **kwargs)
+                model.train(is_selecting, folds, *args, **kwargs)
                 scores.append(model.best_score)
             self._all_scores = scores
             self._best_score = np.max(scores)
             self._best_model = self.models[np.argmax(scores)]
 
         else:
-            self._best_model.train(is_selecting, *args, **kwargs)
+            self._best_model.train(is_selecting, folds, *args, **kwargs)
```
fverac, line 616: `is_selecting` will be False here. Do we need to pass `folds` in this case?

kbattocchi: It's academic; if `is_selecting` is False then `folds` will be None anyway so we could pass `folds` or we could equivalently pass `None`.

But I think this is slightly more robust to changes in the upstream logic, since if for some reason we ever change that invariant then the logic here might not need to be changed.

fverac (approved):
Looks good!

PR #861, opened 2024-03-11
Build updates

kbattocchi:
This PR makes several improvements to our build process and package metadata:
* Switches to a new library `uv` for installing dependencies on our test machines, which saves ~1-2 minutes on each machine
* Adds Python 3.12 to our test matrix
* Enables compatibility with dowhy 3.12
* Makes it so that the [all] and [tf] extras only apply restrictions to the versions of dependencies that we use when necessary (i.e. when we're using Python 3.8), which would otherwise cause incompatibility with the versions supported by Python 3.12
* Switches away from using `setup.py pytest` to run tests, which has been deprecated for a long time, in favor of calling `pytest` directly
* Updates the LKG files to include more recent versions of dependencies
* Migrates to updated versions of the GitHub Actions artifact API, which eliminates the warnings we were getting
* Migrates more of our package's configuration from setup.cfg to pyproject.toml


fverac (approved):
looks good!

PR #867, opened 2024-03-29
Fix the test_accuracy function by modifying the assertion logic

kgao:
- The original code is testing the accuracy of different estimators by checking if the true Average Treatment Effect (ATE) falls within the calculated confidence interval. However, the check is done only once, using a single-point estimate (ate), which may not be sufficient to validate the estimator's performance. So it failed when the proportion of a true ATE within the confidence interval is NOT greater than 0.5 (50%).

			
- The new logic: To check that 50% of the values are in the 90% confidence interval (which makes sense), but it's testing this with the ate, which returns a single value, so actually the threshold isn't important, it's a single point that is either in the interval or not.  Instead, what we should do is generate W, D, and Y several times and check that most of the time the ate is in the bounds (like, generate 10 sets of W, D, Y and check that at least 8 of those times the true ate was inside the interval.

- Also applied the logic for test_accuracy_iv (And reduced the sample size n=1000 to improve the test time)

### DIFF
--- a/econml/tests/test_discrete_outcome.py
+++ b/econml/tests/test_discrete_outcome.py
@@ -52,6 +52,7 @@ class TestDiscreteOutcome(unittest.TestCase):
                 else:
                     est.fit(Y, D, W=W)
                     ate_lb, ate_ub = est.ate_interval()
+
                 if isinstance(est, LinearDRLearner):
                     est.summary(T=1)
                 else:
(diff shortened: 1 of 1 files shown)


fverac:
Nice fix! 

Wonder if it makes sense to apply the same logic to the test_accuracy_iv method in test_discrete_outcome.py, since the logic there is also doing this weird proportion_in_interval stuff that doesn't really make much sense with ate (but might make more sense the new way you are doing it)

kbattocchi (changes requested):
This is a good start, but the logic is not quite right.

In addition, please also address @fverac's comment about making parallel changes to the corresponding IV test.

kbattocchi (approved):

Issue #866, opened 2024-03-27
`shap_values` for tree-based models doesn't set `check_additivity=False` as expected


jcreinhold:
For, e.g., `CausalForestDML`, the desired behavior (#458) of EconML is to set `check_additivity=False` when computing `shap_values` (otherwise, the computation will frequently fail for reasons unbeknownst to me).

This kwarg is set [here](https://github.com/py-why/EconML/blob/6219695cd1a6a0ff492a22a5585c15537d5d41a6/econml/_shap.py#L164) where, if the `Explainer` class is `Tree`, `check_additivity` is set to `False`.

[This works with `shap<=0.42.1`](https://github.com/shap/shap/blob/36556dd4ad3caac13c7967d9506da6a34ef6bacf/shap/explainers/_tree.py#L49); however, [in `shap>=0.43.0`](https://github.com/shap/shap/blob/4316c41e8c2086beb4128cb414e648f13e1bccef/shap/explainers/_tree.py#L49) the name of the class was changed to `TreeExplainer`.

Given that EconML states it's compatible with [`shap >= 0.38.1, < 0.44.0`](https://github.com/py-why/EconML/blob/6219695cd1a6a0ff492a22a5585c15537d5d41a6/pyproject.toml#L30), this represents a bug.

Some potential solutions:

1. Change the maximum version of `shap` to 0.44.0 to 0.42.1. (This doesn't _seem_ desirable.)
2. Change the code to:

```python
# Simple, but potentially not future-proof.
if explainer.__class__.__name__ in {"Tree", "TreeExplainer"}:
    shap_out = explainer(F, check_additivity=False)
```

3. Change to code to:

```python
# This might have false positives but might be more future proof.
if "Tree" in explainer.__class__.__name__:
    shap_out = explainer(F, check_additivity=False)
```

4. Change the code to:

```python
import inspect

# This seems like the most robust option and will always be correct
if "check_additivity" in inspect.signature(explainer).parameters:
    shap_out = explainer(F, check_additivity=False)
```

(It seems that this parameter is only used in the `TreeExplainer` and the [`DeepExplainer`](https://github.com/shap/shap/blob/74db96c333c41cee0bb90044f25d8b2dc102cdb6/shap/explainers/_deep/__init__.py#L115), so the signature inspection would probably be fine, and we probably want to set this kwarg to `False` all the time regardless (at least in this package)).

Happy to submit a PR with whatever approach though.

kbattocchi:
[...]

PR #872, opened 2024-04-04
Fix `shap_values` compatibility with shap>=0.43.0 by adjusting `check_additivity` parameter handling for `TreeExplainer`

jcreinhold:
This addresses the issue where the `shap_values` method in EconML did not correctly set `check_additivity=False` for shap>=0.43.0 due to a class name change from `Tree` to `TreeExplainer​`.

See #866 for details.

### DIFF
--- a/econml/_shap.py
+++ b/econml/_shap.py
@@ -12,6 +12,7 @@ Scott Lundberg, Su-In Lee (2017)
 
 """
 
+import inspect
 import shap
 from collections import defaultdict
 import numpy as np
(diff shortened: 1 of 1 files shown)


jcreinhold:
Adding @kbattocchi for review.

kbattocchi (approved):
Thanks for the contribution!  As I mentioned in the linked issue, if you also want to tackle passing a seed to SHAP, that would be even better, but we'll take the contribution as-is if not.

jcreinhold:
Great! I'm happy to add the seed functionality, but I think it's be best to add it to another PR. Ideally, I'll be able to add it within a week or so, but I might get pulled into something which will push back that timeline.

PR #885, opened 2024-05-10
Avoid double impurity calculation inside criterion.pyx in tree module

gdaiha:
Inspired by https://github.com/scikit-learn/scikit-learn/pull/18203/files# in Sklearn`s RandomForest, this PR aims to avoid double calculation of impurity inside criterion.pyx of tree module


gdaiha:
Can't understand exactly why these tests are failing. The only thing I've done here was to move impurity_improvement after the calculation of best.impurity_left and best.impurity_right and changed impurity_improvement signature. Tests are claiming for errors in shape of outputs of grf and cf tests

PR #891, opened 2024-06-27
Minor fixes

kbattocchi:


econml/tests/test_policy_forest.py
```diff
@@ -321,8 +321,8 @@ def test_non_standard_input(self,):
                                 model_propensity=DummyClassifier(strategy='uniform'),
                                 featurizer=PolynomialFeatures(degree=1, include_bias=False),
                                 cv=GroupKFold(n_splits=2),
-                                n_estimators=20, n_jobs=1, random_state=123).fit(y, T, X=X,
-                                                                                 groups=groups)
+                                n_estimators=100, n_jobs=1, random_state=123).fit(y, T, X=X,
```
fverac, line 324: Is this for test stability?

kbattocchi: Because `n_estimators` was never passed correctly to DRPolicyForest in the past, this has always effectively run with `n_estimators=100` (the default value), and the tests don't pass if we actually pass 20 through instead.

imatiach-msft:
this looks great, thank you - when can a new version be released to pypi so we can depend on it?

imatiach-msft:
just a gentle ping on the ETA for this fix, as we need it to resolve a security vulnerability in scikit-learn

kbattocchi:
> just a gentle ping on the ETA for this fix, as we need it to resolve a security vulnerability in scikit-learn

We hope to put out a bugfix release containing these fixes by the end of this week.

fverac (approved):
Looks good!

PR #902, opened 2024-07-17
Enable pre-commit hooks

kbattocchi:


kbattocchi:
Added these changes to the ruff PR instead.

Issue #905, opened 2024-08-05
Bug when passing featurizer that requires more than 1 row


itamarfaran:
in file econml/utilities.py:75 (function name `check_high_dimensional`):

`d_x = clone(featurizer, safe=False).fit_transform(X[[0], :]).shape[1]`

when passing a transformer such as `SplineTransformer`, it raises an error:

`ValueError: Found array with 1 sample(s) (shape=(1, 1)) while a minimum of 2 is required by SplineTransformer.`

from what I understand this line only runs to infer the number of rows out, but it fails the whole fit process.
this would also fail for transformers as `OneHotEncoder` as the number of new rows is dependent on the data

itamarfaran:
https://github.com/py-why/EconML/pull/906

PR #906, opened 2024-08-05
fix bug in check_high_dimensional

itamarfaran:


itamarfaran:
https://github.com/py-why/EconML/issues/905

kbattocchi (changes requested):
Thanks for the contribution! Please add a test that would fail without your fix.

itamarfaran:
added :)

kbattocchi (approved):

PR #911, opened 2024-08-19
add compiled .c files to .gitignore

vojavocni:
Previously, .c files were generated whenever we install the environment. 
image

I added them to .gitignore so we don't make a mistake merging them by accident if they should be ignored.


kbattocchi (approved):
Looks good, thanks for the contribution!

Issue #837, opened 2024-01-02
NonParamDMLIV sometimes ZeroDivisionError


fverac:
Seems that one can run into ZeroDivisionErrors when using NonParamDMLIV. Reproduction code below. Note, the error is not consistent. You may have to run it multiple times before getting the ZeroDivisionError. 

From briefly looking into it, it seems the T residuals are all 0's when the error occurs.

econml version 0.15.0b1. Haven't tried other versions.

Let me know if I'm missing something!

```
from econml.iv.dml import NonParamDMLIV
import numpy as np
from sklearn.linear_model import LinearRegression

n = 100
d_x = 3

Y = np.random.normal(size=(n,))
T = np.random.normal(size=(n,))
X = np.random.normal(size=(n, d_x))
Z = np.random.normal(size=(n,))

est = NonParamDMLIV(discrete_instrument=False, discrete_treatment=False, model_final=LinearRegression())

est.fit(Y, T, Z=Z, X=X)
```

kbattocchi:
This is an interesting failure mode - if the estimates for E[T|Z,X,W] are always identical to E[T|X,W] then since the final model weights the rows by the estimated variance (E[T|Z,X,W]-E[T|X,W])^2, all the weights are zero which leads to this problem.

In this particular case, Lasso is regularizing all weights to 0 so the estimators always (correctly) predict E[T] = 0 regardless of whether we condition on Z,X,W or just on X,W.

Hopefully with real world data this is less likely to occur, but we could at least throw a more meaningful error message if we do run into this scenario. But I think it is a real error condition in that we depend on the instrument affecting treatment for identification, so I don't think ignoring it and producing an estimate (say, by using all 1s for the weights if they turn out to all be 0s) would be appropriate.

ronikobrosly:
Hi @fverac , I'm looking into implementing this error message. I can reproduce this error with the following random seed:

```
from econml.iv.dml import NonParamDMLIV
import numpy as np
from sklearn.linear_model import LinearRegression

np.random.seed(784)

n = 100
d_x = 3

Y = np.random.normal(size=(n,))
T = np.random.normal(size=(n,))
X = np.random.normal(size=(n, d_x))
Z = np.random.normal(size=(n,))

est = NonParamDMLIV(discrete_instrument=False, discrete_treatment=False, model_final=LinearRegression())

est.fit(Y, T, Z=Z, X=X)
```

PR #919, opened 2024-09-08
added warning message and test for issue #837

ronikobrosly:
Hi @kbattocchi , could I get your 👀 on this?

### DIFF
--- a/econml/_ortho_learner.py
+++ b/econml/_ortho_learner.py
@@ -834,6 +834,15 @@ class _OrthoLearner(TreatmentExpansionMixin, LinearCateEstimator):
                 elif not np.array_equal(fitted_inds, new_inds):
                     raise AttributeError("Different indices were fit by different folds, so they cannot be aggregated")
 
+            if nuisances[1].sum() == 0:
+                raise ValueError(
+                    """
+                    In fitting nuisances, the estimates for E[T|Z,X,W] are identical to E[T|X,W],
+                    resulting in a situation where the rows will all be weighted to zero. Please
+                    examine your instrument variable `Z`. 
+                    """
+                )
+
             if self.mc_iters is not None:
                 if self.mc_agg == 'mean':
                     nuisances = tuple(np.mean(nuisance_mc_variants, axis=0)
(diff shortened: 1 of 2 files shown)


kbattocchi (changes requested):
Thanks for the contribution! The fix actually needs to be made elsewhere - see my comment.

ronikobrosly:
Does this look better @kbattocchi ?

econml/iv/dml/_dml.py
```diff
@@ -761,6 +761,14 @@ def predict(self, Y, T, X=None, W=None, Z=None, sample_weight=None, groups=None)
             TX_pred = np.tile(TX_pred.reshape(1, -1), (T.shape[0], 1))
         Y_res = Y - Y_pred.reshape(Y.shape)
         T_res = TXZ_pred.reshape(T.shape) - TX_pred.reshape(T.shape)
+        if T_res.sum() == 0:
```
kbattocchi, line 764: Again, the issue is not that the residuals sum to zero, it's that every entry is zero.

```suggestion
        if not T_res.any():
```

econml/iv/dml/_dml.py
```diff
@@ -761,6 +761,14 @@ def predict(self, Y, T, X=None, W=None, Z=None, sample_weight=None, groups=None)
...
+            raise ValueError(
+                """
+                All values of the treatment residual are 0, 
+                which then makes them unsuitable to use as weights 
+                in downstream in econml/dml/dml.py
```
kbattocchi, line 769: Linting is failing because these lines all end in trailing whitespace.  I'd tweak the last part of the message slightly:

```suggestion
                All values of the treatment residual are 0,
                which then makes them unsuitable to use as weights.
                DRIV requires that the instrument Z has an effect on the
                expected treatment value of at least some rows.

                If you are using regularized models, it's possible that this error is a
                result of regularizing too strongly, so that all predictions from both
                models are constant.
```

kbattocchi (changes requested):
This looks much better, just needs a couple of tweaks.

ronikobrosly:
Thank you for the suggestions and your patience @kbattocchi . Accepted your changes.

kbattocchi:
One of your commits was not signed, so this is not passing the DCO check, see https://github.com/py-why/EconML/pull/919/checks?check_run_id=31527396832 for instructions on how to fix this (under Rebase the Branch).

Also it looks like the test that you had originally added has subsequently been removed in another commit - could you please add it back?

PR #923, opened 2024-10-09
Fix coverage upload

kbattocchi:
A breaking change to the upload-artifact GitHub Action has broken uploading coverage files (not only for us, see https://github.com/actions/upload-artifact/issues/614) and has not been reverted. This change will fix that by enabling hidden files only for those uploads.


fverac (approved):
Looks good. 

Worth adding a comment to clarify why we need the hidden files flag? Up to you

PR #927, opened 2024-11-14
Enable Discrete Outcomes w/ RScorer

jakepenzak:
**Motivation:**
- To enable usage of `RScorer` class for scoring and model ensembling in the presence of discrete outcomes

**Implementation:**
- Add `discrete_outcome` attribute to `RScorer` class and pass to `LinearDML` in `RScorer.fit()` method

**File modifications:**
- `econml/score/rscorer.py` - Added `discrete_outcome` attribute
- `econml/tests/test_rscorer.py` - Added test case for discrete outcomes and CATEs that take probabilistic form

### DIFF
--- a/econml/tests/test_rscorer.py
+++ b/econml/tests/test_rscorer.py
@@ -33,6 +33,9 @@ class TestRScorer(unittest.TestCase):
         else:
             y = X[:, 0] * T + np.random.normal(size=(100000,))
             true_eff = X[:, 0]
+
+        y = y.reshape(-1, 1)
+        T = T.reshape(-1, 1)
         return y, T, X, true_eff
 
     def test_comparison(self):
@@ -114,10 +117,6 @@ class TestRScorer(unittest.TestCase):
                                 for _, mdl in models]
                 # Checking neg corr between rscore and rootpehe (precision in estimating heterogeneous effects)
                 assert LinearRegression().fit(np.array(rscore).reshape(-1, 1), np.array(rootpehe_score)).coef_ < 0.5
-
-                if discrete_outcome:
-                    # Removing SLearner. Only utilized for introducting variation in r- & rootpehe score
-                    models = [m for m in models if m[0] != 'slearner']
                 mdl, _ = scorer.best_model([mdl for _, mdl in models])
                 rootpehe_best = np.sqrt(np.mean((true_eff_val.flatten() - mdl.effect(X_val).flatten())**2))
                 # Checking best model selection behaves as intended
--- a/econml/tests/test_rscorer.py
+++ b/econml/tests/test_rscorer.py
@@ -20,52 +20,106 @@ def _fit_model(name, model, Y, T, X):
 
 class TestRScorer(unittest.TestCase):
 
-    def _get_data(self):
+    def _get_data(self, discrete_outcome=False):
         X = np.random.normal(0, 1, size=(100000, 2))
         T = np.random.binomial(1, .5, size=(100000,))
-        y = X[:, 0] * T + np.random.normal(size=(100000,))
-        return y, T, X, X[:, 0]
+        if discrete_outcome:
+            y = (X[:, 0] * T + np.random.normal(size=(100000,)) > 0).astype(int)
+            # Difference in conditional probabilities P(y=1|X,T=1) - P(y=0|X,T=0)
+            true_eff = (1 / (1 + np.exp(-X[:, 0]))) - (1 / (1 + np.exp(0)))
+        else:
+            y = X[:, 0] * T + np.random.normal(size=(100000,))
+            true_eff = X[:, 0]
+        return y, T, X, true_eff
 
     def test_comparison(self):
+
         def reg():
             return LinearRegression()
 
         def clf():
             return LogisticRegression()
 
-        y, T, X, true_eff = self._get_data()
-        (X_train, X_val, T_train, T_val,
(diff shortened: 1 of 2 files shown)


econml/tests/test_rscorer.py
```diff
@@ -20,52 +20,109 @@ def _fit_model(name, model, Y, T, X):
...
         X = np.random.normal(0, 1, size=(100000, 2))
         T = np.random.binomial(1, .5, size=(100000,))
-        y = X[:, 0] * T + np.random.normal(size=(100000,))
-        return y, T, X, X[:, 0]
+        if discrete_outcome:
+            eps = np.random.normal(size=(100000,))
+            log_odds = X[:, 0]*T + eps
+            y_sigmoid = 1/(1 + np.exp(-log_odds))
+            y = np.array([np.random.binomial(1, p) for p in y_sigmoid])
+            # Difference in conditional probabilities P(y=1|X,T=1) - P(y=0|X,T=0)
```
fverac, line 31: Should this be
```suggestion
            # Difference in conditional probabilities P(y=1|X,T=1) - P(y=1|X,T=0)
```
?

jakepenzak: Yes, that was a typo on my end. Thanks for catching that!

econml/tests/test_rscorer.py
```diff
@@ -20,52 +20,109 @@ def _fit_model(name, model, Y, T, X):
...
+
+                scorer.fit(Y_val, T_val, X=X_val)
+                rscore = [scorer.score(mdl) for _, mdl in models]
+                rootpehe_score = [np.sqrt(np.mean((true_eff_val.flatten() - mdl.effect(X_val).flatten())**2))
+                                for _, mdl in models]
+                # Checking neg corr between rscore and rootpehe (precision in estimating heterogeneous effects)
+                assert LinearRegression().fit(np.array(rscore).reshape(-1, 1), np.array(rootpehe_score)).coef_ < 0.5
+
+                if discrete_outcome:
+                    # Removing SLearner. Only utilized for introducting variation in r- & rootpehe score
```
fverac, line 119: I see that if you leave in SLearner in with the current code, the test does fail, but I think we can make a different change such that we can remove this if statement.

The ensemble tests fail when SLearner is included and `discrete_outcome=True` because there ends up being an inconsistency between the output shape of `const_marginal_effect` between all of the other discrete outcome models and `SLearner`. In short, the reason for this is that when `discrete_outcome=True`. the shape of y in supported models internally becomes (n, 1). However, since metalearners (and thus SLearner) don't support discrete_outcome, the shape of y internally still remains (n,). So we end up getting different const_marginal_effect shapes which makes numpy complain when e.g. taking the average over these const_marginal effects.

There are a few ways around this.
1. We can include the if-statement to exclude SLearner as you do currently.
2. We can revise the _get_data to make `y` and `T` of shape (n ,1) rather than (n,). Testing locally, this works.
3. A more appropriate long term fix might be to extend discrete_outcome functionality to metalearners, but I think that is out-of-scope for this PR.

Can we implement the 2nd workaround? I'll add a code suggestion shortly

econml/tests/test_rscorer.py
```diff
@@ -20,52 +20,109 @@ def _fit_model(name, model, Y, T, X):
...
+            eps = np.random.normal(size=(100000,))
+            log_odds = X[:, 0]*T + eps
+            y_sigmoid = 1/(1 + np.exp(-log_odds))
+            y = np.array([np.random.binomial(1, p) for p in y_sigmoid])
+            # Difference in conditional probabilities P(y=1|X,T=1) - P(y=1|X,T=0)
+            true_eff = (1 / (1 + np.exp(-(X[:, 0]+eps)))) - (1 / (1 + np.exp(-eps)))
+        else:
+            y = X[:, 0] * T + np.random.normal(size=(100000,))
+            true_eff = X[:, 0]
+        return y, T, X, true_eff
```
fverac, line 36:
```suggestion
        y = y.reshape(-1, 1)
        T = T.reshape(-1, 1)
        return y, T, X, true_eff
```

Code suggestion RE: discussion below

econml/tests/test_rscorer.py
jakepenzak: Yes, I definitely agree option 2 makes more sense in this scenario. Thanks for the recommendation here!

econml/tests/test_rscorer.py
jakepenzak: Implemented these changes below :100:

fverac:
Code looks ready to merge.

The DCO check is failing though, which probably means you'll have to rebase the branch. Can you follow the instructions here?
https://github.com/py-why/EconML/pull/927/checks?check_run_id=33428718307

fverac:
Hm seems like we're not done yet. 

Says the branch is out-of-date with main. I think the problem is that in the current rebase there is a commit that merged in some updates from main ("pre commit autoupdate"), and git is not liking that.

Instead, can you update your local main branch and rebase from there?

jakepenzak:
> Hm seems like we're not done yet.
> 
> Says the branch is out-of-date with main. I think the problem is that in the current rebase there is a commit that merged in some updates from main ("pre commit autoupdate"), and git is not liking that.
> 
> Instead, can you update your local main branch and rebase from there?

Ah yes, I missed that on my end. I believe everything should be good to go from here.

fverac (approved):

fverac:
Merging now, thanks for the contribution!

PR #929, opened 2024-11-21
fix `dowhy_` not existance issue

louis-she:
1. `object` has no `__getattr__` attribute so calling `super().__getattr__` is wrong, should change it to `getattr(self, name)`.
2. `dowhy_` is not exist until `fit` is called, just have a more intuitive error message when someone calls methods without `fitting`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.6.2
+  rev: v0.7.3
   hooks:
     # Run the linter.
     - id: ruff
--- a/econml/dowhy.py
+++ b/econml/dowhy.py
@@ -233,11 +233,14 @@ class DoWhyWrapper:
             raise AttributeError(attr)
         elif attr == "dowhy_":
             if "dowhy_" not in dir(self):
-                raise RuntimeError("Please call DoWhyWrapper.fit first before any other operations")
+                raise AttributeError("Please call `DoWhyWrapper.fit` first before any other operations.")
             else:
                 return self.dowhy_
         elif attr in ['_cate_estimator', 'identified_estimand_', 'estimate_']:
-            return getattr(self, attr)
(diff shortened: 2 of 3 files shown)


kbattocchi (changes requested):
Thanks for the contribution! I have two requests:

1. This repo requires all commits to be signed off (which yours isn't), which is causing the DCO check to fail - please amend your commit to sign off on it (there are detailed instructions that you can see if you look at the details of the failing DCO check).
2. Please add a test case that fails with the current version of the code but passes with your changes.

louis-she:
if without this patch,
```python
LinearDRLearner().dowhy.refute_estimate(method_name="random_common_cause", num_simulations=3)
# or
LinearDRLearner().dowhy._estimator_object
```
will raise
```
  File "/Users/louisshe/Downloads/EconML/econml/tests/test_dowhy.py", line 100, in test_dowhy_without_fit
    LinearDRLearner().dowhy.refute_estimate(method_name="random_common_cause", num_simulations=3)
  File "/Users/louisshe/Downloads/EconML/econml/dowhy.py", line 220, in refute_estimate
    return self.dowhy_.refute_estimate(
  File "/Users/louisshe/Downloads/EconML/econml/dowhy.py", line 236, in __getattr__
    return super().__getattr__(attr)
AttributeError: 'super' object has no attribute '__getattr__'
```

`super()` is `object` because the defination `class DoWhyWrapper`. And object does not has `__getattr__`.

```
>>> object.__getattr__
  File "<stdin>", line 1, in <module>
AttributeError: type object 'object' has no attribute '__getattr__'. Did you mean: '__setattr__'?
```

So this is incorrect without a doult, instead we should use getattr here, and to prevent infinite recursion, should add a check before calling `getattr` or `hasattr` in `__getattr__`.

----------------------------------------

All the cases will only happen when use in a wrong way, "accessing attributes without calling `fit`", so i just added some more intuitive error message comparing the old error message `AttributeError: 'super' object has no attribute '__getattr__'
`.

kbattocchi (approved):
This looks great, thanks!

Issue #932, opened 2024-11-29
Fix the "Help wanted" link in the readme


KomiB-Atsou:
The "Help Wanted" link in the readme is no longer working.
Current link : https://github.com/py-why/EconML?tab=readme-ov-file#help-wanted
Correct link : https://github.com/py-why/EconML?tab=readme-ov-file#finding-issues-to-help-with

PR #934, opened 2024-11-29
Correct "Help Wanted" url in readme

KomiB-Atsou:
https://github.com/py-why/EconML/issues/932

The "Help Wanted" link in the readme is no longer working.
Current link : https://github.com/py-why/EconML?tab=readme-ov-file#help-wanted
Correct link : https://github.com/py-why/EconML?tab=readme-ov-file#finding-issues-to-help-with

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ For information on use cases and background material on causal inference and het
 
 # News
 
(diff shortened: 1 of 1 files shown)


kbattocchi (approved):
Thanks for fixing this!

PR #939, opened 2024-12-11
Miscellaneous fixes

### LINKED ISSUE #938: sklearn.tree.plot_tree does not work with econml.grf.CausalForest
Hello, I would like to plot a causal tree from a causal forest. I tried running the "1. Causal Forest: Heterogeneous causal effects with no unobserved confounders" section from [this notebook](https://github.com/py-why/EconML/blob/main/notebooks/Generalized%20Random%20Forests.ipynb) but I receive an error when trying to plot a causal tree using plot_tree from sklearn.tree (see cell 8 in the notebook):

> InvalidParameterError: The 'decision_tree' parameter of plot_tree must be an instance of 'sklearn.tree._classes.DecisionTreeClassifier' or an instance of 'sklearn.tree._classes.DecisionTreeRegressor'. Got GRFTree(criterion='het', max_features='auto', min_var_leaf_on_val=True,
>         random_state=1465622306) instead."

It seems that this worked in the past, perhaps this works with an older version of sklearn or econml? For reference, I am using sklearn version '1.4.2' and econml version '0.15.0'.

kbattocchi:
Fixes #938, plus cleans up some other minor issues, including enabling sklearn 1.6 and fixing our documentation builds.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -231,7 +231,7 @@ intersphinx_mapping = {'python': ('https://docs.python.org/3', None),
                        'sklearn': ('https://scikit-learn.org/1.5/', None),
                        'matplotlib': ('https://matplotlib.org/stable/', None),
                        'shap': ('https://shap.readthedocs.io/en/stable/', None),
-                       'dowhy': ('https://www.pywhy.org/dowhy/v0.8/', None),
+                       'dowhy': ('https://www.pywhy.org/dowhy/main/', None),
                        'statsmodels': ('https://www.statsmodels.org/stable/', None)}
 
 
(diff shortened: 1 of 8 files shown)


doc/conf.py
```diff
@@ -228,10 +228,10 @@
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {'python': ('https://docs.python.org/3', None),
                        'numpy': ('https://numpy.org/doc/stable/', None),
-                       'sklearn': ('https://scikit-learn.org/stable/', None),
+                       'sklearn': ('https://scikit-learn.org/1.5/', None),
```
kbattocchi, line 231: Even though this PR enables support for sklearn 1.6, that doesn't help here because our docs still build on Python 3.8 (because of the tensorflow dependency for Deep IV), where sklearn 1.5 is the latest version.

We need to specifically link to this version of the sklearn docs because there was a change in the structure of their docs between 1.5 and 1.6, so the mismatch between what we were building against vs. what we were linking to was causing broken links.

fverac (approved):
Looks good!

PR #950, opened 2025-01-29
carlg/alternative-metrics

carl-offerfit:
- Adds an optional scoring argument to _OrthoLearner.score so that the score can be viewed according to several other sklearn metrics. Currently supported are: mean_absolute_error, mean_squared_error, r2_score.
- The original default behavior of using MSE for the model fitting is preserved, now calculated in sklearn mean_square_error function instead of written out in code.
- Adds a new public method score_nuisances to _OrthoLearner.score so that the first stage models can be evaluated by other sklearn metrics. If sample weights are not used, any metric supported by sklearn get_scoring can be used. If sample weights are used, f1_score, log_loss, roc_auc_score are supported for binary outcomes/treatments; and for real valued outcome treatments it will be the same as _OrthoLearner.score for the final model mean_absolute_error, mean_squared_error, r2_score.


carl-offerfit:
Closing PR because of rebase issues - will make a new PR

PR #964, opened 2025-03-31
Carl/metrics

carl-offerfit:
- Adding the option to use alternative sklean score functions for both first and second stage models in DML
- Add score_nuisances function


carl-offerfit:
More rebase issues - going to try again with a totally new branch

PR #965, opened 2025-03-31
Alternative scoring metrics

carl-offerfit:
- Allow specification of an alternative sklearn score function for double ML models
- Add score_nuisances function

### DIFF
--- a/econml/_ortho_learner.py
+++ b/econml/_ortho_learner.py
@@ -1027,7 +1027,7 @@ class _OrthoLearner(TreatmentExpansionMixin, LinearCateEstimator):
 
     effect_inference.__doc__ = LinearCateEstimator.effect_inference.__doc__
 
-    def score(self, Y, T, X=None, W=None, Z=None, sample_weight=None, groups=None):
+    def score(self, Y, T, X=None, W=None, Z=None, sample_weight=None, groups=None, scoring=None):
         """
         Score the fitted CATE model on a new data set.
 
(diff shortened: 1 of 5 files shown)


carl-offerfit:
@kbattocchi Here's an update:

1. I fixed the type hints to use the old `Union` approach rather than the python 3.10 `|`. That should fix the Python 3.8 & 3.9 errors.
2. I added a unit test for the ML score signature validation
3. I researched sklearn `make_scorer` and as far as I can tell it doesn't do any validation at all - if the function you passed to `make_scorer` is not valid it just fails when you try to use it. That makes me wonder if my approach was over engineered, and instead we should just let it fail. At the same time, I'm a bit attached to the approach since I spent time on it. ;-) WDYT?

carl-offerfit:
@kbattocchi I fixed a bug that was cause of the test failures. Please let me know if there are still issues.

pyproject.toml
```diff
@@ -105,7 +105,7 @@ exclude = ["econml.tests"]
 
 [tool.pytest.ini_options]
 testpaths = ["econml/tests"]
-addopts = "--junitxml=junit/test-results.xml -n auto --strict-markers --cov-config=pyproject.toml --cov --import-mode=importlib"
+addopts = "--junitxml=junit/test-results.xml -n0 --strict-markers --cov-config=pyproject.toml --import-mode=importlib"
```
kbattocchi, line 108: Were you just changing these to run locally? We definitely want to use xdist and get coverage results, so please revert this change. If your own test needs to run serially you can use a `@pytest.mark.serial` attribute to mark your method (or class).

kbattocchi (changes requested):
Please revert the changes to pyproject.toml; also the DCO check is still failing because of commits that weren't signed off correctly.

pyproject.toml
carl-offerfit: Yes, it was to run the tests locally. I did not mean to commit it, sorry!

carl-offerfit:
@kbattocchi I think I fixed the issues with the pyproject.toml and the sign-off. Let me know if you see any other issues.

kbattocchi:
Closing in favor of #988

PR #968, opened 2025-05-07
Fix CI build by limiting builds to not py 13

BlueDrink9:
Example of passing build here:
https://github.com/BlueDrink9/EconML/actions/runs/14893898354


kbattocchi:
It is by design that we include python 3.13 in our nightly builds, just as we also include versions of our dependencies more recent than those that are not known to be good - we do not expect the nightly tests to always pass, they flag issues with forward compatibility that we eventually want to address.

PR #969, opened 2025-05-08
Try to support numpy version 2

BlueDrink9:
This is well and truly in draft mode, and may never get off the ground without some serious decision making. As such, I'm not really expecting this to have this merged, but perhaps the work that I have done here might be helpful for someone else down the line.

Numpy 2 does not support python 3.8, which straightaway rules out some of the optional dependencies that EconML has (eg pytorch). To get it building and testing, I have just removed them, and so it would probably require some more work if it was desirable to keep those.

I'm not experienced with cython, and a lot of the cython code here was not particularly clear to me from the sparse comments, so most of what I've done is really on a best guess effort to get it to successfully compile. That being said, some of the tests are failing with segfaults so clearly I have made some mistakes somewhere. I have highlighted where I think these most likely are, but it will need someone who actually understands cython/the numpy API to assess and fix.


econml/tree/_tree.pyx
BlueDrink9, line 1: Damn, I think this was a end of line change. There are some legitimate changes in here as well as that. I tried rebasing and then forcing the correct line endings but it hasn't removed it from the diff, so I will have to come back to this

BlueDrink9: Unfortunately, this file contains most of the concrete cython changes.

In particular, I suspect one of these types of changes is the issue.
`NPY_ARRAY_DEFAULT` was `NPY_DEFAULT`, but I cannot find any reference to the latter anywhere on the internet other than three obscure pages, certainly nothing in the numpy documentation.
That parameter should take a flags argument, which `NPY_ARRAY_DEFAULT` is and was the only close sounding flag that I could find to the original. Alternative consideration was `NPY_DEFAULT_TYPE`, but I don't think that is the right type to fit here.

The second change was using an api setter method rather than directly setting self.arr, since that is now a read only field. The setter steals the reference, which might be responsible for the segfaults if it is being freed then used. The INCREF was already there, but I have left it to ensure that the steal doesn't take every reference. It might be worth stacking another reference increment there and see if that fixes the segfaults.

```
        arr = PyArray_NewFromDescr(<PyTypeObject *> np.ndarray,
                                   <np.dtype> NODE_DTYPE, 1, shape,
                                   strides, <void*> self.nodes,
                                   np.NPY_ARRAY_DEFAULT, None)
        Py_INCREF(self)
        PyArray_SetBaseObject(arr, <PyObject*> self)
```

BlueDrink9:
I'm a bit confused by the lkg.txt files, how do we regenerate or skip the requirement to use those?

BlueDrink9:
`git ls-files --eol` shows that crlf in this repo is incredibly inconsistent. I would highly recommend, at some point, doing dos2unix or similar on every file, to avoid issues around line endings in the future!

econml/tree/_tree.pyx
BlueDrink9: I managed to fix and rebase the line ending change, for posterity the issue was this entire file showed up as changed because it uses eol=crlf whereas other parts of the repo uses eof=lf. Git checked it back in with lf. I'm surprised that this doesn't seem to have caused issues in previous PRs, I would highly encourage normalizing the line endings and creating a git attributes file to specify the endings going forward

BlueDrink9:
A bunch of the continuous integration steps will need to be updated to use a later version of python rather than 3.8 as part of this

BlueDrink9:
Passing tests with a lot of warnings https://github.com/BlueDrink9/EconML/actions/runs/14917865406/job/41907412454

kbattocchi:
Thanks, I've cherry-picked one of your commits into another PR (#984) attempting to address this and other issues.

One issue with both your attempt and mine is that several tests run *dramatically* slower (see, e.g. the ubuntu-latest 3.10 ray tests, which take about 2 hours with these changes vs. 20 minutes without them). If you've got any insights on how to mitigate that performance regression I'm all ears.

PR #975, opened 2025-05-22
add Validation docs

fverac:


doc/spec/validation.rst
```diff
@@ -0,0 +1,68 @@
...
+to assess the robustness of causal estimates to unobserved confounding. 
+
+Specifically, select estimators (subclasses of :class:`.DML` and :class:`.DRLearner`)
+have access to ``sensitivity_analysis``, ``robustness_value``, and ``sensitivity_summary`` methods.
+
+``sensitivity_analysis`` provides an updated confidence interval for the ATE based on a specified level of unobserved confounding.
+
+
+``robustness_value`` computes the minimum level of unobserved confounding required
+to make it impossible to reject a null hypothesis (default 0).
```
kbattocchi, line 27: The phrasing here feels slightly awkward to me; maybe something more like

```suggestion
so that confidence intervals around the ATE would begin to include the given point (0 by default).
```

kbattocchi (approved):
This is great, just one minor suggestion on wording.

doc/spec/validation.rst
```diff
@@ -0,0 +1,68 @@
...
+
+Sensitivity Analysis
+---------------------
+
+For many EconML estimators, unobserved confounding can lead to biased causal estimates.
+Moreover, it is impossible to prove the absence of unobserved confounders.
+This is a fundamental problem for observational causal inference.
+
+To mitigate this problem, EconML provides a suite of sensitivity analysis tools,
+based on `Chernozhukov et al. (2022) <https://www.nber.org/papers/w30302>`_,
```
kbattocchi, line 17: One more comment - consider adding this paper to the references page and linking to that here instead.

doc/spec/validation.rst
fverac: How about this. Might be good to use the words null hypothesis because that is the relevant kwarg in the actual message.
```suggestion
so that confidence intervals around the ATE would begin to include the null hypothesis (0 by default).
```

kbattocchi: I think the way I'd use "null hypothesis" in a sentence is something like "the null hypothesis is that the ATE is 0", not "the null hypothesis is 0"; in the context of our API I think it's fine that we take a number as the argument with that name but in prose it seems awkward to me to claim that a confidence interval could "include the null hypothesis". I think your original sentence about rejecting the null is accurate (but I'd characterize the default null as `(that the ATE is 0 by default)` rather than your parenthetical), but the "impossible to reject" seems harder to think about than a more positive framing even though it's correct.

PR #996, opened 2025-08-13
Add Clustered Standard Errors Support

msukiasyan:
Adding support for cluster-robust standard errors.

Implementation:
- Add clustered variance computation methods to `StatsModelsLinearRegression` and `StatsModels2SLS` classes, using existing groups parameter plumbing. Matches `linearmodels` package implementation.
- Add `cov_type` parameter to `OrthoIV` constructor to expose the option since `model_final` isn't an argument.

File modifications:
- `econml/sklearn_extensions/linear_model.py -` Added `_compute_clustered_variance()` methods
- `econml/iv/dml/_dml.py` - Added `cov_type` parameter to `OrthoIV` class, pass `groups` parameter to `model_final`.
- `econml/dml/_rlearner.py` - Pass groups parameter to `model_final`.
- `econml/tests/test_clustered_se.py` - Tests validating against `statsmodels` implementation (up to small sample correction).

First contribution, so please let me know if I've missed anything!

### DIFF
--- a/econml/sklearn_extensions/linear_model.py
+++ b/econml/sklearn_extensions/linear_model.py
@@ -1943,14 +1943,18 @@ class StatsModelsLinearRegression(_StatsModelsWrapper):
         if len(models) == 0:
             raise ValueError("Must aggregate at least one model!")
         cov_types = set([model.cov_type for model in models])
+        cov_options = set([str(model.cov_options) for model in models])
         fit_intercepts = set([model.fit_intercept for model in models])
         enable_federation = set([model.enable_federation for model in models])
         _n_outs = set([model._n_out for model in models])
         assert enable_federation == {True}, "All models must have enable_federation=True!"
         assert len(cov_types) == 1, "All models must have the same cov_type!"
+        assert len(cov_options) == 1, "All models must have the same cov_options!"
         assert len(fit_intercepts) == 1, "All models must have the same fit_intercept!"
         assert len(_n_outs) == 1, "All models must have the same number of outcomes!"
-        agg_model = StatsModelsLinearRegression(cov_type=models[0].cov_type, fit_intercept=models[0].fit_intercept)
+        agg_model = StatsModelsLinearRegression(cov_type=models[0].cov_type,
+                                                cov_options=models[0].cov_options,
+                                                fit_intercept=models[0].fit_intercept)
 
         agg_model._n_out = models[0]._n_out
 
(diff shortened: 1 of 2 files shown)


kbattocchi:
> @kbattocchi was wondering if you had a chance to look into this PR and have any feedback I can work on. Thank you!

Apologies for the slow response, I was traveling and missed this.

First of all, thanks for the contribution, this is definitely a valuable feature that we'd love to have.

In terms of high-level feedback, one thing is that I think we'd want to support federalization even with clustered standard errors, with the restriction that groups are assumed to be fully partitioned to each individual estimator (or in other words, group "1" for the first estimator is assumed to be different from group "1" in the second estimator). I believe that this should then be straightforward because again you can just compute the moments locally and combine them, except that you may add a count to the state to keep track of the number of groups when aggregating later (I'm not sure).

The only other thing that comes to mind immediately is a question: is the bias-correction for clustered errors completely standard, or would it make sense to provide options for this akin to HC0 vs HC1 for non-clustered errors, where we can optionally correct for the degrees of freedom?

msukiasyan:
@kbattocchi thanks for the feedback!

On federalization: I think that makes sense, I'll try to add that soon.

On bias-correction: libraries do differ a bit on the defaults and whether they allow switching bias-correction on/off. Across Statsmodels/linearmodels/Stata, there is only a single "cluster" string option. Statsmodels and linearmodels both allow to toggle bias-correction via a separate argument but, from what I can tell, Stata just sticks with the (G/(G-1))*((N - 1)/(N - k)) correction ([manual](https://www.stata.com/manuals/u20.pdf#u20.22Obtainingrobustvarianceestimates), page 51). I think it makes sense to have that same correction as default but I'm unsure if we want to a) expose a separate use_correction argument in high level API vs b) use modified strings like "cluster_HC0" vs c) no toggling allowed at all. Any thoughts on this?

msukiasyan:
Updates:
1. Federated learning
    a. Added federated support for clustered covariance. This requires storing a few additional moments and n_groups for each learner. Added tests.
    b. Fixed an issue with summarized data (frequencies > 1) and clustered covariance. Added a test to catch this.
2. Small sample corrections
    a. Set small sample correction to match Stata/statsmodels defaults, edited tests to reflect
    b. Added cov_options to allow toggling like in statsmodels

Will mark this ready for review for now.

msukiasyan:
Hi @kbattocchi - quick ping on this PR. If any reviewer has time to look into the updates, I’d appreciate it!

PR #1000, opened 2025-09-05
Add scoring parameter to drlearner

gdaiha:
Add scoring parameter to drlearner, to allow score other metrics than MSE in Doubly Robust models


kbattocchi:
Yes, the failure is unrelated to your changes, it looks like the build runners upgraded to Python 3.13.7 and we're hitting https://github.com/python/cpython/issues/138031. I'll see what I can do to address it.

gdaiha:
I'm going to add the adjustments needed in _drlearner.py to make scoring parameter work, very similar to what've been done in DML

PR #1003, opened 2025-09-10
Update lkg

kbattocchi:
Our build machines running Python 3.13 have updated to using Python 3.13.7, which has resulted in some Windows tests failing due to https://github.com/python/cpython/issues/138031; this has been worked around in a more recent version of joblib, so updating the LKG should fix this.


vamsikondru2001:
Hi @kbattocchi  how to builid a multi treatement and binary outcome model for email and sms optimization ?
can I use causal forest dml for this ? will it worl for my use case ? like email and sms optimization use case..

PR #1007, opened 2025-09-22
Add support for new scoring methods in DRLearner

gdaiha:
Add support for customize scoring methods in DRLearner, similar done to DML. 
This is a remake of #1000 to avoid problems with not signed-off commits.


kbattocchi:
Please add tests that cover this code

PR #1011, opened 2025-10-17
License problem with 16 when installing on Databricks

dwsmith1983:
PEP 621 violation causes issues with installing databricks.

Signed-off-by: Dustin Smith dustin.william.smith@gmail.com

Fixes py-why/EconML#1010

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ authors = [{ name = "PyWhy contributors" }]
 description = "This package contains several methods for calculating Conditional Average Treatment Effects"
 readme = "README.md"
 keywords = ["treatment-effect"]
-license = "MIT"
+license = { text = "MIT" }
 dynamic = ["version"]
 
 classifiers = [


pyproject.toml
```diff
@@ -5,7 +5,7 @@ authors = [{ name = "PyWhy contributors" }]
 description = "This package contains several methods for calculating Conditional Average Treatment Effects"
 readme = "README.md"
 keywords = ["treatment-effect"]
-license = "MIT"
```
kbattocchi: PEP 621 has been superseded by PEP 639 and is now deprecated - see https://packaging.python.org/en/latest/specifications/pyproject-toml/#license and https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files, so we definitely do not want to make this change.

This is not a problem when using a recent version of setuptools.  I think just updating setuptools (via `pip install -U setuptools` if you're in an environment managed by pip) should fix any issues; perhaps we could change line 69 to `"setuptools>77.03"` to force a more recent version but reverting to the deprecated license scheme doesn't make sense.

PR #1029, opened 2026-05-18
Fix off-by-one in score_nuisances for discrete treatments (closes #1006)

### LINKED ISSUE #1006: score_nuisances with discrete treatment returns incorrect score
When using score_nuisances with a discrete treatment, the function does not return the correct score.

The issue comes from the inverse_onehot function in econml/utilities.py. Currently, when it receives as input a DataFrame generated by pandas.get_dummies(), it incorrectly decodes the treatment.

For example, in case of binary treatments, labels originally coded as 0 and 1 are shifted and end up being decoded as 1 and 2, due to the following implementation:

```
def inverse_onehot(T):
    """
    Given a one-hot encoding of a value, return a vector reversing the encoding to get numeric treatment indices.

    Note that we assume that the first column has been removed from the input.
    """
    assert ndim(T) == 2
    # note that by default OneHotEncoder returns float64s, so need to convert to int
    return (T @ np.arange(1, T.shape[1] + 1)).astype(int)
```

This logic introduces an off-by-one error when decoding treatments.

**Expected behavior**

The function should return zero-based indices, ensuring that discrete treatments (e.g. 0/1) remain consistent after decoding. A corrected implementation would have the following code:

```
def inverse_onehot(T):
    assert econml.utilities.ndim(T) == 2

    indices = (
        np.arange(0, T.shape[1])
        if isinstance(T, pd.DataFrame)
        else np.arange(1, T.shape[1] + 1)
    )

    return (T @ indices).astype(int)
```

This change guarantees that score_nuisances computes the correct score for discrete treatments.

**Contributed by @Cantal00p, @f5ilverio**

immu4989:
https://github.com/py-why/EconML/issues/1006#issuecomment-4392015743

## Evidence

Reproducer: ternary T (3 classes) with signal in X, `LinearDML` +
`LogisticRegression` nuisance, `cv=3`. Mean `score_nuisances` across folds:

| Metric                | Pre-fix | Post-fix | Notes                              |
| --------------------- | ------- | -------- | ---------------------------------- |
| T default score (acc) | **0.13**| **0.63** | 1/3 chance floor for 3-class       |
| Y default score (R²)  |  0.44   |  0.44    | unchanged — bug isolates to T path |

Pre-fix the classifier scores **below random** because `pd.get_dummies(T)`
keeps all *k* columns and `inverse_onehot` decodes labels as `{1..k}` while
the classifier was trained on `{0..k-1}`. Post-fix the same classifier
recovers genuine signal.

The existing `test_forest_dml_score_fns` already exercised this path and
its expected values silently encoded the bug; they're updated here to the
correct post-fix values:

| Assertion (`T_*`)   | Was (pre-fix)  | Now (post-fix) |
| ------------------- | -------------- | -------------- |
| `mean_squared_error`|  1.5           |  0.48          |
| `mean_absolute_error`|  1.0          |  0.48          |
| `r2`                | −5.1           | −0.93          |

`roc_auc`, `log_loss`, and `pearsonr` are approximately invariant under
the off-by-one label shift, which is why the existing test didn't fail
loud — the three metrics where label distance matters were the ones that
silently absorbed the gap.

### DIFF
--- a/econml/dml/_rlearner.py
+++ b/econml/dml/_rlearner.py
@@ -34,7 +34,7 @@ from sklearn.metrics import (
 )
 from typing import Callable, Union
 from ..sklearn_extensions.model_selection import ModelSelector
-from ..utilities import (filter_none_kwargs)
+from ..utilities import (filter_none_kwargs, reshape)
 from .._ortho_learner import _OrthoLearner
 
 class _ModelNuisance(ModelSelector):
@@ -628,9 +628,13 @@ class _RLearner(_OrthoLearner):
             T_Key : []
         }
 
-        # For discrete treatments, these will have to be one hot encoded
         Y_2_score = pd.get_dummies(Y) if self.discrete_outcome and (len(Y.shape) == 1 or Y.shape[1] == 1) else Y
-        T_2_score = pd.get_dummies(T) if self.discrete_treatment and (len(T.shape) == 1 or T.shape[1] == 1) else T
+        if self.transformer:
+            if self.discrete_treatment:
+                T = reshape(T, (-1, 1))
+            T_2_score = self.transformer.transform(T)
+        else:
+            T_2_score = T
 
         for m in self._models_nuisance[0]:
             Y_score, T_score = m.score(Y_2_score, T_2_score, X=X, W=W, Z=Z, sample_weight=sample_weight,
--- a/econml/tests/test_dml.py
+++ b/econml/tests/test_dml.py
@@ -789,19 +789,19 @@ class TestDML(unittest.TestCase):
                                   t_scoring='mean_squared_error',
                                   y_scoring='mean_squared_error')
         np.testing.assert_allclose(sn1['Y_mean_squared_error'], [2.8,2.8], rtol=0, atol=.1)
-        np.testing.assert_allclose(sn1['T_mean_squared_error'], [1.5,1.5], rtol=0, atol=.1)
+        np.testing.assert_allclose(sn1['T_mean_squared_error'], [0.48,0.48], rtol=0, atol=.1)
 
         sn2 = est.score_nuisances(Y=y, T=T, X=X, W=W,
                                   t_scoring='mean_absolute_error',
                                   y_scoring='mean_absolute_error')
         np.testing.assert_allclose(sn2['Y_mean_absolute_error'], [1.3,1.3], rtol=0, atol=.1)
-        np.testing.assert_allclose(sn2['T_mean_absolute_error'], [1.0,1.0], rtol=0, atol=.1)
+        np.testing.assert_allclose(sn2['T_mean_absolute_error'], [0.48,0.48], rtol=0, atol=.1)
 
         sn3 = est.score_nuisances(Y=y, T=T, X=X, W=W,
                                   t_scoring='r2',
                                   y_scoring='r2')
         np.testing.assert_allclose(sn3['Y_r2'], [0.27,0.27], rtol=0, atol=.005)
-        np.testing.assert_allclose(sn3['T_r2'], [-5.1,-5.1], rtol=0, atol=0.25)
+        np.testing.assert_allclose(sn3['T_r2'], [-0.93,-0.93], rtol=0, atol=0.05)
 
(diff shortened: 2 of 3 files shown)


econml/dml/_rlearner.py
```diff
@@ -628,9 +628,15 @@ def score_nuisances(self, Y, T, X=None, W=None, Z=None, sample_weight=None, y_sc
...
+        # For discrete treatments, re-use the OneHotEncoder fit during training
+        # (drop='first') so the encoding matches what the nuisance classifier
+        # learned. pd.get_dummies(T) keeps all k columns, which makes
+        # inverse_onehot decode labels off-by-one in _FirstStageWrapper.score.
         Y_2_score = pd.get_dummies(Y) if self.discrete_outcome and (len(Y.shape) == 1 or Y.shape[1] == 1) else Y
-        T_2_score = pd.get_dummies(T) if self.discrete_treatment and (len(T.shape) == 1 or T.shape[1] == 1) else T
+        if self.discrete_treatment and (len(np.shape(T)) == 1 or np.shape(T)[1] == 1):
+            T_2_score = self.transformer.transform(np.asarray(T).reshape(-1, 1))
+        else:
+            T_2_score = T
```
kbattocchi: Thanks for submitting this fix!

Unfortunately, I believe the issue is slightly broader than the bug that prompted this - a similar issue will occur even with continuous treatments if a treatment featurizer is used. I think that this slight modification to your logic will cover it, but please add to one of the tests in `test_treatment_featurization` to also cover this logic.

```suggestion
        if self.transformer:
            T_2_score = self.transformer.transform(T)
        else:
            T_2_score = T
```

.gitignore
```diff
@@ -117,4 +117,4 @@ double_ml/proto/*.jpg
 doc/_autosummary/
 
 # compiled C files
-*.c
\ No newline at end of file
+*.cscripts/
```
kbattocchi: Was this inadvertent? Please revert.

kbattocchi (changes requested):
Thanks for the contribution! I really appreciate your taking the time to track down the issue and fix it, but I think the scope should be slightly broader - see comment in _rlearner.py. Once that's addressed I'd be happy to approve and merge this in.

immu4989:
Thanks for the review @kbattocchi — both points were on the money.

Pushed an updated commit that:

1. **Broader fix in `_rlearner.py`**: applied your suggestion, with one small adjustment — kept the `reshape(T, (-1, 1))` step when `discrete_treatment=True` since `OneHotEncoder.transform` requires 2D input (the literal `self.transformer.transform(T)` broke the existing discrete test). The result mirrors the existing pattern in `_OrthoLearner._fit_nuisances`:

   ```python
   if self.transformer:
       if self.discrete_treatment:
           T = reshape(T, (-1, 1))
       T_2_score = self.transformer.transform(T)
   else:
       T_2_score = T
2. Reverted the .gitignore change — inadvertent, sorry about that.

3. Added test_score_nuisances_applies_treatment_featurizer in test_treatment_featurization.py. Uses LinearDML with the existing polynomial_treatment_featurizer, fits on continuous T, and asserts score_nuisances returns finite scores. Verified pre fix it raises ValueError: y_true and y_pred have different number of output (1!=2) (the shape-mismatch path you anticipated); post-fix it passes.

Existing tests still pass: test_score_nuisances_discrete_treatment_label_alignment, test_forest_dml_score_fns, and a broader sanity sweep (27 passed across test_dml.py + test_treatment_featurization.py, no regressions outside the pre-existing *_with_ray env failures on main).



## Why this is a good outcome

- **Substantive maintainer engagement** — kbattocchi gave specific, actionable feedback; you responded within hours with a tested broader fix and a sensible departure from his literal suggestion (with justification).
- **Refined fix is more general** than the original PR scope. The petition reviewer reading the contribution graph end to end sees: investigation comment → PR → maintainer-requested broadening → broader fix landed. That's the "originality / significance / recognition" trifecta in a single artifact.
- **Two regression tests** now cover both bug modes (discrete-T encoding off by one and continuous-T+featurizer shape mismatch).

kbattocchi (approved):
Thanks for the updates, this looks great.

PR #1031, opened 2026-05-24
Fix scikit-learn 1.7+ FutureWarnings in WeightedLassoCV and WeightedMultiTaskLassoCV

justinchen033:
### Summary
This PR proactively addresses compatibility with `scikit-learn 1.7+` by resolving deprecation and `FutureWarning` messages triggered by the `n_alphas` and `alphas=None` parameters in `LassoCV` and `MultiTaskLassoCV` classes.

On environments running modern versions of `scikit-learn` (version 1.7+ / 1.8.0), running EconML's test suite generates **over 62,000 FutureWarnings** in `test_linear_model.py` alone. This PR resolves all of these warnings and future-proofs the package ahead of the `scikit-learn 1.9` release, where the deprecated parameters are scheduled to be removed entirely.

### Root Cause
In scikit-learn 1.7, the `n_alphas` parameter in CV estimators was deprecated in favor of passing an integer directly to the `alphas` parameter, and setting `alphas=None` is deprecated. In version 1.9, `n_alphas` will be removed entirely, which would otherwise cause breaking errors in EconML.

### Solution
We modified the constructors of `WeightedLassoCV` and `WeightedMultiTaskLassoCV` inside `econml/sklearn_extensions/linear_model.py` to check the scikit-learn version dynamically:
- **scikit-learn >= 1.7**: We pass `alphas = alphas if alphas is not None else n_alphas` and omit `n_alphas` when calling `super().__init__`.
- **scikit-learn < 1.7**: We fall back to the original signature to maintain full backward compatibility with older environments.

### Verification Results
1. **Warning Elimination**: Running `pytest econml/tests/test_linear_model.py` completed with **21 passed, 0 warnings** (compared to 62,538 warnings previously!).
2. **Regression & Correctness**: Verified model selection, RScorer, and bootstrap tests (`test_rscorer.py`, `test_model_selection.py`, `test_bootstrap.py`) run successfully and pass with no errors.

### Developer Certificate of Origin (DCO)
All commits have been signed off (`Signed-off-by`) in compliance with DCO guidelines.

### DIFF
--- a/econml/sklearn_extensions/linear_model.py
+++ b/econml/sklearn_extensions/linear_model.py
@@ -426,12 +426,25 @@ class WeightedLassoCV(WeightedModelMixin, LassoCV):
                  copy_X=True, cv=None, verbose=False, n_jobs=None,
                  positive=False, random_state=None, selection='cyclic'):
 
-        super().__init__(
-            eps=eps, n_alphas=n_alphas, alphas=alphas,
-            fit_intercept=fit_intercept,
-            precompute=precompute, max_iter=max_iter, tol=tol, copy_X=copy_X,
-            cv=cv, verbose=verbose, n_jobs=n_jobs, positive=positive,
-            random_state=random_state, selection=selection)
+        from packaging.version import parse
+        import sklearn
+
+        if parse(sklearn.__version__) >= parse("1.7"):
+            super().__init__(
+                eps=eps, alphas=alphas if alphas is not None else n_alphas,
+                fit_intercept=fit_intercept,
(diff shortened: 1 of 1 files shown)


kbattocchi:
Thanks, this seems like a change worth addressing and I'm in favor of it in concept.  Unfortunately, your commit includes a lot of formatting (whitespace and parenthesizing) changes that are unrelated to the core change and which make the diff hard to read. Can you create a cleaner commit that only includes the changes to the logic?

kbattocchi (approved):
Thanks for the contribution!

PR #1041, opened 2026-06-12
ci: redesign LKG to per-cell freeze files on orphan branch

kbattocchi:
**Do not merge yet.** This PR is for review only. Before merging, the following remains:

1. (Optional) Seed the `lkg` branch from a recent green nightly using `.github/workflows/seed_lkg_branch.py`. If skipped, the first post-merge nightly will populate it automatically.
2. Run through the verification recipe in the new `Maintainer guide` section of the README (manual `workflow_dispatch` smoke tests).

## What this PR does

Replaces the single merged `lkg.txt` / `lkg-notebook.txt` flow (generated by a marker-merging Python script and prone to two known bugs around partial reruns and cross-extras pinning) with **per-matrix-cell freeze files committed to a long-lived orphan `lkg` branch**. Each CI cell reads its own freeze directly — no merging anywhere.

Coverage extended to `build_sdist` and `create_docs` cells (which also benefit from pinning); `build_wheels` deliberately untouched since cibuildwheel manages its own envs.

## Commits (review in order)

1. **ci: add push-lkg job** — new producer; commits per-cell freezes to `lkg` from nightly / dispatch-from-main runs. Gated to `github.repository == 'py-why/EconML'` so forks don't pollute their own lkg branches.
2. **ci: install from per-cell lkg branch files** — swaps the install path in `tests`, `notebooks`, `build_sdist`, `create_docs` to read from a sparse checkout of `lkg`, with a bootstrap fallback. Also deletes the old root-level `lkg.txt` / `lkg-notebook.txt`.
3. **ci: remove dead LKG-generation jobs** — drops `store-reqs-per-env`, `generate-lkg`, `merge-lkg`, and `generate_lkg.py`.
4. **ci: add seed_lkg_branch.py and lint-job test step** — manual seed/recovery tool with pytest coverage; new lint-job step runs the tests.
5. **docs: add Maintainer guide section** — README section covering the design, file naming, trigger gating, one-time setup (ruleset recipe), seeding, and operational notes.
6. **docs: --no-verify in lkg seed recipe** — a fixup of #4/#5 left standalone so the gotcha (pre-commit blocks commits in the orphan worktree because there's no `.pre-commit-config.yaml`) is visible in history. Squash-friendly if you `Rebase and merge`.

## Testing done locally

- `ruff check` clean across all changed files.
- `pytest .github/workflows/tests/` — 13/13 pass.
- `python -c "import yaml; yaml.safe_load(...)"` on all three modified workflow files.
- Doctests in `seed_lkg_branch.py` pass.

## What can't be tested without merging

The `push-lkg` job's gating restricts it to `schedule` or `workflow_dispatch` from `main` with `use_lkg=false` — so it only runs after the redesign is on `main`. Validation requires the manual verification recipe in the README.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


kbattocchi:
## Design gap: a PR can no longer change pins and code atomically

Flagging something I ran into while debugging the nightly failures, because I think it needs an answer before this merges.

### The problem

Today `lkg.txt` is tracked in the repo, so a single PR can change dependency pins **and** the code that depends on those pins in one commit. PR CI validates the combination, and it either goes green together or red together.

After this PR, pins live on the long-lived `lkg` branch and are pulled in with `actions/checkout` (`ref: lkg`, `path: lkg-cache`), while `push-lkg` is gated to `github.ref == 'refs/heads/main'` plus schedule/dispatch. So a PR has no way to influence the pins its own CI runs against, and as far as I can tell there is currently no override or constraint escape hatch anywhere in the reworked `ci.yml`.

### A concrete deadlock (this is not hypothetical)

scikit-learn 1.9 changed `RandomForest`/`ExtraTrees` to use `sample_weight` when drawing the bootstrap sample rather than forwarding weights multiplied by a uniform mask. `NonParamDML` is maximally exposed because its final-model weights are literally `T_res**2`, so the doctest in `econml/dml/dml.py` moves:

```
sklearn 1.8:  array([0.35318952, 1.28760096, 0.83506031])
sklearn 1.9:  array([0.22215931, 1.6484125 , 0.9278963 ])
```

That is a ~37% change; no float tolerance will absorb it. Going green therefore requires **both** the LKG pinned at sklearn 1.9 **and** updated doctest values. Under this PR's model:

- Update the `lkg` branch first, and every open PR's docs job goes red until the doctest PR lands.
- Merge the doctest PR first, and its own required checks fail, because LKG still pins 1.8.

With the current in-repo `lkg.txt`, neither happens: one PR changes both. Worth stressing this is not doctest-specific, it applies to any upstream change that needs a coordinated code adaptation.

### One possible fix (not necessarily the right one)

An optional repo-tracked `lkg-overrides.txt` applied through uv's `--overrides`. I tested the mechanics against uv 0.12.0 because the obvious spelling does not work:

| form | result |
|---|---|
| `-c lkg.txt --overrides overrides.txt` | fails: `Because you require packaging==25.0 and packaging==24.2 ... unsatisfiable` |
| `-r lkg.txt --overrides overrides.txt` | override wins; real install produced `packaging==25.0`, annotated `# via --override overrides.txt` |

Constraints are not overridable, so `-c` is a dead end. This PR's install step already uses `-r "$LKG_FILE"`, so it would drop in roughly as:

```bash
OVR=""
[ -f lkg-overrides.txt ] && OVR="--overrides lkg-overrides.txt"
uv pip install --system -e ".${{ matrix.extras }}" -r "$LKG_FILE" $OVR
```

**The part I do not like about it:** it is not actually atomic, it just moves the seam. The PR lands with an override entry, the next nightly refreshes the `lkg` branch, and then a *second* change is needed to empty the override file again. That is easy to forget, and stale entries would silently keep pinning things after the branch has moved on.

### Other directions worth considering

Listing these mostly to open the design discussion rather than to advocate for any one:

1. **Auto-clear the overrides.** Keep the override file, but have `push-lkg` drop entries that the refreshed branch already satisfies (either committing directly or opening a small PR). Still two changes, but no human follow-up and no silent staleness.
2. **Per-PR opt-out instead of an override file.** A label or marker file meaning "this PR intentionally moves the dependency baseline, install floating for these cells." No cleanup step exists because there is no file to clean, and the PR is validated against exactly the versions that are about to become the new pins. Downside: that run is not reproducible later, and it opts out of pinning wholesale rather than for one package.
3. **Per-PR lkg ref.** Let an opted-in PR compute fresh pins and push them to a scratch ref (`lkg-pr-1234`), have the install step prefer it when present, and have the merge promote it into `lkg`. Genuinely atomic from the author's point of view, at the cost of more machinery and refs to garbage-collect.
4. **Keep per-cell files, but in the repo.** Most of the win here comes from going per-cell (killing the marker-merge bug and the cross-`kind` contamination), not from relocating the data off-repo. A `lkg/` directory of per-cell files would preserve atomicity by construction. The cost is nightly commits touching many files, which could be softened with `.gitattributes` `linguist-generated` so the diffs collapse by default.

I do not have a strong preference yet, but I would rather not merge without a story for the coordinated-change case, since the sklearn 1.9 doctest update is a live example that needs one shortly.

PR #1043, opened 2026-06-16
fix: Deprecate n_alphas across estimators for scikit-learn 1.9 compatibility

### LINKED ISSUE #1032: Scikit-learn v1.9 drops `n_alphas` from `LassoCV` and related estimators, causing `econml` to error on import

TasfinMahmud:
Fixes #1032 by introducing the new alphas parameter to affected estimators (e.g., SparseLinearDML, LinearDRLearner, DebiasedLasso), safely deprecating the old n_alphas parameter with a FutureWarning, and passing alphas directly down to scikit-learn. cc: @jakevdp

### DIFF
--- a/econml/dml/dml.py
+++ b/econml/dml/dml.py
@@ -1134,8 +1134,10 @@ class SparseLinearDML(DebiasedLassoCateEstimatorMixin, DML):
                  model_y='auto', model_t='auto',
                  alpha='auto',
                  n_alphas=100,
+                 alphas=None,
                  alpha_cov='auto',
                  n_alphas_cov=10,
+                 alphas_cov=None,
                  max_iter=1000,
                  tol=1e-4,
                  n_jobs=None,
@@ -1153,10 +1155,19 @@ class SparseLinearDML(DebiasedLassoCateEstimatorMixin, DML):
                  allow_missing=False,
                  use_ray=False,
                  ray_remote_func_options=None):
+        import warnings
+        if n_alphas != 100:
+            warnings.warn("The n_alphas parameter is deprecated and will be removed in a future release. "
+                          "Use the alphas parameter instead.", FutureWarning)
+        if n_alphas_cov != 10:
+            warnings.warn("The n_alphas_cov parameter is deprecated and will be removed in a future release. "
+                          "Use the alphas_cov parameter instead.", FutureWarning)
         self.alpha = alpha
         self.n_alphas = n_alphas
+        self.alphas = alphas
         self.alpha_cov = alpha_cov
         self.n_alphas_cov = n_alphas_cov
+        self.alphas_cov = alphas_cov
         self.max_iter = max_iter
         self.tol = tol
         self.n_jobs = n_jobs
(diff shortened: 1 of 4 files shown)


kbattocchi:
Thanks for taking on #1032.

Unfortunately this PR doesn't fix the reported bug. jakevdp's crash originates from `WeightedLassoCV.__init__` passing `n_alphas=n_alphas` to `super().__init__()` on sklearn 1.9 (see line 429 in the traceback), which this PR doesn't modify. With this patch applied, `import econml.orf` still crashes on sklearn 1.9 with the same `TypeError`. The internal call in `_get_theta_coefs_and_tau_sq` also still passes `n_alphas=n_alphas_cov` down to `WeightedLassoCV`, so even the extended `alphas_cov` API you're adding to the high-level estimators would fail at that boundary.

We're merging #1042 for the immediate fix — it's a 2-line change to `WeightedLassoCV.__init__` / `WeightedMultiTaskLassoCV.__init__` (drops the `self.alphas = alphas` overwrite that was clobbering sklearn's correctly-dispatched value) plus a targeted regression test and the `pyproject.toml` upper-bound bump.

The wider idea in this PR — deprecating `n_alphas` in favor of `alphas` at our own wrapper/estimator level, matching sklearn's direction — is a good one and roughly matches jakevdp's own suggestion in #1032. But it's a bigger design decision (public API deprecation, deprecation cycle length, how much of the existing `n_alphas`-forwarding to change) and should be a separate discussion once the immediate crash is unblocked. Happy to pick it up in a separate issue if you'd like to pursue it.

Closing in favor of #1042. Thanks again for the attempt.

PR #1052, opened 2026-08-05
Refresh LKG and record the sklearn 1.9 forest behavior change

kbattocchi:
**Stacked on #1051** — base is `kebatt/doctest-drop-ellipses`, so review that first.

This updates `lkg.txt` to use scikit-learn 1.9 (from 1.8), and updates one doctest that needs to change as a result. The updated lkg.txt is the one generated from yesterday's nightly run.

The reason to address this now is that our nightlies are failing due to an sklearn change in 1.9 that changes some values in doctests; this change is actually an improvement: sklearn 1.9 changed `RandomForest`/`ExtraTrees` to use `sample_weight` when drawing the bootstrap sample, rather than forwarding the weights alongside a uniform mask. `NonParamDML` is maximally exposed, because its final-stage weights are literally `T_res**2` and therefore highly skewed, so we need to update one array in our doctests:

```
array([0.35318, 1.28760, 0.83506])  ->  array([0.22215931, 1.6484125 , 0.9278963 ])
```

### DIFF
--- a/econml/dml/dml.py
+++ b/econml/dml/dml.py
@@ -1597,7 +1597,7 @@ class NonParamDML(_BaseDML):
         est.fit(y, T, X=X, W=None)
 
     >>> est.effect(X[:3])
-    array([0.35318, 1.28760, 0.83506])
+    array([0.22215931, 1.6484125 , 0.9278963 ])
     """
 
     def __init__(self, *,
--- a/lkg.txt
+++ b/lkg.txt
@@ -1,32 +1,34 @@
-Cython==3.2.4
+Cython==3.2.9
 Jinja2==3.1.6
 MarkupSafe==3.0.3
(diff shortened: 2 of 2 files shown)


kbattocchi:
Heads-up on CI coverage for this PR: `ci.yml` only triggers on `pull_request` targeting `main`

```yaml
on:
  pull_request:
    branches:
      - main
```

When using the new stacked PR feature on GitHub, this should be okay (because it treats each PR in the stack as if it targets the base of the bottom PR) but we'll have to see whether there are any catches (e.g. I created this stack using the web interface after the fact, so maybe CI won't trigger until there's another push to this PR).

In the future we may want to consider further changes in light of the guidance [here](https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/optimizing-ci-for-stacked-pull-requests), in particular around not running our full matrix for each PR in the stack each time a PR is merged.

kbattocchi:
Closing in favor of a new approach in #1055; I'll grab the doctest updates from here and include them in a commit there.

PR #1055, opened 2026-08-10
ci: per-cell LKG on an orphan branch, plus dependency profiles as uv config files

kbattocchi:
Supersedes #1041, whose commits are included here rebased onto current `main`.

Replaces the merged `lkg.txt` / `lkg-notebook.txt` files with per-cell freeze files on an orphan `lkg` branch, and introduces named dependency profiles.

## Dependency profiles

A profile is a [uv configuration file](https://docs.astral.sh/uv/reference/settings/) under `profiles/`, passed via `uv --config-file`. uv parses its own format, so `ci.yml` needs no parser and knows only the profile name.

| profile | pins | cells |
| --- | --- | --- |
| `current` | newest known-good everything | the existing matrix |
| `all-floor` | `resolution = "lowest-direct"` | ubuntu / 3.12, three kinds |

`all-floor` resolves every direct dependency to its declared lower bound, making it the only thing in CI that exercises the bounds in `pyproject.toml`. Test cells go from 94 to 97.

Dependency coverage was previously a side effect of which interpreter pinned what: the 18 (os, python) cells collapse to five distinct dependency sets, and the scikit-learn 1.6 / numpy 2.0 floors were reachable only through the Python 3.9 row.

## Freeze invalidation

Each freeze records the hash of the profile recipe it was generated from:

```
# recipe-sha: 4ed249e49727
scikit-learn==1.9.0
```

A cell uses its freeze only while that hash matches; otherwise it resolves from the recipe. Two consequences:

* **A fallback preserves the profile's intent.** An unconstrained install resolves to newest, which would silently turn `all-floor` into a duplicate of `current` — floating selects numpy 2.5.2 where `lowest-direct` selects 2.0.2.
* **A dependency change and the code it forces can land together.** Editing a recipe invalidates that profile's stale freezes, so the cells re-resolve and the PR is green on its own, with nothing to clean up afterwards. `MAINTAINING.md` documents this under "Changing dependencies and code together".

`profiles/all-floor.toml` records no freeze, since re-resolving from the floors is its whole purpose.

## Also included

* README's developer sections split into `CONTRIBUTING.md` and `MAINTAINING.md`, sequenced as the first commit so later commits document the new system in `MAINTAINING.md` directly.
* `verify` now depends on `eval`. It is the only required check on `main` and a skipped job counts as a pass, so an `eval` failure previously made every downstream job skip while the required check reported green.
* `NonParamDML`'s doctest updated for scikit-learn 1.9, which changed how `RandomForest` draws its weighted bootstrap. Without `lkg.txt` the docs job resolves 1.9, so the recorded value has to match it.
* `.github/workflows/tests/test_ci_matrix.py` expands the matrix the way Actions does and asserts that every cell names a profile with a recipe on disk, that `include`-created combinations carry the fields the job bodies dereference (they inherit nothing), and that recipes parse with no unrecognised keys.


kbattocchi:
## First CI run: what it tells us

Two things, one expected and one a real defect I've since fixed.

**1. The docs job failure is pre-existing, not from this change.** It is the scikit-learn 1.9 ellipsis drift:

```
Failed example: est.const_marginal_effect(X[:2])
Expected: array([[0.516888..., 0.995747...], ...])
Got:      array([[0.51691875, 0.99584224], ...])
```

That is exactly what #1051 fixes (the ellipsis defeats scipy-doctest's float-tolerant checker, so a 6e-5 drift is fatal) and #1052 completes.

It surfaced here because this run exercised the **bootstrap path for real**: the `lkg` branch still holds files under the old naming convention, so no cell found a matching freeze and every one resolved fresh — which pulled scikit-learn 1.9. So the run doubles as an unplanned end-to-end test of the fallback, and it behaved correctly. The `::notice::` lines confirm the intended branch was taken.

Practical consequence: **land #1051 and #1052 before this**, otherwise this PR's docs job stays red for reasons unrelated to its diff.

**2. Job names were leaking matrix internals.** The all-floor cells rendered as:

```
Run tests (ubuntu-latest, 3.12, other, all-floor, false, -m "cate_api and not ray" -n auto, [plt])
```

A combination created by `include:` exposes *every* key in its auto-generated name — including `record_freeze` and the raw pytest `opts`. Unreadable, and unstable in a way that matters: adding any matrix property renames the check and would silently break a branch protection rule naming it.

Fixed in c12dd2c1 with explicit names (`Run tests (current, ubuntu-latest, 3.12, dml)`), plus a test asserting the job name pins the displayed identity so a future property cannot leak back in.

Still waiting on the `all-floor` cells, which are the genuinely unknown part — nothing here has ever run scikit-learn 1.6 on a modern interpreter.

